PR #24, opened 2021-05-14
fix: webpack.config.js in TesterApp

deorst:
fix ReactNative dependency issue in webpack.config.js

### Summary
Faced an error trying to start iOS app with the webpack.config.js
The current change resolves that issue.


examples/TesterApp/webpack.config.js
```diff
@@ -1,6 +1,6 @@
 const webpack = require('webpack');
 const path = require('path');
-const ReactNative = require('../..');
+const ReactNative = require('@callstack/nativepack');
```
zamotany, line 3: This change will break tester app. There's no `@callstack/nativepack` dependency inside tester app.

zamotany (changes requested):

PR #32, opened 2021-05-19
feat: remote chunks android

souhe:
### Summary

Android support for remote chunks. Implementation of below methods on NativeModule:
- loadChunk - only improvements
- preloadChunk
- invalidateChunks

### Test plan

I added a Remote section in Tester app where user can test downloading chunk from a remote server. Creating a server with chunk is required in order to do that

### DIFF
--- a/android/src/main/java/com/callstack/nativepack/RemoteChunkLoader.kt
+++ b/android/src/main/java/com/callstack/nativepack/RemoteChunkLoader.kt
@@ -9,11 +9,13 @@ import java.io.IOException
 import java.io.OutputStreamWriter
 import java.net.URL
 
-class RemoteChunkLoader(private val reactContext: ReactContext) : ChunkLoader {
+val CHUNKS_DIR = "chunks"
+
+class RemoteChunkLoader(private val reactContext: ReactContext) {
     private val client = OkHttpClient()
 
     private fun getChunkFilePath(hash: String, id: String): String {
-        return "$hash/$id.chunk.bundle"
+        return "${CHUNKS_DIR}/$hash/$id.chunk.bundle"
     }
 
     private fun downloadAndCache(hash: String, id: String, url: URL, onSuccess: () -> Unit, onError: (code: String, message: String) -> Unit) {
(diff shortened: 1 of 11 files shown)


src/client/chunks-api/ChunkManager.ts
```diff
@@ -61,12 +61,21 @@ class ChunkManager {
...
-      const chunkHash = this.getChunkHash(url);
-      delete this.resolveCache[chunkId];
-      await NativeModules.ChunkManager.invalidateChunk(chunkHash, chunkId, url);
+      const chunks = await Promise.all(
+        chunksIds.map(async (chunkId) => {
+          const url = await this.resolveChunk(chunkId);
+          const chunkHash = this.getChunkHash(url);
+          delete this.resolveCache[chunkId];
+
+          return {
```
souhe, line 72: @zamotany is there a possibility that we will add more data here (ex. chunkId)? If not I can just have a list of hashes instead of objects and send it to native

android/src/main/java/com/callstack/nativepack/RemoteChunkLoader.kt
```diff
@@ -4,18 +4,27 @@ import android.content.Context.MODE_PRIVATE
 import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactContext
 import okhttp3.*
+import java.io.File
 import java.io.IOException
 import java.io.OutputStreamWriter
 import java.net.URL
 
-class RemoteChunkLoader(private val reactContext: ReactContext) : ChunkLoader {
+val CHUNKS_DIR = "chunks"
```
zamotany, line 12: This can be a instance member of `RemoveChunkLoader`

zamotany (changes requested):
Loos fine for most part. We need to improve the error handling though - add additional error codes and messages to differentiate load/preload error messages ideally.

souhe:
> Loos fine for most part. We need to improve the error handling though - add additional error codes and messages to differentiate load/preload error messages ideally.

Any preferences for the error codes @zamotany?

zamotany (approved):

PR #34, opened 2021-05-24
chore: improve TesterApp to include more real-life examples

zamotany:
### Summary

Screenshot 2021-05-24 at 14 20 33


examples/TesterApp/src/MiniAppsContainer.js
```diff
@@ -0,0 +1,41 @@
+import React, { useCallback, useRef, useState } from 'react';
+import { Button, View } from 'react-native';
+import { ChunkManager } from '../../../client';
+
+export function MiniAppsContainer() {
+  const [isInstalled, setIsInstalled] = useState(false);
+  const [isVisible, setIsVisible] = useState(false);
+  const miniAppContent = useRef();
```
lukewalczak, line 8: suggestion: can we simplify further `miniAppContent` uses by destructuring current from ref?

lukewalczak (approved):

examples/TesterApp/src/MiniAppsContainer.js
zamotany: Kinda but we will still need `miniAppContent` so that code like `miniAppContent.current = bundle.default;` works. If we destruct `current` as the reassign it `current = bundle.default` this is likely not going to work. With Babel it might, but it's not a good solution. You can try this code to see the problem:
```
const ref = { current: undefined };
let { current } = ref;
current = 'whatever';
console.log(ref); // `current` is still `undefined`
```

PR #39, opened 2021-05-31
feat: runInBackground thread for iOS

deorst:
### Summary

Added method `runInBackground` which lets to run methods `loadChunk`, `preloadChunk` and `invalidateChunks` without blocking the UI thread for iOS.

### Test plan

Manually test `TesterApp` using 2s delay in methods `loadChunk`, `preloadChunk` and `invalidateChunks`.
The app is responsive hence UI thread is not blocked.

https://user-images.githubusercontent.com/12425395/120200366-b0cdc100-c23d-11eb-8b59-1dc07c7078c3.mov

### DIFF
--- a/ios/ChunkManager.mm
+++ b/ios/ChunkManager.mm
@@ -65,11 +65,11 @@ RCT_EXPORT_METHOD(preloadChunk:(nonnull NSString*)chunkId
                   withResolver:(RCTPromiseResolveBlock)resolve
                   withRejecter:(RCTPromiseRejectBlock)reject)
 {
-    [self runInBackground:^(){
-        if (!fetch) {
-            // Do nothing, chunk is already preloaded
-            resolve(nil);
-        } else {
+    if (!fetch) {
+        // Do nothing, chunk is already preloaded
+        resolve(nil);
+    } else {
+        [self runInBackground:^(){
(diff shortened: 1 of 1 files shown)


ios/ChunkManager.mm
```diff
@@ -29,87 +29,93 @@ @implementation ChunkManager
...
+                    if (error) {
+                        reject(ChunkDownloadFailure, error.localizedFailureReason, nil);
+                    } else {
+                        resolve(nil);
+                    }
+                }];
+            } else {
+                reject(UnsupportedScheme,
+                       [NSString stringWithFormat:@"Scheme in URL '%@' is not supported", chunkUrlString], nil);
+            }
```
zamotany, line 85: I sugget to change it to:
```objc
if (!fetch) {
  // Do nothing, chunk is already preloaded
  resolve(nil);
} else {
  [self runInBackground:^() {
    ...
  }];
}
```
To avoid unnecessarily spawning new threads.

zamotany (changes requested):

ios/ChunkManager.mm
deorst: Done.

zamotany (approved):

Issue #41, opened 2021-06-04
Default Android configuration fails out of the box


DaveWelling:
## Environment

Tried on both a Pixel 3 XL and a Zebra TC52 (android) device.

  "dependencies": {
    "react": "17.0.1",
    "react-native": "0.64.2"
  },
  "devDependencies": {
    "@babel/core": "^7.12.9",
    "@babel/runtime": "^7.12.5",
    "@callstack/nativepack": "^1.4.2",
    "@react-native-community/eslint-config": "^2.0.0",
    "babel-jest": "^26.6.3",
    "babel-loader": "^8.2.2",
    "eslint": "7.14.0",
    "jest": "^26.6.3",
    "metro-react-native-babel-preset": "^0.64.0",
    "react-test-renderer": "17.0.1",
    "terser-webpack-plugin": "^5.1.3",
    "webpack": "^5.38.1"
  },

node version 15.3.0

## Description

If I configure a new project for Windows/Android as described here: https://reactnative.dev/docs/environment-setup

Then make the changes described here:
https://github.com/callstack/nativepack#installation--setup

I paste in the webpack.config.js template from here: 
https://github.com/callstack/nativepack/blob/main/templates/webpack.config.js

I can run the bundling server normally and create a bundle without any errors.  However, when the app starts, I will receive the following error:
```
06-04 19:06:02.565  8033 10134 E unknown:ReactNative: Exception in native call from JS
06-04 19:06:02.565  8033 10134 E unknown:ReactNative: com.facebook.react.devsupport.JSException: Unexpected token '='. Expected an opening '(' before a method's parameter list.
06-04 19:06:02.565  8033 10134 E unknown:ReactNative:   at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
06-04 19:06:02.565  8033 10134 E unknown:ReactNative:   at android.os.Handler.handleCallback(Handler.java:938)
06-04 19:06:02.565  8033 10134 E unknown:ReactNative:   at android.os.Handler.dispatchMessage(Handler.java:99)
06-04 19:06:02.565  8033 10134 E unknown:ReactNative:   at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
06-04 19:06:02.565  8033 10134 E unknown:ReactNative:   at android.os.Looper.loop(Looper.java:223)
[...]

PR #42, opened 2021-06-11
fix: small windows support fix in babel loader in webpack config

DaveWelling:
### Summary

Small regex change on babel loader in webpack to support windows.

### Test plan

n/a


DaveWelling:
resolves #41

zamotany (approved):

PR #65, opened 2021-08-12
fix: Use fastify-reply-from plugin to ensure headers are not lost when proxying

jaridmargolin:
### Summary

**Background:** I encountered issues while trying to incorporate an imported mp4 into my react-native application using `react-native-video`. After investigation, I found that the video was unable to play due to missing headers (specifically, range headers). I found that the headers are correctly set by webpack-dev-server, but are then lost when proxying through repack.

**Description:** This pull request makes updates to the repack DevServerProxy module, choosing to use `fastify-reply-from` as a more robust mechanism to forward requests to the underlying webpack-dev-server.

**Alternatives:** The proxying logic could have been added inline rather than pulling in a 3rd party module, but using a more robust implementation seemed like a smaller burden in both upfront cost and continued maintenance.

### Test plan

I did not have a clear/easy path for testing this change directly within `repack` due to the current state of testing infra.
* I would have been happy to add a unit/integration test had there been a bit more scaffolding in place.
* Introducing a dependency like react-native-video into the e2e test seemed heavy-handed.

In a world where I had more time, I would have probably tried to add a light integration/unit test within repack that confirmed headers weren't dropped.

So the current state of testing is that I have confirmed this fixes my issue by manually copying the build output into the repack install within my project's node_modules.

I'd prefer not to spend too much more time on this issue, but happy to iterate if you see low-hanging fruit for increasing test coverage and/or building confidence in the PR.

### DIFF
--- a/packages/repack/src/server/DevServerProxy.ts
+++ b/packages/repack/src/server/DevServerProxy.ts
@@ -1,9 +1,9 @@
 import path from 'path';
 import { Writable } from 'stream';
 import execa from 'execa';
-import fetch from 'node-fetch';
 import getPort from 'get-port';
 import split2 from 'split2';
+import fastifyReplyFrom from 'fastify-reply-from';
 import { CliOptions, StartArguments } from '../types';
 import { Reporter } from '../Reporter';
 import {
@@ -211,14 +211,7 @@ export class DevServerProxy extends BaseDevServer {
         method: request.method,
         body: request.body,
       });
-      const response = await fetch(compilerWorkerUrl, {
-        method: request.method,
-        body: typeof request.body === 'string' ? request.body : undefined,
-      });
-      const payload = await response.buffer();
-      reply
-        .type(response.headers.get('Content-Type') || 'text/plain')
-        .send(payload);
+      return reply.from(compilerWorkerUrl);
     }
   }
 
@@ -244,6 +237,13 @@ export class DevServerProxy extends BaseDevServer {
 
     await super.setup();
 
+    this.fastify.register(fastifyReplyFrom, {
+      undici: {
+        headersTimeout: 5 * 60 * 1000,
+        bodyTimeout: 5 * 60 * 1000,
+      },
+    });
+
     this.fastify.post('/symbolicate', async (request, reply) => {
       const { stack } = JSON.parse(request.body as string) as {
         stack: ReactNativeStackFrame[];
@@ -252,7 +252,7 @@ export class DevServerProxy extends BaseDevServer {
       if (!platform) {
         reply.code(400).send();
       } else {
-        await this.forwardRequest(platform, request, reply);
+        return this.forwardRequest(platform, request, reply);
       }
     });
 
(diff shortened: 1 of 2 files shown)


packages/repack/src/server/DevServerProxy.ts
```diff
@@ -282,17 +282,12 @@ export class DevServerProxy extends BaseDevServer {
...
-          } catch (error) {
-            console.error(error);
-            reply.code(500).send();
-          }
+          Promise.resolve(!this.workers[platform] && this.runWorker(platform))
+            .then(() => this.forwardRequest(platform, request, reply)) // eslint-disable-line
+            .catch((error) => {
+              console.error(error);
+              reply.code(500).send();
+            });
```
zamotany, line 290: What's the motivation for this change? It doesn't fix anything and makes the code less readable. I'd suggest to revert it.

zamotany (changes requested):

packages/repack/src/server/DevServerProxy.ts
jaridmargolin: The motivation for this workaround was a bug I ran into while trying to implement `reply.from()` in the `forwardRequest` method.

> Promise may not be fulfilled with 'undefined' when statusCode is not 204

As I understood the error, within an async handler, you need to ensure that all data is sent prior to the handler resolving. TS shows that the `reply.from`, now implemented within `forwardRequest`, returns void, which means it cannot be awaited. I assumed that meant the internals of the function would call send.  My solution then was to modify the handler to not return a promise and instead complete the request via the send method called within `reply.from`.

Revisiting the issue after your comment, and perhaps there is another solution: https://github.com/fastify/fastify-reply-from/issues/105#issuecomment-707696530.

TBH, I'm not entirely sure why it would work if the `reply.from` returns void, but perhaps this is an issue with the types and I can actually await the `reply.from` call. I'll have to investigate further if this is a blocking issue for you.

zamotany: Ok so it's more nuanced than I thought.
Based on the docs: https://www.fastify.io/docs/latest/Routes/#async-await, more specifically section about returning reply:
```js
fastify.get('/', options, async function (request, reply) {
  setImmediate(() => {
    reply.send({ hello: 'world' })
  })
  return reply
})
```
and the comment you've posted I think that this should works:
```js
async (request, reply) => {
  // ... rest of the code
  reply.from(...)
  return reply;
}
```

zamotany: Would this work:
```js
async forwardRequest(...) {
  // ...
  reply.from(...)
  return reply;
}
// ... 
handler: async (request, reply) => {
  // ...
  return await this.forwardRequest(...)
}
```
??

jaridmargolin: Alright, I just pushed updates containing the requested changes.

According to the docs of fastify, `decorateReply()` which is what **fastify-reply-from** utilizes, [sets the context to the Reply instance](https://www.fastify.io/docs/latest/Decorators/#decoratereplyname-value-dependencies). And looking at the source code of **fastify-reply-from**, the return value of `from()` [is actually `this`](https://github.com/fastify/fastify-reply-from/blob/069ee15af3797c7b75a42d90019711efba25c901/index.js#L158), [not void](https://github.com/fastify/fastify-reply-from/blob/069ee15af3797c7b75a42d90019711efba25c901/index.d.ts#L63). In short, this means I can update the handler to return`this.forwardRequest(...)` as long as it returns `reply.from()`.

I have manually tested, and everything appears to be behaving correctly.

One additional note is that the log output from proxying is a little bit noisier than the previous implementation. The following logs are now included and it doesn't appear there is a simple option within **fastify-reply-from** to disable them:

```
ℹ [21:14:10.836Z][DevServerProxy] fetching from remote server {
  source: 'http://localhost:56008/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=app.staple.local'
} 
ℹ [21:14:10.853Z][DevServerProxy] response received 
```

PR #86, opened 2021-09-20
fix: windows support fix in HMR

joe06102:
fix: windows support fix in TargetPlugin for HMRClient && remove default 'localhost' in getDevServerOptions for windows

### Summary

HMR is not working on windows due to the incorrect regexp in `webpack.NormalModuleReplacementPlugin`.
Besides, the fallback value for `host` in `getDevServerOptions` should be removed to allow access for DevServer by IP in LAN.

### DIFF
--- a/packages/repack/src/types.ts
+++ b/packages/repack/src/types.ts
@@ -82,7 +82,7 @@ export interface StartArguments extends CommonArguments {
 export interface DevServerOptions {
   /** Whether to start development server. */
   enabled?: boolean;
-  /** Hostname under which to run the development server. Defaults to `localhost`. */
+  /** Hostname under which to run the development server. */
   host?: string;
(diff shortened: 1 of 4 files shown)


packages/repack/src/types.ts
```diff
@@ -82,7 +82,7 @@ export interface StartArguments extends CommonArguments {
 export interface DevServerOptions {
   /** Whether to start development server. */
   enabled?: boolean;
-  /** Hostname under which to run the development server. Defaults to `localhost`. */
+  /** Hostname under which to run the development server. */
```
zamotany, line 85:
```suggestion
  /**
   * Hostname or IP address under which to run the development server.
   * When left unspecified, it will listen on all available network interfaces, similarly to listening on '0.0.0.0'.
   */
```

zamotany (approved):

PR #92, opened 2021-09-23
fix: force the webpack.config.js cache invalidation

joe06102:
force the cache of webpack.config.js in bundle.ts to invalidate to allow multiple bundles for different platforms

### Summary

android & ios needs bundling separately and we call the `commands/bundle.ts` as below:

```javascript
  /**
  * platformOptions contain options for android & ios platforms
  */
  for (const opt of platformOptions) {
    bundle(/** bundle related args */)
  }
```

Bundles for each platform are expected but what we got is a duplicate of ios bundle for android platform, which does not work on android devices.

And the reason is that nodejs caches every module required throughout the program. I wonder if it is better to move all the webpack resolving logic to some function like getWebpackConfig?

Besides, I'd like to be noticed when the bundling is done. So I append a done callback to the bundle params. And `compiler.close` is supposed to called for webpack5 to allow for the low-priority work.

### DIFF
--- a/packages/repack/src/commands/bundle.ts
+++ b/packages/repack/src/commands/bundle.ts
@@ -38,6 +38,10 @@ export function bundle(_: string[], config: Config, args: BundleArguments) {
     process.env[VERBOSE_ENV_KEY] = '1';
   }
 
+  /**
+   * force nodejs to invalidate the webpack config to ensure the bundling works as expected.
+   */
+  delete require.cache[require.resolve(webpackConfigPath)];
   const compiler = webpack(require(webpackConfigPath));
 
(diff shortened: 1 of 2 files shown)


zamotany:
Hey, thanks for the contribution, but I cannot accept it. This is not an intended use for `bundle` command. Re.Pack follows the behaviour of Metro in that regard, where `bundle` command is platform specific. Based on the changes and your description, what you are looking for is a Node API, which Re.Pack currently doesn't have and it's not planned for immediate future. If you want to run Re.Pack from Node script, the recommended way is to spawn separate processes for each platform which would run `bundle` command.

PR #109, opened 2021-10-05
fix: prevent crash when using custom Webpack path

### LINKED ISSUE #105: Custom webpack.config.js path broken
## Description

Since #80 repack will throw when the CLI runner calculates the default value for the `--webpackConfig` argument if none of the default locations exists even if the passed argument is valid. Wrapping the `default` option in a try/catch makes the command work again.

zamotany:
### Summary

- Fixes #105 

### Test plan

1. Run Re.Pack's start/bundle in TesterApp (it should not crash)
1. Change `webpack.config.js` to `webpack2.config.js`
1. Run Re.Pack's start/bundle in TesterApp (it should crash)
1. Run Re.Pack's start/bundle in TesterApp with `--webpackConfig ./webpack2.config.js` (it should not crash)

### DIFF
--- a/packages/repack/commands.js
+++ b/packages/repack/commands.js
@@ -47,7 +47,12 @@ const webpackConfigOption = {
     const {
       getWebpackConfigPath,
     } = require('./dist/commands/utils/getWebpackConfigPath');
-    return getWebpackConfigPath(config.root);
+
+    try {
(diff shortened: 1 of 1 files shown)


jgornick:
I was just about to submit an issue for this 😛 

Thank you for fixing!

PR #81, opened 2021-09-03
feat: add configurable asset loader

zamotany:
### Summary

Closes #73 

### Test plan

1. Run TesterApp
2. The application should display SVG image

PR #133, opened 2021-11-04
Fix scalable asset resolution

oblador:
### Summary

#81 introduced a regression where all assets were treated as non-scalable. This PR fixes that by fixing a faulty regular expression and stop using a faulty regular expression where one isn't needed. 

### Test plan

Rename `callstack-dark.png` to `callstack-dark@2x.png` and observe that it is not included in the `webpack-bundle` output – yet not throwing any errors. With this change applied it is included.


zamotany:
The regexp is fine, the problem is `assetLoader` when calling `collectScales` is passing `type` without `.` here:
https://github.com/callstack/repack/blob/main/packages/repack/src/webpack/loaders/assetsLoader.ts#L105

Simple addition of a `.` will solve the problem:
```diff
const scales = AssetResolver.collectScales(
      getAssetExtensionsRegExp(options.scalableAssetExtensions),
      files,
      {
        name: filename,
-       type,
+       type: `.${type}`,
        platform: options.platform,
      }
    );
```

In other words `type` in `assetLoader` is not the same `type` in `AssetResolver`.
To me `type` should include the leading `.`. In all the places but one (https://github.com/callstack/repack/blob/main/packages/repack/src/webpack/loaders/assetsLoader.ts#L277), the leading `.` is needed.

oblador:
Right, but I would argue using regexp here is unnecessary.

oblador:
What do you want to do with this PR/bug?

zamotany:
I'm going to test it and merge

zamotany (approved):

zamotany:
It seems to work fine for `webpack-bundle`. With development server, React Native is always requesting asset without a scale, but it seems to be an issue that existed before this PR.

PR #134, opened 2021-11-15
fix: absolute flag and promise resolving inside the file system chunk loader

TMaszko:
### Summary
Fix setting absolute flag and resolving promise inside the file system chunk loader

### Test plan
Add tests for absolute equals`true`.


packages/repack/src/client/api/__tests__/ChunkManagerBackend.spec.ts
```diff
@@ -241,4 +241,31 @@ describe('ChunkManager', () => {
       timeout: DEFAULT_TIMEOUT,
     });
   });
+  it('should resolve with absolute path', async () => {
```
zamotany, line 244:
```suggestion

  it('should resolve with absolute path', async () => {
```

zamotany (approved):

Issue #141, opened 2021-12-14
Invalid URL port: "undefined"


cstranex:
## Environment

- React-Native 0.66.3
- Android

## Description

Running `react-native webpack-start` results in error message on Android device: `Invalid URL port: "undefined"`

## Reproducible Demo

The webpack.config.js is the same as per the template. Hermes is enabled and there isn't really any other special configuration enabled. Logs show:
```
[20:27:04.678Z][DevServerProxy] Server listening at http://[::]:8081 
ℹ [20:27:04.680Z][DevServerProxy] Dashboard available at: http://localhost:8081/dashboard 
ℹ [20:27:06.024Z][DevServerProxy] Hermes device connected { deviceId: 0 } 
ℹ [20:27:25.051Z][DevServerProxy] Dashboard client connected { clientId: 'client#0' } 
ℹ [20:27:39.099Z][DevServerProxy] Starting compiler worker { platform: 'android', port: 46031 } 
ℹ [20:27:39.966Z][DevServer@android] Server listening at http://[::]:46031 
ℹ [20:27:39.970Z][DevServerProxy] fetching from remote server { source: 'http://localhost:46031/index.bundle?platform=android' } 
ℹ [20:27:40.044Z][DevServerProxy] fetching from remote server {
  source: 'http://localhost:46031/api/dashboard/stats?platform=android'
} 
ℹ [20:27:40.046Z][DevServerPlugin] Successfully run: /opt/android-sdk/platform-tools/adb reverse tcp:46031 tcp:46031 
ℹ [20:27:40.055Z][webpack-dev-middleware] wait until bundle finished: /index.bundle?platform=android 
ℹ [20:27:40.061Z][DevServer@android] Dashboard client connected { clientId: 'client#0' } 
ℹ [20:27:40.080Z][webpack-dev-middleware] wait until bundle finished: /api/dashboard/stats?platform=android 
ℹ [20:28:06.399Z][LoggerPlugin] Bundle built { time: 26241 } 
ℹ [20:28:06.432Z][DevServerProxy] response received 
ℹ [20:28:06.472Z][DevServerProxy] response received 
ℹ [20:28:55.618Z][DevServerProxy] fetching from remote server {
  source: 'http://localhost:46031/index.bundle?platform=android&dev=true&minify=false&app=com.dominate&modulesOnly=false&runModule=true'
} 
ℹ [20:28:55.642Z][DevServerProxy] response received 
[...]

PR #145, opened 2021-12-22
fix: replace process.env.__PUBLIC_PORT__ by __PUBLIC_PORT__

roshangm1:
### Summary

In some cases, value returned by process.env.__PUBLIC_PORT__ was undefined because of which hot reload wasn't working.
Fixes https://github.com/callstack/repack/issues/141

### Test plan

Make sure that hot reload is working as expected.


roshan-lmi:
@zamotany Anything I need to do here to fix the issue with preview_website pipeline ?

zamotany:
No, it's not necessary. I'll merge the PR once I test it.

PR #153, opened 2022-02-02
fix: images with higher resolution not used when possible

### LINKED ISSUE #148: Higher resolution images are not used
## Description

If i include higher resolution version of an image in the form of suffixes (ex test_image.png & test_image@2x.png) the higher resolution is never used. Even if the device supports it.

## Reproducible Demo

Create two versions of an image and name them something like 'test_image.png' and 'test_image@2x.png'.
Import and use them in the TesterApp App.js.

Expected result is that 'test_image@2x.png'  is shown if the device support it. But its the default image that is shown allways.

## Source of bug & suggested fix

I think i've tracked down the error to how assets are registered in the assetLoader webpack loader.  This line:

https://github.com/callstack/repack/blob/b9c3473842e7699620587bdab0e7c3555815d6ab/packages/repack/src/webpack/loaders/assetsLoader.ts#L274

registers the scales of the asset by setting the property 'scales' to an object containing the different versions of the asset looking like this: 

`{"@1x": {"name": "test_image.png"}, "@2x": {"name": "test_image@2x.png"}} `

But accordning to React Native that property should be an array of numbers.

https://github.com/facebook/react-native/blob/37bd9793d50cc8011ec371fb87d12b6066d4a4ae/packages/assets/registry.js#L13-L24

By changing line 274 in assetLoader to:

`scales: ${JSON.stringify(Object.keys(scales).map(scaleKey => parseFloat(scaleKey.replace(/[^\d.]/g, ''))))},`

It sets the property to [1,2]

And with that change I'm able to get the correct image resolution to be shown in the app.

TMaszko:
### Summary
It fixes #148 .

### Test plan
Add two images and name it same way but one of them should have `@2x` suffix. On an iOS simulator image with mentioned suffix should be displayed.

### DIFF
--- a/packages/repack/src/webpack/loaders/assetsLoader.ts
+++ b/packages/repack/src/webpack/loaders/assetsLoader.ts
@@ -112,6 +112,9 @@ export default async function reactNativeAssetsLoader(this: LoaderContext) {
         parseInt(b.replace(/[^\d.]/g, ''), 10)
     );
 
+    const scaleNumbers = scaleKeys.map((scale) =>
+      parseInt(scale.replace(/[^\d.]/g, ''), 10)
+    );
     const assets = await Promise.all(
       scaleKeys.map(
         (
@@ -271,7 +274,7 @@ export default async function reactNativeAssetsLoader(this: LoaderContext) {
       var AssetRegistry = require('react-native/Libraries/Image/AssetRegistry');
       module.exports = AssetRegistry.registerAsset({
         __packager_asset: true,
-        scales: ${JSON.stringify(scales)},
+        scales: ${JSON.stringify(scaleNumbers)},
         name: ${JSON.stringify(filename)},
         type: ${JSON.stringify(type)},
         hash: ${JSON.stringify(hashes.join())},


zamotany (approved):

packages/repack/src/webpack/loaders/assetsLoader.ts
```diff
@@ -112,6 +112,9 @@ export default async function reactNativeAssetsLoader(this: LoaderContext) {
         parseInt(b.replace(/[^\d.]/g, ''), 10)
     );
 
+    const scaleNumbers = scaleKeys.map((scale) =>
+      parseInt(scale.replace(/[^\d.]/g, ''), 10)
```
bjornallvin, line 116: Should not this have been a parseFloat? In the switch statement on row 140 you expect resolutions to be able to come in as floats. Not sure that is actually used in the wild. But if they do this could be a potentail bug   @zamotany

TMaszko: Thanks :) you're right. I'll fix it 👍

TMaszko: https://github.com/callstack/repack/pull/156

PR #154, opened 2022-02-03
fix: wrong error position on JSC engine

### LINKED ISSUE #128: `getInitializationEntries` contents causes incorrect `Error` position
## Environment

Both iOS and Android, with or without hermes.

## Description

When throwing an error inside a repack app (including the TesterApp) the line + column combination is not matching the actual position in the bundle file, therefore the source map lookup ends up on the wrong file/line. The produced source map itself is correct and maps perfectly if getting the position by inspecting the bundle manually or running the bundle via node. 

This seems to be caused by the import of `react-native` package somehow. 

## Reproducible Demo

Edit `TesterApp/index.js` to be:
```
throw new Error('Lol');
```

and remove `ReactNative.getInitializationEntries` spread from `webpack.config.js` and observe that the line+column thrown in RN are correct. Add back `ReactNative.getInitializationEntries` and see that it doesn't match anymore.

TMaszko:
### Summary
In this PR including any comments to the minified bundle has been disabled.
It fixes #128

### Test plan
JSC engine should return correct column number when we've thrown error.

### DIFF
--- a/packages/TesterApp/webpack.config.js
+++ b/packages/TesterApp/webpack.config.js
@@ -36,7 +36,7 @@ const dev = mode === 'development';
 const context = ReactNative.getContext();
 const entry = ReactNative.getEntry();
 const platform = ReactNative.getPlatform({ fallback: process.env.PLATFORM });
-const minimize = ReactNative.isMinimizeEnabled({ fallback: !dev });
+const minimize = true;
 const devServer = ReactNative.getDevServerOptions();
 const reactNativePath = ReactNative.getReactNativePath();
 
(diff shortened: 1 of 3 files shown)


packages/TesterApp/webpack.config.js
```diff
@@ -36,7 +36,7 @@ const dev = mode === 'development';
 const context = ReactNative.getContext();
 const entry = ReactNative.getEntry();
 const platform = ReactNative.getPlatform({ fallback: process.env.PLATFORM });
-const minimize = ReactNative.isMinimizeEnabled({ fallback: !dev });
+const minimize = true;
```
TMaszko, line 39: revert this

zamotany (approved):

PR #190, opened 2022-06-30
feat(repack, dev-server): improve Module Federation support

zamotany:
### Summary

- Add `Federated` utilities
- Tweak bundle output paths


tom-sherman:
Would you be able to publish to a `@v3` tag on npm at all?

zamotany:
> Would you be able to publish to a `@v3` tag on npm at all?

I could, but I need to revisit publishing process which will take some time.
In the meantime I want to create a new `repack-example` repro, which you have Re.Pack dependencies packed into `.tgz` archive, so you'll be able to use those archive for testing - this will also be a good showcase for the API and utilities.

singhalpranav22:
Is @v3 ready to use @zamotany ?

zamotany:
> Is `@v3` ready to use @zamotany ?

Not yet.

packages/repack/src/webpack/federated.ts
```diff
@@ -0,0 +1,60 @@
...
+   *         remotes: {
+   *           app1: Repack.Federated.createRemote('app1'),
+   *         },
+   *       }),
+   *     ],
+   *   };
+   * };
+   * ```
+   */
+  export function createRemote(remoteName: string) {
```
tom-sherman, line 36: As I understand it this still requires calling `ScriptManager.configure` before being able to federate anything right?

Would it be possible to support `"app@remoteUrl"` syntax? ie. configuring the ScriptManager via the remotes config directly. This would remove the need for a top level `.configure()` call and allow  omnidirectional hosts.

zamotany: Not a `app@remoteUrl` syntax but technically it could be possible to support `Federated.createRemote('app', 'remoteUrl')` or `Federated.createRemote('app@remoteUrl')`. Could you create a diagram how would omnidirectional hosts look like? I'll create an example to see if there won't be any additional problems.

My main concern at the moment is that, when I tried using any remote in `host`, it was and it will fail because it tries to call `initExternal` for that remote, before React/React Native and Re.Pack itself have a chance to finish initialising. So using remotes in `host` looks more or less like:

1. Create `__webpack_require__`
2. Create `__webpack_require__.repack` runtime
3. Evaluate entries: React Native polyfills, `react-native/Libraries/Core/InitializeCore`
4. Somewhere in `react-native/Libraries/Core/InitializeCore` there's an import to `webpack/sharing/consume/default/react-native/react-native`
5. `webpack/sharing/consume/default/react-native/react-native` is being pulled into default scope
6. `default` scope is created in `__webpack_require__.I` (`webpack/runtime/sharing`) and `register`s `react`, `react-native` and calls `initExternal('app')`
7. `initExternal('app')` will use code created by `Federated.createRemote`, which has to use `Promise`
8. The promise from 7. is retuned and it throws on:
   ```ts
   if(typeof factory !== "function") throw new Error("Shared module is not available for eager consumption: " + id);
   ```

tom-sherman: Yes sorry I should have lead with my specific use case.

In the following diagram dotted lines indicate federation.

Each app is a separate respository deployed independently. They may or may not each publish to the app store, but they need to have this ability in order to support local development where they federate their remotes (including the Shell). The crucial workflow here being I can checkout `FooApp`, start the app, and have fast refresh work correctly for my local code inside FooApp.

```mermaidjs
graph TD
   %% Dotted lines indicate federation
    subgraph ShellApp
        ShellAppIndex["index.js"] --> ShellAppBootstrap["bootstrap.js"]
        ShellAppBootstrap -- React.lazy --> ShellAppApp["App.js"]
        ShellAppApp --> ShellAppShell["Shell.js"]
    end

    subgraph FooApp
        FooAppIndex["index.js"] --> FooAppBootstrap["bootstrap.js"]
        FooAppBootstrap -- React.lazy --> FooAppApp["App.js"]
        FooAppApp -.-> ShellAppShell["Shell.js"]
        FooAppComponent["SomeComponent.js"]
    end

    subgraph BarApp
        BarAppIndex["index.js"] --> BarAppBootstrap["bootstrap.js"]
        BarAppBootstrap -- React.lazy --> BarAppApp["App.js"]
        BarAppApp -.-> ShellAppShell["Shell.js"]
        BarAppApp -.-> FooAppComponent
    end
```

This is just one specific case, but being able to to declare your remotes like this opens up many possibilities as seen in https://github.com/module-federation/module-federation-examples

> before React/React Native and Re.Pack itself have a chance to finish initialising.

As I understand it this is why the need for async boundaries exist, they ensure you have at least one dynamic import before any sync imports. The boundary is where the runtime waits for things to be initialised.

See https://github.com/tom-sherman/module-federation-repack/blob/466d5fddcb22391c380ebbe5b4583415c311721d/host/bootstrap.js

tom-sherman: > With this code Webpack will request someRemote to be loaded even before it get's to index.js.

Any why is this a problem exactly? Because the bridge won't be initialised blocking you from calling `executeApplicationScript`/`loadScriptFromFile`? The webpack runtime itself is setup at this point, can you hold onto the remotes until the bridge is initialised if this happens?

zamotany: > Any why is this a problem exactly?

There are couple:
1. Repack runtime is, but ScriptManager isn't - but we could work around this, by deferring `ScriptManager.shared.loadScript` call until everything is ready.
2. The bigger problem is that if we return `new Promise` from remote Webpack will throw error on this condition:
   ```js
   if(typeof factory !== "function") throw new Error("Shared module is not available for eager consumption: " + id);
   ```
   `factory` here is a `Promise`.

ScriptedAlchemy: We should be able to resolve eager errors and were able to use the externalized/promise syntax 
If theres a specific shared module throwing eager errors, even when shared as eager - we can assign things into share scope manually or with a loader. 

Would love to investigate @ syntax support further

zamotany: https://github.com/callstack/repack/discussions/199

ScriptedAlchemy: Yeah no problem, Im happy to help debug or solution it.

PR #198, opened 2022-07-07
fix: track initialized remote containers

ScriptedAlchemy:
### Summary

Remotes should not try to re-init if they have already been initialized.
In webpack runtime, this causes perf drags. 
In low level api, this will cause apps to crash if you try to load 2 modules from the same remote

A solution i use in many places is adding a initilized flag onto the remote, so if i use webpack runtime or low level api to initialize then container, they can be dynamically mixed, nested and so on without causing problems 

### Test plan

Try loading remote containers or multiple exposed modules from one remote (low level api)
Try loading nested remote containers that import the same container at both tiers (high level api)

### DIFF
--- a/packages/repack/src/webpack/federated.ts
+++ b/packages/repack/src/webpack/federated.ts
@@ -46,7 +46,7 @@ export namespace Federated {
             init: (arg) => {
               if(!self.${remoteName}._initialized) {
                 try {
-                  self.${remoteName}._initialized = true
+                  self.${remoteName}._initialized = true;
                   return self.${remoteName}.init(arg);
                 } catch (e) {
                   console.log('[Repack/Federated] Remote container ${remoteName} already initialized.');


packages/repack/src/webpack/federated.ts
```diff
@@ -44,10 +44,13 @@ export namespace Federated {
...
-              try {
-                return self.${remoteName}.init(arg);
-              } catch (e) {
-                console.log('[Repack/Federated] Remote container ${remoteName} already initialized.');
+              if(!self.${remoteName}._initialized) {
+                try {
+                  self.${remoteName}._initialized = true
+                  return self.${remoteName}.init(arg);
+                } catch (e) {
+                  console.log('[Repack/Federated] Remote container ${remoteName} already initialized.');
```
ScriptedAlchemy, line 52: good idea to still keep the try catch just incase someone loaded a remote in some non standard way & the initialized flag was not set

zamotany:
Thanks for the PR!

It needs to be rebased on `v3` first and make sure you run `yarn install`.

packages/repack/src/modules/ScriptManager/federated.ts
```diff
@@ -261,9 +261,11 @@ export namespace Federated {
     await ScriptManager.shared.loadScript(containerName);
 
     const container = self[containerName];
-
-    // Initialize the container, it may provide shared modules
-    await container.init(__webpack_share_scopes__[scope]);
+    if(!container._initialized) {
```
zamotany, line 264:
```suggestion
    if (!container._initialized) {
```
Otherwise Prettier will complain

ScriptedAlchemy:
My bad. I thought this was off v3. Will do!

zamotany:
My bad, I forgot to commit updated yarn.lock onto `v3`. Could you rebase again?

ScriptedAlchemy:
ahh gotcha, hold on will do

packages/repack/src/modules/ScriptManager/federated.ts
```diff
@@ -261,9 +261,11 @@ export namespace Federated {
     await ScriptManager.shared.loadScript(containerName);
 
     const container = self[containerName];
-
-    // Initialize the container, it may provide shared modules
-    await container.init(__webpack_share_scopes__[scope]);
+    if (!container._initialized) {
+      // Initialize the container, it may provide shared modules
+      await container.init(__webpack_share_scopes__[scope]);
+      container._initialized = true
```
zamotany, line 267:
```suggestion
      container._initialized = true;
```

ScriptedAlchemy:
Thank you

PR #227, opened 2022-09-08
fix: Exit with non-zero exit code if compilation has errors

eps1lon:
### Summary

Exit with non-zero exit code if compilation has errors

### Test plan

-  [x] CI
- [x] Tested in a local patch with the Klarna app


jbinda:
Hi @eps1lon - thanks for raising the PR. I think the proposed change makes sense! Will approve and merge soon

jbinda (approved):

PR #231, opened 2022-09-13
refactor: adjust webpack.config templates

jbinda:
### Summary

Attach additional statement related ith `devServer.hmr` for MF. 

No changeset required.

### DIFF
--- a/templates/webpack.config.cjs
+++ b/templates/webpack.config.cjs
@@ -36,6 +36,18 @@ module.exports = (env) => {
   }
 
   /**
+   * Using Module Federation might require disabling hmr.
+   * Uncomment below to set `devServer.hmr` to `false`.
+   *
+   * Keep in mind that `devServer` object is not available
+   * when running `webpack-bundle` command and crash
+   * because of reading from `undefined` will occur. 
+   */
+  // if (devServer) {
+  //   devServer.hmr = false;
+  // }
+
+  /**
    * Depending on your Babel configuration you might want to keep it.
    * If you don't use `env` in your Babel config, you can remove it.
    *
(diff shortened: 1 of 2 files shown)


templates/webpack.config.cjs
```diff
@@ -35,6 +35,18 @@ module.exports = (env) => {
     throw new Error('Missing platform');
   }
 
+  /**
+   * Using Module Federation might require disabling hmr.
+   * Uncomment below to set `devServer.hmr` to `false`.
+   *
+   * Keep in mind that `devServer` object is not available
+   * when running `webpack-bundle` command and crash
+   * because of reading from `undefined` will occur.
```
RafikiTiki, line 44:
```suggestion
   * Keep in mind that `devServer` object is not available
   * when running `webpack-bundle` command. Be sure
   * to check its value to avoid accessing undefined value,
   * otherwise an error might occur.
```

I think that is clearer to understand.

templates/webpack.config.mjs
```diff
@@ -37,6 +37,18 @@ export default (env) => {
     throw new Error('Missing platform');
   }
 
+    /**
+   * Using Module Federation might require disabling hmr.
+   * Uncomment below to set `devServer.hmr` to `false`.
+   *
+   * Keep in mind that `devServer` object is not available
+   * when running `webpack-bundle` command and crash
+   * because of reading from `undefined` will occur.
```
RafikiTiki, line 46:
```suggestion
   * Keep in mind that `devServer` object is not available
   * when running `webpack-bundle` command. Be sure
   * to check its value to avoid accessing undefined value,
   * otherwise an error might occur.
```

I think that is clearer to understand.

RafikiTiki (approved):
A little change, but I'm not pushing for it

andrewworld (approved):

PR #234, opened 2022-09-13
refactor: tester app

jbinda:
### Summary

- rearrange files structure in `TesterApp` to expose important logic 
- hide less important UI elements behind additional abstraction layer under `ui` folder
- place `Async`, `Remote` and `Miniapp` in separate folders
- mIgrate `TesterApp` files to ts


packages/TesterApp/src/ui/types.ts
```diff
@@ -0,0 +1 @@
+export type WithChildren<T> = React.PropsWithChildren<T>;
```
RafikiTiki, line 1: Since this type is just an alias for built-in React type wouldn't it be better to use it explicitly?

RafikiTiki (approved):

PR #237, opened 2022-09-15
fix: expose port reverse argument

jbinda:
### Summary

When working with `repack-examples` on Android there the error `[ScriptManager] Failed to load script: [NetworkFailure] ...` occured when an attempt to load scripts from remote dev servers runs on different port that host app. It's because those remote ports are not reversed when ScriptManager tries to resolve from `localhost`. Logic to reverse ports is implemented on `watchRun` which seems to be 'too' late to prevent mentioned crash.

This PR expose `--reverse-port` flag which allow to reverse port upfront when starting dev server.

### Test plan

- [x] run Module Federation example on Android device or Emulator and not facing `[ScriptManager] Failed to load script: [NetworkFailure] ...`

### DIFF
--- a/.changeset/brown-jokes-remain.md
+++ b/.changeset/brown-jokes-remain.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+Expose `--reverse-port` argument in start command to fix dev server on Android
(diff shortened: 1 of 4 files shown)


RafikiTiki (approved):
Nice and clean solution!

PR #241, opened 2022-09-16
fix: typo in dirname in templates/webpack.config.cjs fix

gorbypark:
This fixes a typo in `templates/webpack.config.cjs` which causes webpack to not work/crash.  It changes `dirname` to `__dirname`.


jbinda:
hey @gorbypark thank you for catching that ! Appproved

jbinda (approved):

PR #244, opened 2022-09-21
docs: versioned docs

jbinda:
### Summary

Cherry-picked versioned docs from V2 + add version dropdown


jbinda:
still need to handle links from footer docs - that's why CI fails

thymikee (approved):
As a next step let's add Algolia search here :) You'll need to apply for it, following the docs on Docusaurus site

thymikee:
CI says "Error: Docusaurus found broken links"

jbinda:
> CI says "Error: Docusaurus found broken links"

yes, it's because:
- footer doesn't handle versioning (I removed path: 3x from config to fallback to most recent docs in that case) - ideally we would like to have generic page in footer that shows main page with proper sidebar
- v2 of docs has `React Native API` and `Node API` in navbar while V3 has only `API`

jbinda:
I have fixed all links to versioned docs in v2. Thanks for review !

thymikee (approved):

PR #256, opened 2022-09-29
fix(webpack/Compiler): assetsCache miss on windows

### LINKED ISSUE #247: Problem loading assets from dev server on windows
Currently there's a serious problem on windows, which took me hours to track down and do a local fix

**tl;dr**: made the issue to create a PR fixing this.

The issue is that `filename` in [assetsCache](https://github.com/callstack/repack/blob/8480e2cb878c56bea5d6adf13ded5e579b300a2c/packages/repack/src/webpack/Compiler.ts#L113) in windows uses backslashes. when the app tries to load the asset, it requests the localhost server using forward slashes, so it will miss. my quick solution to this was adding a duplicate key with slashes replaced by forward slashes and pointing to the same data. duplication was done because I do not know if the `assetsCache` keys are used inside production builds or not. since removing the backslashed paths would cause local file system access failure, I did not remove it. it will also simply replace the other key on unix paths.

meypod:
fixes #247

### DIFF
--- a/packages/repack/src/webpack/Compiler.ts
+++ b/packages/repack/src/webpack/Compiler.ts
@@ -111,13 +111,16 @@ export class Compiler extends EventEmitter {
           this.isCompilationInProgress[platform] = false;
           this.statsCache[platform] = value.stats;
           this.assetsCache[platform] = value.assets.reduce(
-            (acc, { filename, data, info }) => ({
-              ...acc,
-              [filename]: {
+            (acc, { filename, data, info }) => {
+              const asset = {
                 data: Buffer.from(data),
                 info,
-              },
-            }),
+              };
+              return { ...acc,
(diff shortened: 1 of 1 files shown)


jbinda (approved):

jbinda:
hey @meypod thanks for rebasing according to my previous request ! Many appreciate. After merging your second PR and few others I had to sync with most recent main - I used Github rebase but I've noticed that CI fails on linting... Can I have another request to you to fix that issue ? PR looks good, already approved it and I'll merge as soon as all `Test / * ` jobs pass

meypod:
hi @jbinda 
I made the required changes
please check again
thanks

jbinda:
thank you once again

meypod:
any idea when will we get a new release with these fixes?

PR #265, opened 2022-10-10
fix: ensure file lists in manifests are sorted

justusburger:
### Summary

After building an app, the manifest (`{chunk}.json`) files contain file lists in the `files` and `auxiliaryFiles` fields. These file lists are not sorted, causing the total "output" of a build to change, even if the source did not change.

This PR adds sorting to fix this.


justusburger:
Happy for you to decide if you want to bump the version.

mikeduminy:
It'd also be great to get this in v2 🙏

jbroma (approved):
LGTM 🚀

PR #268, opened 2022-10-13
feat: Raise error on missing native module

robik:
This fix detects missing React Native module (e.g. due to missing `pod install` after installation) and raises error if true.

### Summary

This change improves developer experience. Should not affect any other areas.

Done a simple test on TesterApp and seems that nothing is broken by this change.

### Test plan

 1. Setup a RN application
 2. Install `repack` without updating native dependencies
 3. Setup Module Federation in app and use it
 4. New error `repack react-native module was not found` should appear instead of instead of `[ScriptManager] Failed to load script`

### DIFF
--- a/.changeset/tricky-papayas-perform.md
+++ b/.changeset/tricky-papayas-perform.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
+raise error on missing native module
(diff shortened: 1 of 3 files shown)


jbinda:
@robik thanks for that PR - it is a nice distinguish when it comes to point user to exact reason of Failing to Load script !

jbinda (approved):

PR #276, opened 2022-12-12
refactor(webpack/Compiler): remove duplicated assets in assetCache

RafikiTiki:
### Summary

This PR slightly refactors what was already implemented by @meypod in #256 – instead of duplicating each asset under a key which is valid pathname for Windows I'm modifying (or not) the filename based on the platform in context of which `Complier` is run. This should be safe as it will not affect production builds because the `Compiler` in only used with devServer.

### Test plan

Unfortunately I don't have access to a machine running Windows – @meypod would you be able to test that change and let me know if it's working properly on your end?

### DIFF
--- a/packages/repack/src/webpack/utils/adaptFilenameToPlatform.ts
+++ b/packages/repack/src/webpack/utils/adaptFilenameToPlatform.ts
@@ -1,0 +1,10 @@
+import os from 'os';
+
+const isWindows = os.platform() === 'win32';
+
+export const adaptFilenameToPlatform = (filename: string) => {
+  if (isWindows) {
+    return filename.replace(/\\/g, '/');
+  }
+  return filename;
+};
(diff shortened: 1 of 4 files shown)


packages/repack/src/webpack/utils/adaptFilenameToPlatform.ts
```diff
@@ -0,0 +1,10 @@
+import os from 'os';
+
+const isWindows = os.platform() === 'win32';
+
+export const adaptFilenameToPlatform = (filename: string) => {
+  if (isWindows) {
+    return filename.replace(/\\/g, '/');
```
thymikee, line 7: not sure if we use it already, but if we de, we could use `slash` library to do so

thymikee (approved):
LGTM

meypod:
Hi
thanks for this
I'll try to make these changes manually in node_modules and get back to tell you if it works within the hour
but I wonder if it's important to not have duplicated assets on windows ? since my change should be noop for linux users

meypod:
tested, seems to be working as expected, didn't encounter any issue

PR #278, opened 2022-12-19
docs: update old TesterApp urls for code splitting

TMisiukiewicz:
### Summary

Updated TesterApp urls in Code Splitting  v3 and v2 docs - old urls threw 404

### Test plan

n/a


thymikee (approved):
Thanks!

PR #287, opened 2023-01-04
feat: over-the-air updates

andrewworld:
### Summary
Repack provides a way to implement OTA updates out-of-the-box.
Since `ScriptManager` has `prefetchScript` and `loadScript` methods which use `resolveScript` under the hood, there is a possible way to add a point between resolving script and loading it. In this point user could control the way Repack updates the script (fetch from network or use cached one).
This PR introduces new callback, that could be passed into so called locator or a kind of config in `addResolver` callback function return statement.
```
shouldUpdateScript?: (
    scriptId?: string,
    caller?: string,
    isScriptCacheOutdated?: boolean
) => Promise<boolean> | boolean;
```
it is worth to mention that all the changes are optional, so nothing will be broken if user will not provide new callback and will do no changes at all.
`ShouldUpdateScript` callback receives `scriptId` and `caller` to identify the script. New `isScriptCacheOutdated` argument helps to identify if script is changed comparing to the cached one, since user has no ability to work directly with cache. It compares method, url, query, headers and body of the script request and set true if any of this params was changed or false when the script cache is up to date or there is no cache for the script.
Since `shouldUpdateScript` will be called on every `resolveScript` call it is necessary to provide correct conditions when script should be loaded or use already existing cache logic. 
If cache is opt-out for the script, `isScriptCacheOutdated` will always be false.
Here is the example of `shouldUpdateScript` usage:
```
const callers = {};

<SOME CODE...>

shouldUpdateScript: async (scriptId, caller, isScriptCacheOutdated) => {
      if (caller && callers[caller] !== undefined) {
        return callers[caller];
      }

      if (!isScriptCacheOutdated) {
        return true;
      }

      const shouldUpdate = await alertAsync(
        'Update available',
        'A new version of the app is available. Do you want to update?',
      );

      if (!caller) {
        callers[scriptId] = shouldUpdate;
      }

      return shouldUpdate;
    },
```
The implementation above uses `isScriptCacheOutdated` argument to check if script is changed and could be updated. If it is true, then application will show an alert with the question should the new script version be loaded or use the previous one from the cache. Callers is an object which stores the first user answer regarding the script. Since Webpack with Module Federation creates bundle for each remote container and then creates small chunks for it exposed modules, `shouldUpdateScript` will be called on every script (container bundle and its chunks). Callers object saves the first value from the alert (true or false) and then use it for resolving chunks calls. E.g. Host app has `auth` remote container, so Repack calls `shouldUpdateScript` for the container bundle with `scriptId` `auth`, saves the answer in callers object with key `auth` and then it will use this value from the first `alertAsync` call on any chunk load intent from `auth` remote container (caller will be `auth` and `scriptId` will be the chunk script id).

### Test plan
To test new functionality you need to add `shouldUpdateScript` callback in `addResolver` callback return statement and return boolean value depending on your case.

### DIFF
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -272,6 +272,24 @@ export class ScriptManager extends EventEmitter {
       const script = Script.from({ scriptId, caller }, locator, false);
       const cacheKey = `${scriptId}_${caller ?? 'unknown'}`;
 
+      if (locator.shouldLoad) {
+        const fetch = await locator.shouldLoad(
+          scriptId,
+          caller,
+          script.shouldUpdateCache(this.cache[cacheKey])
+        );
+
+        if (fetch) {
+          script.locator.fetch = true;
+          this.cache[cacheKey] = script.getCacheData();
+          await this.saveCache();
+        }
+
+        this.emit('resolved', script.toObject());
+
+        return script;
+      }
+
       if (!this.cache[cacheKey]) {
         script.locator.fetch = true;
         this.cache[cacheKey] = script.getCacheData();
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -272,8 +272,8 @@ export class ScriptManager extends EventEmitter {
       const script = Script.from({ scriptId, caller }, locator, false);
       const cacheKey = `${scriptId}_${caller ?? 'unknown'}`;
 
-      if (locator.shouldLoad) {
-        const fetch = await locator.shouldLoad(
+      if (locator.shouldUpdateScript) {
+        const fetch = await locator.shouldUpdateScript(
           scriptId,
           caller,
           script.shouldUpdateCache(this.cache[cacheKey])
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -290,10 +290,11 @@ export class ScriptManager extends EventEmitter {
 
         this.emit('resolved', script.toObject());
 
+        // if it returns false, we don't need to fetch the script
         return script;
       }
 
-      // If no custom shouldUpdateScript function was provided, we use the default one
+      // If no custom shouldUpdateScript function was provided, we use the default behaviour
       if (!this.cache[cacheKey]) {
         script.locator.fetch = true;
         this.cache[cacheKey] = script.getCacheData();
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -272,13 +272,16 @@ export class ScriptManager extends EventEmitter {
       const script = Script.from({ scriptId, caller }, locator, false);
       const cacheKey = `${scriptId}_${caller ?? 'unknown'}`;
 
+      // Check if user provided a custom shouldUpdateScript function
       if (locator.shouldUpdateScript) {
+        // If so, we need to wait for it to resolve
         const fetch = await locator.shouldUpdateScript(
           scriptId,
           caller,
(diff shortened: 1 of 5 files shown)


packages/repack/src/modules/ScriptManager/types.ts
```diff
@@ -88,6 +88,25 @@ export interface ScriptLocator {
...
+   * @param isOutdated Boolean indicating whether the script cache is outdated or not. It's `true` when the script
+   * cache is outdated and `false` when the script cache is up to date or there is no cache for the script.
+   * Outdated cache means that the script was previously downloaded and put into cache,
+   * but the script locator data (method, url, query, headers, or body) has changed since then.
+   * @returns Boolean indicating whether the script should be loaded or not
+   */
+  shouldUpdateScript?: (
+    scriptId?: string,
+    caller?: string,
+    isOutdated?: boolean
```
RafikiTiki, line 108: I have a hard time wrapping my head around the name `isOutdated`, especially in the context of how it is used in the description of the PR, specifically this part:

```
if (!isOutdated) {
        return true;
      }
```

It just looks weird to return `true` from function named `shouldUpdateScript` if _something_ is NOT outdated (`!isOutdated`). It just reads weird. I think if you would change the parameter name to `isCacheOutdated` or `isScriptCacheOutdated` it will be much clearer. It makes even more sense, because the actual value of that parameter is the result of evaluating `return this.checkIfCacheDataOutdated(cachedData)` and it reads like `return is the cache outdated`

packages/repack/src/modules/ScriptManager/types.ts
```diff
@@ -88,6 +88,25 @@ export interface ScriptLocator {
    * Setting this flat to `false`, disables that behavior.
    */
   cache?: boolean;
+  /**
+   * Function called before loading or getting from the cache and after resolving the script locator.
```
RafikiTiki, line 92: It might be a good idea to add one or two sentences here describing that it's the right place in the 'lifecycle' of consuming Script to ask the user if they want to update/download new version of the Script/s

RafikiTiki (approved):
After naming changes looks good to me!

packages/repack/src/modules/ScriptManager/ScriptManager.ts
```diff
@@ -272,6 +272,28 @@ export class ScriptManager extends EventEmitter {
       const script = Script.from({ scriptId, caller }, locator, false);
       const cacheKey = `${scriptId}_${caller ?? 'unknown'}`;
 
+      // Check if user provided a custom shouldUpdateScript function
+      if (locator.shouldUpdateScript) {
+        // If so, we need to wait for it to resolve
+        const fetch = await locator.shouldUpdateScript(
```
jbinda, line 278: Despite the same name of parameter under`srcipt.locator` object, I would suggest changing the name to something more meaningful. E.g. I thought at first glance that it's a method not a boolean.

Proposed name is eg `shoudlFetchScript`

@RafikiTiki wdyt ?

packages/repack/src/modules/ScriptManager/types.ts
```diff
@@ -88,6 +88,28 @@ export interface ScriptLocator {
...
+   *
+   * @param scriptId Id of the script to resolve.
+   * @param caller Name of the calling script - it can be for example: name of the bundle, chunk or container.
+   * @param isScriptCacheOutdated Boolean indicating whether the script cache is outdated or not. It's `true` when the script
+   * cache is outdated and `false` when the script cache is up to date or there is no cache for the script.
+   * Outdated cache means that the script was previously downloaded and put into cache,
+   * but the script locator data (method, url, query, headers, or body) has changed since then.
+   * @returns Boolean indicating whether the script should be loaded or not
+   */
+  shouldUpdateScript?: (
```
jbinda, line 108: here I also think might be better to rename to e.g. `getShouldUpdateScript` - the goal here is to point within it's name it's a method

jbinda (commented):
Hi @andrewworld and @RafikiTiki - I did review and logic seems good to me.

I left some comments related to naming convention that might be discussed

PR #317, opened 2023-03-15
Update "Configure Android" section

jaworek:
Newer versions of RN set `bundleCommand` in  a different place in`build.gradle`. This documentation change aims to update information on where it should be placed.


RafikiTiki (approved):
Thanks for this update!

PR #318, opened 2023-03-16
feat: code-signing

jbroma:
### Summary

This PR adds code-signing capabilities in form of signing the chunks & verifying them after downloading.

### Description

The changes implemented in this PR include:
- [x] `CodeSigningPlugin` - plugin for generating code-signing mapping file 
- [x] `CodeSigningUtils`- extension for ScriptManager that enables verifying signed bundles
   - [x] Android
   - [x] iOS

As we are using 3rd party libs for decoding & verifying the signature, the iOS part minimum deployment iOS version was raised to iOS 12. This is on par with current requirements of newest version of React-Native (0.71.4).

I've also cleaned up the iOS part of the `repack` package by removing the `.pbxproj` and associated files which were unnecessary when developing a library. The iOS part was written in Swift and the iOS project was adjusted to handle interoperability between Swift and Obj-C.

### Usage

1. Generate a pair of private & public RS256 keys that will be used for code-signing (like [this](https://gist.github.com/ygotthilf/baa58da5c3dd1f69fae9))
2. Add `Repack.CodeSigningPlugin` to the list of plugins in `webpack.config` and configure it with path to the private key and name of the output file.
```js
new Repack.plugins.CodeSigningPlugin({
        privateKeyPath: './private.pem',
        outputFile: 'code_signing_mapping.json',
        outputPath: path.join(dirname, 'build/outputs', platform, 'remote'),
}),
```
3. Configure the ScriptManager. Inside of `ScriptManager.shared.addResolver` do the following:
  a. Create an instance of `Federated.createTokenResolver` with the code signing mapping JSON converted to JS object.
  b. Use the token resolver to get the associate token with the bundle.
  c. Add `verifyScriptSignature` key set to `true` and pass the resolved `token` from the previous step to the config returned from `ScriptManager.shared.addResolver`
```js
ScriptManager.shared.addResolver(async (scriptId, caller) => {
  // ...
  // obtain your code-signing-mapping.json from filesystem or fetch it from a remote location
  const resolveToken = Federated.createTokenResolver(tokens);
  
  let url;
  let token;
  if (caller === 'main') {
    url = Script.getDevServerURL(scriptId);
  } else {
    url = resolveURL(scriptId, caller);
    token = resolveToken(scriptId, caller);
  }
  
  if (!url) {
    return undefined;
  }
  
  return {
    url,
    token,
    verifyScriptSignature: true,
    query: {
      platform: Platform.OS,
    },
  };
});
```

### Test plan

As TesterApp has no ModuleFederation example, the best way to test this feature is to use a new project with ModuleFederation. Setup everything as shown in the [Usage](#usage) section and you can verify code-signing by trying to tamper with the JWTs produced by the plugin.
### Collaborators

Thank you @andrewworld for doing most of the research, implementation and heavy lifting in this PR, especially for the android part of this feature!


RafikiTiki (approved):
Awesome work!

packages/repack/android/src/main/java/com/callstack/repack/RemoteChunkLoader.kt
```diff
@@ -44,9 +44,14 @@ class RemoteScriptLoader(private val reactContext: ReactContext) {
                             File(reactContext.filesDir, scriptsDirName).mkdir()
                         }
 
+                        val body = response.body?.string()
+
+                        if (config.verifyScriptSignature != null && config.verifyScriptSignature == true) {
```
dratwas, line 49: Checking the `true` case is not enough?

packages/repack/src/webpack/plugins/CodeSigningPlugin.ts
```diff
@@ -75,12 +95,23 @@ export class CodeSigningPlugin implements WebpackPlugin {
...
+              path.join(
+                compiler.context,
+                this.config.outputPath,
+                this.config.outputFile!
+              ),
+              json
+            );
+          } else {
+            compilation.emitAsset(
+              this.config.outputFile!,
```
RafikiTiki, line 111: Should't the `outputFile` be marked as not-optional in the type definition? Since it's used in both branches of the if statement I'd say it's required.

RafikiTiki (approved):

packages/repack/src/webpack/plugins/CodeSigningPlugin.ts
RafikiTiki: I didn't see the default value in the Plugin's constructor. All good.

chenyulun:
This is a breaking change,  react-native@0.64.4, after installing the dependency IOS can not find the native dependency #351

chenyulun:
Version 0.68.0 executes `npx react-native config` without `dependencies.platforms.ios`.
```json
 "dependencies": {
    "@callstack/repack": {
      "root": "/Users/userName/learn/rntest/rnxKitTest/node_modules/@callstack/repack",
      "name": "@callstack/repack",
      "platforms": {
        "android": {
          "sourceDir": "/Users/userName/learn/rntest/rnxKitTest/node_modules/@callstack/repack/android",
          "packageImportPath": "import com.callstack.repack.ScriptManagerPackage;",
          "packageInstance": "new ScriptManagerPackage()",
          "buildTypes": []
        }
      }
    }
  },
````

IOS dependencies can only be resolved in versions 0.69.0 or higher
```json
 "dependencies": {
    "@callstack/repack": {
      "root": "/Users/userName/learn/rntest/rnxKitTest/node_modules/@callstack/repack",
      "name": "@callstack/repack",
      "platforms": {
        "ios": {
          "podspecPath": "/Users/userName/learn/rntest/rnxKitTest/node_modules/@callstack/repack/callstack-repack.podspec",
          "configurations": [],
          "scriptPhases": []
        },
        "android": {
          "sourceDir": "/Users/userName/learn/rntest/rnxKitTest/node_modules/@callstack/repack/android",
          "packageImportPath": "import com.callstack.repack.ScriptManagerPackage;",
          "packageInstance": "new ScriptManagerPackage()",
          "buildTypes": []
        }
      }
    }
  },
````

PR #331, opened 2023-04-05
feat(repack): remote-assets

### LINKED ISSUE #290: Remote Assets
## Motivation
When using Repack with Module Federation the only way to access the assets is to inline them via Repack's `assetLoader`. This solution results in assets being bigger in size by themselves and embedded into the bundle itself resulting in bigger size of it as well.

## Description
Proposed solution is to handle the assets the same way they're handled on the web - by keeping them separate and downloading them in the app only if needed. To achieve that, we would need to emit a directory containing all of the assets that would need to be uploaded to the remote host (e.g. CDN). Another step would be to change the result of RePack's `require` when using remote assets. The goal here is to make it compatible with `ReactNative`'s `Image` component. Similarly to how inlining assets works, the result of the `require` would look something like this:
```ts
const source = {
    uri: url.href,
    width: size?.width,
    height: size?.height,
  };
```
This solution involves encoding asset's dimension during compilation which eliminates the problem where the remote asset won't be displayed by `Image` component without styling the component (similar to inlined assets).

The `URL` inside of source object would be the result of prepending `PUBLIC_PATH` provided via `assetsLoader` configuration to the relative path of the resource. There is an important limitation in here that `PUBLIC_PATH` needs to be known at compile-time, similar to how it's done in Webpack

Here's how the configuration object for `assetsLoader` could like with options for remote assets:
```ts
export interface Options {
  platform: string;
  scalableAssetExtensions: string[];
  devServerEnabled?: boolean;
  inline?: boolean;
  publicPath?: string;
  remote?: {
    enabled: boolean;        // useful for switching between development and production env
    publicPath: string;      // hostname of the target remote location
    outDir?: string;         // name of the resulting directory containing remote assets
    publicOutDir?: string;   // name of the directory appended to the publicPath in case it's different than the outDir
    scalableAssets: 'min' | 'max' | 'default' | 'all'  // what scale to choose for the remote asset
  };
}
```
By proper configuration of the RePack's `assetsLoaders`, the developer could work on the local assets as before, and only use the remote functionality for the production build (after uploading the assets to the remote host).

When dealing with remote assets there's also aspect of asset's scalability. Simple solution to that would be to only choose 1 scale of the asset to emit as remote, and let the user configure the behaviour that governs it, e.g. whether we want to take the smallest one, the biggest or the default one with `@1x` scale. It would be a good idea to leave it as a default solution while also providing an option to disable that behaviour in case someone wants to emit every asset at every scale and handle the scalability themselves (e.g. via query params?). 

## Summary
What needs to be done involves the following:
1. Add asset transformer that will replace the result of RePack's `require` with `Source` object containg the URL
2. Add mechanism to select only one scale from scalable assets
3. Handle outputting the asset's directory
4. Add relevant configuration to `assetsLoader` options

## Relevant PRs
[feat(assetLoader): support remote assets](https://github.com/callstack/repack/pull/217) - this PR aims to deliver similar functionality

jbroma:
### Summary
closes #290 which contains relevant details of the feature.

Added functionality for converting local assets into remote-assets with full support for scaled assets that resolve to device scale in runtime.

### Description
In comparison with the solution proposed in #290, the `assetsLoader` options have been reduced:
```ts
export interface Options {
  platform: string;
  scalableAssetExtensions: string[];
  devServerEnabled?: boolean;
  inline?: boolean;
  publicPath?: string;
  remote?: {
    enabled: boolean;        // useful for switching between development and production env
    publicPath: string;       // hostname of the target remote location e.g. http://localhost:9000
  };
}
```
Additionally, `OutputPlugin` can copy the `remote-assets` into the `output` directory, (e.g. `build/outputs/ios/remote`) by specyfing `auxiliaryAssetsPath` in it's options (or via `output` options of the `RepackPlugin`)

### Usage
1. Configure `assetsLoader` to convert selected assets to remote:
```ts
// packages/TesterApp/webpack.config.js
{
  test: Repack.getAssetExtensionsRegExp(
    Repack.ASSET_EXTENSIONS.filter((ext) => ext !== "svg")
  ),
  include: [path.join(dirname, "src/assetsTest/remoteAssets")],
  use: {
    loader: "@callstack/repack/assets-loader",
    options: {
      platform,
      devServerEnabled: Boolean(devServer),
      scalableAssetExtensions: Repack.SCALABLE_ASSETS,
      remote: {
        enabled: true,
        publicPath: "http://localhost:9999",
      },
    },
  },
}
```
2.  Optionally, you can also configure `auxiliaryAssetsPath` if you want your remote-assets to appear in the output directory (remote-assets will also be present in the `build/generated` directory):
```ts
// packages/TesterApp/webpack.config.js
new Repack.RepackPlugin({
  context,
  mode,
  platform,
  devServer,
  output: {
    bundleFilename,
    sourceMapFilename,
    assetsPath,
    auxiliaryAssetsPath: path.join("build/output", platform, "remote"),
  },
});
```

See the example in the TesterApp on how to handle multiple `assetsLoader`'s in the `webpack.config.js`

The final step is to make these assets available to be consumed. For a production environment you would upload `remote-assets` directory to a remote CDN. The `TesterApp` contains a script called `serve-remote-assets` which starts a local `http-server` serving the contents of `build/outputs/ios/remote-assets`. Mind that you need to run the `bundle` command prior to serving the remote-assets. Please note that the output is the same for both `android` and `ios` so this will work perfectly fine regardless of the platform. 

### Limitations
The only limitations to this solution is that the remote-asset's `publicPath` needs to be known at build-time.

### Test plan
- Added unit/integration tests
- Added example in the TesterApp


packages/repack/src/webpack/loaders/assetsLoader/__tests__/assetsLoader.test.ts
```diff
@@ -245,4 +268,63 @@ describe('assetLoader', () => {
       );
     });
   });
+
+  describe('should conver to remote-asset', () => {
```
thymikee, line 272:
```suggestion
  describe('should convert to remote-asset', () => {
```

thymikee (commented):
The only thing I'm missing is documenting this in the docs. Describing when it's helpful to configure this and how to do it would be nice for developers looking to solve the problems you solve here

RafikiTiki (approved):

packages/repack/src/webpack/loaders/assetsLoader/assetsLoader.ts
```diff
@@ -186,24 +198,41 @@ export default async function repackAssetsLoader(this: LoaderContext) {
...
 
-      callback?.(
-        null,
-        await extractAssets(
-          {
-            resourcePath,
+      if (options.remote?.enabled) {
+        callback?.(
+          null,
+          convertToRemoteAsset({
```
RafikiTiki, line 204: convertToRemoteAsset -> convertToRemoteAsset`s`

thymikee (approved):
🚀 A good followup would be to consider including this as a default for MF plugin, so that the configuration scope is smaller for typical user

mlakmal:
@jbroma are we able to use dynamic remote url for publicPath at runtime to fetch the images?

or are we expecting remote assets to be always exposed via specific publicPath not related to remote app?

Issue #215, opened 2022-08-06
[ERR_UNSUPPORTED_ESM_URL_SCHEME] - Module Federation Example


hifiguitarguy:
## Environment

Windows 11
Node v16.13.1

## Description

[start:*host] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
[start:*host]     at new NodeError (node:internal/errors:371:5)
[start:*host]     at defaultResolve (node:internal/modules/esm/resolve:1016:11)
[start:*host]     at ESMLoader.resolve (node:internal/modules/esm/loader:422:30)
[start:*host]     at ESMLoader.getModuleJob (node:internal/modules/esm/loader:222:40)
[start:*host]     at ESMLoader.import (node:internal/modules/esm/loader:276:22)
[start:*host]     at importModuleDynamically (node:internal/modules/cjs/loader:1041:29)
[start:*host]     at importModuleDynamicallyWrapper (node:internal/vm/module:437:21)
[start:*host]     at importModuleDynamically (node:vm:381:46)
[start:*host]     at importModuleDynamicallyCallback (node:internal/process/esm_loader:35:14)
[start:*host]     at loadWebpackConfig (..\host\node_modules\@callstack\repack\dist\webpack\loadWebpackConfig.js:14:5)
[start:*host] Emitted 'error' event on Compiler instance at:
[start:*host]     at Worker.<anonymous> (..\host\node_modules\@callstack\repack\dist\webpack\Compiler.js:106:14)
[start:*host]     at Worker.emit (node:events:390:28)
[start:*host]     at MessagePort.<anonymous> (node:internal/worker:232:53)
[start:*host]     at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
[start:*host]     at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28)

## Reproducible Demo

https://github.com/callstack/repack-examples
start:host on Windows will fail to bundle on request for 'host' on 8081

hifiguitarguy:
I will at least suggest VSCode Container Remotes as a workaround (keep it in Linux). Has been working well for SSH-to-Mac as well.

onurkantar:
No solution yet?

onurkantar:
I changed webpack.config.mjs to webpack.config.cjs (the code too ofc). It worked for me.

[...]

Issue #340, opened 2023-04-11
Does repack works fine on Windows?


ViniciusUrias:
### Discussed in https://github.com/callstack/repack/discussions/339

<div type='discussions-op-text'>

<sup>Originally posted by **ViniciusUrias** April 11, 2023</sup>
I was using windows 10, but all repack/module-federation examples were crashing.
Errors like:
'Can't find __webpack_require__';
'[ERR_UNSUPPORTED_ESM_URL_SCHEME]';
'Unhandled error. ([ScriptManager] Failed to load script:)';

Then I upgraded to windows 11 and the same errors still ocurring.
I think the last solution is change to some linux distro.
Obs: I tested it like 100 times before coming here, used 3 different versions of node, react and react-native, tested with .mjs, .cjs and .js, created a brand new app to configure, cloned the existing examples and I simply can't find a solution</div>

piccinnigius:
Hi, I was able to make it work with Windows 11. 
What can i say is that:

1.  `webpack_require` or cant't find xxx method can be fixed commonly by adding `/node_modules(.*[/\\])+@callstack/` or correspective dependency to babel-loader section in `webpack.config.mjs`.
2. [ERR_UNSUPPORTED_ESM_URL_SCHEME] can be hard-fixed by editing the source code of the @callstack library. This error happens because on windows seems that the require function (on webpack mjs config file) error is fatal and can't be managed by a try catch block and then it raise this error. I'm working on a pull request that can manage this behavior.

hope that this can helps you

Ibnunds:
For windows user need to use webpack-config.cjs template, dont use the .mjs template

jbroma:
Hey @ViniciusUrias,

we've recently released [Re.Pack 3.5.0](https://github.com/callstack/repack/releases/tag/%40callstack%2Frepack%403.5.0) which address a major issue for users on Windows. Try updating `@callstack/repack` to the newest version and see whether your issue is still relevant, thanks!

jbroma:
Closing as issue became stale and solution was provided.

LilMaiaV2:
@callstack/repack not work on windows in react native project

PR #343, opened 2023-04-14
fix: ERR_UNSUPPORTED_ESM_URL_SCHEME on Windows (#215)

piccinnigius:
### Summary
see #215

### Test plan

- check if all runs good on Linux, I tested only on Windows, but I've added Os check to prevent different behaviors from before
- check if failing tests are related to two edits in start.ts and bundle.ts (sorry for this, I'm not so much into tests)

### DIFF
--- a/packages/TesterApp/webpack.config.mjs
+++ b/packages/TesterApp/webpack.config.mjs
@@ -31,7 +31,7 @@ export default (env) => {
     reactNativePath = new URL('./node_modules/react-native', import.meta.url)
       .pathname,
   } = env;
-  const dirname = Repack.getDirname(import.meta.url);
+  const dirname = context;
 
   if (!platform) {
     throw new Error('Missing platform');
--- a/packages/repack/src/commands/utils/getWebpackConfigPath.ts
+++ b/packages/repack/src/commands/utils/getWebpackConfigPath.ts
@@ -1,5 +1,6 @@
 import path from 'path';
 import fs from 'fs';
+import Os from 'os';
 
 // Supports the same files as Webpack CLI.
 const DEFAULT_WEBPACK_CONFIG_LOCATIONS = [
@@ -22,7 +23,15 @@ export function getWebpackConfigPath(root: string, customPath?: string) {
       ? candidate
       : path.join(root, candidate);
     if (fs.existsSync(filename)) {
-      return filename;
+      if (
+        path.isAbsolute(candidate) &&
+        candidate.endsWith('.mjs') &&
+        Os.platform() === 'win32'
+      ) {
+        return `file:\\${filename}`;
+      } else {
+        return filename;
+      }
     }
   }
 


piccinnigius:
i've also seen that linting has not worked :/

packages/repack/src/commands/bundle.ts
```diff
@@ -1,10 +1,11 @@
...
-import { getWebpackEnvOptions } from '../webpack/utils';
-import { getWebpackConfigPath } from './utils/getWebpackConfigPath';
+import { Config } from "@react-native-community/cli-types";
+import webpack from "webpack";
+import { VERBOSE_ENV_KEY } from "../env";
+import { BundleArguments, CliOptions } from "../types";
+import { loadWebpackConfig } from "../webpack/loadWebpackConfig";
+import { getWebpackEnvOptions } from "../webpack/utils";
+import { getWebpackConfigPath } from "./utils/getWebpackConfigPath";
+import Os from "os";
```
wodin, line 8: It looks like the reset of the project uses single quotes, so it seems like a bad idea to change them here.

packages/TesterApp/webpack.config.mjs
```diff
@@ -1,7 +1,7 @@
 import path from 'path';
 import TerserPlugin from 'terser-webpack-plugin';
 import * as Repack from '@callstack/repack';
-
+import Os from 'os';
```
wodin, line 4: The other files seem to have a blank line after the imports

packages/repack/src/commands/bundle.ts
```diff
@@ -19,51 +20,59 @@ import { getWebpackConfigPath } from './utils/getWebpackConfigPath';
...
-    },
-    command: 'bundle',
-    arguments: {
-      bundle: args,
-    },
-  } as CliOptions;
+	_: string[],
+	config: Config,
+	args: BundleArguments
+){
```
wodin, line 26: Switching to tabs make it hard to see your actual changes. Also, since the rest of the project seems to be using two spaces for indents, I think switching to tabs is perhaps not a good idea

packages/repack/src/commands/start.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from '../logging';
 import { Compiler } from '../webpack/Compiler';
 import { getWebpackConfigPath } from './utils/getWebpackConfigPath';
+import Os from "os";
```
wodin, line 18: Quotes inconsistent with the existing code

piccinnigius:
@wodin thanks for the feedback, how can i apply all this changes without change settings in my ide? i use intellij

sorry but i am a newbe in pull requests

wodin:
@piccinnigius hmmm... sorry, I don't know IntelliJ.

Also, I am not in any way involved in this project. I am just watching it. But in general with PRs it's best to avoid changing spacing/quoting etc. at the same time as making other changes, since it obscures the real changes.

I see the project does not have a `.prettierrc` (or equivalent), but there is a [`.editorconfig`](https://github.com/callstack/repack/blob/d81c4992dc902eae0e89849a882e15056c08a61b/.editorconfig#L3-L10) which defines some things (but not the quoting style). Maybe IntelliJ can be made to adhere to `.editorconfig`?

piccinnigius:
@wodin thanks a lot for infos. I'll find out. You know how to resubmit the same pull request with changes, or is better to close this and make a new one?

wodin:
@piccinnigius I'm not sure what the norm is in this project, but no need to close this PR and open a new one. You can either just push more commits to the same branch to fix the issues, or you can rebase the branch with the changes to make it look like the original commits never existed, and then force push the branch.

If I look at a random merged PR it looks like they just fixed the linting issues in a new commit and didn't bother rebasing.
https://github.com/callstack/repack/pull/256/commits
But I didn't check other PRs to see if this is the norm. If you just add new commits to fix the issues, you can always rebase later before the PR is merged if the maintainers want you to do that.

By the way, I see there's [`lint` script in `package.json`](https://github.com/callstack/repack/blob/d81c4992dc902eae0e89849a882e15056c08a61b/package.json#L17), so you should at least be able to check the linting locally by running `yarn lint`.

piccinnigius:
fixed linting problems

edisonpaul4:
@wodin Any news regarding this PR?

We want to start using repack but is a deal breaker that repack doesnt work correctly on windows.

packages/repack/src/commands/bundle.ts
```diff
@@ -23,10 +24,16 @@ export async function bundle(
...
 ) {
-  const webpackConfigPath = getWebpackConfigPath(
+  let webpackConfigPath = getWebpackConfigPath(
     config.root,
     args.webpackConfig
   );
+
+  if (webpackConfigPath.endsWith('.mjs') && Os.platform() === 'win32'){
+    webpackConfigPath = 'file:\\' + webpackConfigPath;
+    args.webpackConfig = 'file:\\' + args.webpackConfig;
```
thymikee, line 34: rewriting `args.webpackConfig` is quite unexpected. Why do we need to do that? You can pass an absolute path to the CLI argument yourself.

Also `getWebpackConfigPath` should be responsible for providing us correct `webpackConfigPath` for any platform, so we don't need to override it (and repeat the code in `start` command). Appending the `file:\\` doesn't sound like an elegant solution here. Can you think about better way to achieve the correct path for Windows?

piccinnigius: I know that this is not the best solution. I'll try to figure out, any hint?

thymikee: @RafikiTiki I'd appreciate your help here

piccinnigius:
hi @RafikiTiki, i've managed to reduce edits in multiple parts of repack to make it works by editing only getWebpackConfigPath and the dirname in tester app config. The last bad thing still the  need of `file://` prepending. do you have any idea how to avoid this platform specific issue? thanks in advance

RafikiTiki (approved):
Big thanks for your contribution @piccinnigius, and sorry it took me so long to finally review.

I recently got my hands on an old Windows machine, and (with some issues, but not related to the scope of this PR) I was able to run your branch successfully. I've also double-checked if this didn't have an impact on Mac users, and unsurprisingly – it didn't!

I'm merging it now, we plan to create a new release with @jbroma soon 🚀 

Edit: this probably can fix #440 #432 #415 #340 and #215

piccinnigius:
> Big thanks for your contribution @piccinnigius, and sorry it took me so long to finally review.
> 
> I recently got my hands on an old Windows machine, and (with some issues, but not related to the scope of this PR) I was able to run your branch successfully. I've also double-checked if this didn't have an impact on Mac users, and unsurprisingly – it didn't!
> 
> I'm merging it now, we plan to create a new release with @jbroma soon 🚀

Thanks so much, i'm glad it works and to be part of this very usefull project <3

PR #344, opened 2023-04-17
Chore: migrate repack native modules to turbo modules

teneeto:
### Summary
This PR aims to enable newArch and make `re.pack `native modules to `turbo modules `while ensuring newArch is implemented with backward compatibility.

`re.pack` current implementation directly uses RN native modules, but with this change, there is a `SPEC` implementation that allows for more efficient communication between native and JavaScript. This change allows for both old and new architecture.

### Test plan

- Make sure you are currently using the newArch and the setup conforms to [React Native Doc directives](https://reactnative.dev/docs/next/new-architecture-app-intro) which also means you're on **`RN>=0.68.0`**

NOTE: TesterApp currently meets this requirement if you're using repack TesterApp.

 #### For Android
- You will only need to update your `android/gradle.properties` file as follows:
**`newArchEnabled=true`**

-  and then **`yarn TesterApp:start`**

#### For iOS
- Run pod install with the flags: 
**`RCT_NEW_ARCH_ENABLED=1 pod install`**

-  and then  **`yarn TesterApp:start`**

### Screen Records
- **newArch - iOS**
----

https://user-images.githubusercontent.com/22779249/232512294-4ada8342-949c-41ab-811f-6fc7a1b2fe52.mov

----
- **oldArch - iOS**
----

https://user-images.githubusercontent.com/22779249/232516328-1b7c5a0e-82ba-44d7-94e6-7367e206e8a4.mov

----
- **newArch - Android**
----

https://user-images.githubusercontent.com/22779249/232515374-b566a38e-538b-482d-bb94-35fc09483043.mov

----
- **oldArch - android**
----

https://user-images.githubusercontent.com/22779249/232516407-e830662d-b5d2-46e6-a2b3-a2f604618f0e.mov


README.md
```diff
@@ -66,6 +66,28 @@ Available at: [Projects](https://github.com/callstack/repack/projects?type=class
 - [ ] Code signing and verification
 - [ ] `webpack-init` command
 
+### New Architecture Support (Turbo Module)
```
troZee, line 69:
```suggestion
### New Architecture Support
```

packages/repack/android/src/newarch/java/com/callstack/repack/ChunkManagerModule.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class ScriptManagerModule(reactContext: ReactApplicationContext) : NativeScriptManagerSpec(reactContext) {
+    private val remoteLoader: RemoteScriptLoader = RemoteScriptLoader(reactApplicationContext)
+    private val fileSystemLoader: FileSystemScriptLoader = FileSystemScriptLoader(reactApplicationContext)
+
+    private var implementation: ScriptManagerModuleImpl = ScriptManagerModuleImpl()
+
+    override fun getName(): String = ScriptManagerModuleImpl.NAME
+
+    override fun loadScript(scriptId: String, configMap: ReadableMap, promise: Promise) {
```
troZee, line 14: Could you check, if this method is called on the TS side for the new arch? AFAIR `@ReactMethod` is also needed for the new arch as well.

packages/repack/ios/ScriptManager.mm
```diff
@@ -262,4 +262,114 @@ - (void)runInBackground:(void(^)())callback
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), callback);
 }
 
+// We won't compile this code when we build for the old architecture.
```
troZee, line 265: Could you remove that comment?

packages/repack/js/index.ts
```diff
@@ -0,0 +1,6 @@
+/* 
+    TurboModuleRegistry.get taps into the old Native Modules API under the hood,
+    we need to re-export our module, to avoid registering it multiple times.
+*/
+
+export default require('./NativeScriptManager').default;
```
troZee, line 6: Does it work for the old arch as well? 
AFAIR I did a check for that 

```
const isTurboModuleEnabled = global.__turboModuleProxy != null;
```

packages/repack/js/NativeScriptManager.ts
```diff
@@ -0,0 +1,12 @@
+import type {TurboModule} from 'react-native';
+import {TurboModuleRegistry} from 'react-native';
+
+export interface Spec extends TurboModule {
+    readonly getConstants: () => {};
+
```
troZee, line 6: unnecessary new line

packages/repack/js/NativeScriptManager.ts
```diff
@@ -0,0 +1,12 @@
...
+    loadScript(scriptId: string, config: Object): Promise<Object>;
```
troZee, line 7: Please use a proper type instead of `Object`

packages/repack/js/NativeScriptManager.ts
```diff
@@ -0,0 +1,12 @@
...
+    prefetchScript(scriptId: string, config: Object): Promise<Object>;
```
troZee, line 8: Please use a proper type instead of `Object`

packages/repack/js/NativeScriptManager.ts
```diff
@@ -0,0 +1,12 @@
...
+    invalidateScripts(scripts: Array<string>): Promise<null>;
```
troZee, line 9: Why promise null? Should this function return something? If not, it should not be async

packages/repack/tsconfig.json
```diff
@@ -69,6 +69,6 @@
     "skipLibCheck": true,                           /* Skip type checking of declaration files. */
     "forceConsistentCasingInFileNames": true        /* Disallow inconsistently-cased references to the same file. */
   },  
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "js/NativeScriptManager.ts"],
```
troZee, line 72: Why `NativeScriptManager` is added as a separate file to the config? We should include all ts files in the config

troZee (changes requested):

packages/repack/js/NativeScriptManager.ts
```diff
@@ -0,0 +1,12 @@
+import type {TurboModule} from 'react-native';
+import {TurboModuleRegistry} from 'react-native';
+
+export interface Spec extends TurboModule {
```
troZee, line 4: @jbinda Should we consider implementing sync API alongside the async one? eg 
```
    loadScript(scriptId: string, config: Object): Promise<Object>;
    loadScriptSync(scriptId: string, config: Object): Object
```

troZee:
Overall it looks good 🎉  I have added a few comments 

cc @thymikee @jbinda

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModuleImpl.kt
```diff
@@ -0,0 +1,91 @@
...
+    private fun runInBackground(fn: () -> Unit) {
+        val handler = Handler()
+        val runnable = Runnable {
+            fn()
+        }
+        handler.postDelayed(runnable, 0)
+
+    }
+
+    fun loadScript(scriptId: String, configMap: ReadableMap, promise: Promise, remoteLoader: RemoteScriptLoader, fileSystemLoader: FileSystemScriptLoader) {
```
troZee, line 20: [nit]: as an improvement, we can consider rewriting it to follow the bob pattern: 
https://github.com/callstack/react-native-builder-bob/tree/main/packages/create-react-native-library/templates/java-library-mixed/android/src

cc @thymikee

packages/repack/js/NativeScriptManager.ts
teneeto: as described in this [TypeScript to Native Type Mapping](https://reactnative.dev/docs/next/new-architecture-appendix#iii-typescript-to-native-type-mapping) doc: `Object` generates the equiv of `ReadableMap` type on Kotlin and `NSDictionary` on objc. and the definitions of these types weren't explicit, but it did generate as needed.

packages/repack/js/NativeScriptManager.ts
teneeto: A Promise is needed on JS side to generate `resolve` and` reject `on objc and a `Promise` for kotlin which is included in the `invalidateScripts ` func, Although I skipped changing this to string. I will fix it now. Thanks.

packages/repack/android/src/newarch/java/com/callstack/repack/ChunkManagerModule.kt
teneeto: AFAIK, `@ReactMethod` isn't needed for the newArch, it instead uses the `override`

packages/repack/js/NativeScriptManager.ts
teneeto: as described in this [TypeScript to Native Type Mapping](https://reactnative.dev/docs/next/new-architecture-appendix#iii-typescript-to-native-type-mapping) doc: `Object` generates the equiv of `ReadableMap` type on Kotlin and `NSDictionary` on objc. and the definitions of these types weren't explicit, but it did generate as needed.

packages/repack/tsconfig.json
teneeto: Will fix. Thanks

packages/repack/js/NativeScriptManager.ts
teneeto: Will fix. Thanks.

README.md
teneeto: Will fix. Thanks.

packages/repack/js/index.ts
teneeto: Yes, this works and as confirmation I checked how modules are called on [react native svg](https://github.com/software-mansion/react-native-svg/blob/d31ffdb762d34171fd934ac97b031ccf5e0b645d/src/elements/Svg.tsx#L96)

README.md
```diff
@@ -66,6 +66,28 @@ Available at: [Projects](https://github.com/callstack/repack/projects?type=class
 - [ ] Code signing and verification
 - [ ] `webpack-init` command
 
+### New Architecture Support
+----
+This library supports new architecture!. [turbo module](https://reactnative.dev/docs/next/the-new-architecture/pillars-turbomodules) offers a new architecture for initializing native modules.
```
okwasniewski, line 71:
```suggestion
This library supports new architecture! Using [turbo modules](https://reactnative.dev/docs/next/the-new-architecture/pillars-turbomodules) offers a new way of initializing native modules.
```

packages/repack/android/src/newarch/java/com/callstack/repack/ChunkManagerModule.kt
```diff
@@ -0,0 +1,25 @@
+package com.callstack.repack
+
+import android.os.Handler
+import com.facebook.react.bridge.*
+
+class ScriptManagerModule(reactContext: ReactApplicationContext) : NativeScriptManagerSpec(reactContext) {
+    private val remoteLoader: RemoteScriptLoader = RemoteScriptLoader(reactApplicationContext)
+    private val fileSystemLoader: FileSystemScriptLoader = FileSystemScriptLoader(reactApplicationContext)
+
+    private var implementation: ScriptManagerModuleImpl = ScriptManagerModuleImpl()
```
okwasniewski, line 10: Why do we need to initialize the shared implementation class? I think a better way to do this is just calling methods on it, same as pager-view does: https://github.com/callstack/react-native-pager-view/blob/16173bd3222305eebc5c77bf326aab448c712717/android/src/paper/java/com/reactnativepagerview/PagerViewViewManager.kt#L79

teneeto: Fixed.

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt
```diff
@@ -1,17 +1,30 @@
...
-    }
-
-    override fun createViewManagers(reactContext: ReactApplicationContext): List<ViewManager<*, *>> {
-        return emptyList()
-    }
+ override fun getReactModuleInfoProvider() = ReactModuleInfoProvider {
+    val isTurboModule: Boolean = BuildConfig.IS_NEW_ARCHITECTURE_ENABLED
+   mapOf(
+    ScriptManagerModuleImpl.NAME to ReactModuleInfo(
+        ScriptManagerModuleImpl.NAME, // name
```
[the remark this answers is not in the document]
teneeto: Fixed.

packages/repack/tsconfig.json
teneeto: Fixed.

packages/repack/js/NativeScriptManager.ts
teneeto: Fixed.

packages/repack/js/NativeScriptManager.ts
teneeto: Fixed

packages/repack/js/NativeScriptManager.ts
```diff
@@ -0,0 +1,47 @@
...
+type NormalizedScriptLocator = {
+    method: string;
+    url: string;
+    fetch: boolean; 
+    timeout: Int32;
+    absolute: boolean;
+    query?: string;
+    headers?: string;
+    body?: string;
+    token?: string;
```
RafikiTiki, line 19: `token` was removed entirely in recent update to CodeSigning: https://github.com/callstack/repack/pull/348/files#diff-6e569776db8e37a5482958fa2218c1d9fa5353c38c9de03df173a5b830bbce36L20-L21

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt
```diff
@@ -1,17 +1,30 @@
...
-    override fun createViewManagers(reactContext: ReactApplicationContext): List<ViewManager<*, *>> {
-        return emptyList()
+    override fun getReactModuleInfoProvider() = ReactModuleInfoProvider {
+        val isTurboModule: Boolean = BuildConfig.IS_NEW_ARCHITECTURE_ENABLED
+        mapOf(
+                ScriptManagerModuleImpl.NAME to ReactModuleInfo(
+                        ScriptManagerModuleImpl.NAME, // name
+                        ScriptManagerModuleImpl.NAME, // className
+                        false, // canOverrideExistingModule
+                        false, // needsEagerInit
```
RafikiTiki, line 23: I don't know what exactly `needsEagerInit` means, but I'd say Repack should be initialized & available as soon as possible, especially in the Module Federation architecture we're now treating as a recommended one.

Take a look at the Super App Showcase Host App – https://github.com/callstack/super-app-showcase/blob/main/packages/host/index.js ScriptManager is used before the App component is registered.

Of course, what I'm saying here might be completely wrong since I'm not that familiar with the New Arch – if that's the case, sorry for the confusion :)

README.md
```diff
@@ -56,6 +56,37 @@ Re.Pack uses Webpack 5 and React Native CLI's plugin system to allow you to bund
 - [x] Code signing and verification
 - [x] `@callstack/repack-init` - tool that automates the integration of the Re.Pack into React-Native projects
 
+### Planned features
+
+Available at: [Projects](https://github.com/callstack/repack/projects?type=classic)
+
+- [ ] `ChunksToHermesBytecodePlugin` plugin to automatically transform async chunks to bytecode format.
+- [ ] Code signing and verification
+- [ ] `webpack-init` command
```
RafikiTiki, line 65: This whole section is irrelevant – these features are already implemented & released. Either this branch needs sync with `main,` or conflicts were resolved in a wrong way

RafikiTiki (commented):
I don't have much experience when it comes to New Architecture so if @troZee says it's looking good from the configuration perspective I trust his judgment :)

However, one thing that I don't like is that the native implementation of the `prefetchScript`, `loadScript`, and `invalidateScripts` methods is duplicated, thus making the maintenance cost higher in the future. Is this expected when writing backward-compatible modules? That would be a weird design TBH.

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt
teneeto: On repack side, i'm not sure how eager it needs to be initialised (even though i feel it's still initialised and available), however this is the default spec from docs and i'm sure it can be set to Eagerly initialise.

teneeto:
> I don't have much experience when it comes to New Architecture so if @troZee says it's looking good from the configuration perspective I trust his judgment :)
> 
> However, one thing that I don't like is that the native implementation of the `prefetchScript`, `loadScript`, and `invalidateScripts` methods is duplicated, thus making the maintenance cost higher in the future. Is this expected when writing backward-compatible modules? That would be a weird design TBH.

Yes, that's the implementation for backward compatibility. There's a common module with the actual implementation of `prefetchScript`, `loadScript`, and `invalidateScripts` within the `ChunkManagerModuleImpl.kt` file. The only difference between the old and new arch is how these methods are exposed, where the new arch uses `override` for methods, the old arch uses the `@ReactMethod`. And that's the only time we use the old or new arch files.

karimb11:
Hi, any ETA on this improvement? Thanks!

jbroma:
Closing in favour of https://github.com/callstack/repack/pull/495

jbroma:
Thank you @teneeto for your all work on this PR 🙏

PR #348, opened 2023-04-24
feat: code-signing integrated tokens

jbroma:
### Summary

This PR integrates the the tokens generated by `CodeSigningPlugin` to the bundles. This removes the need for matching the tokens with the bundles in the `ScriptManager` setup.

### Description

By embedding the tokens directly into the bundles we can eliminate the need of creating `tokenResolver` and fetching the tokens separately. After downloading the bundle, RePack will now evaluate whether the bundle is signed or not, extract the token and perform verification (if enabled). This solution comes with one tradeoff

The behaviour of `verifyScriptSignature` flag (as well as it's type) was changed. From now on, by default, the flag is set to `none`, which means no code-signature verification will take place. You can set the flag to one of the following string literals to control the verification process: `strict`, `lax`, or `off`.

| Value   | Description |
|---------------|-------------|
| `strict` | The script's code-signature will be verified regardless of the signature being present in the bundle. |
| `lax`     | The script's code-signature will be verified only when the signature is present in the bundle. If the signature is not present in the bundle, the script will be loaded without verification. |
| `off`     | The script's code-signature will not be verified. |

This solution is also compatible with `.hbc` bundles produced by hermes compiler.

### Usage

1. Generate a pair of private & public RS256 keys that will be used for code-signing (like [this](https://gist.github.com/ygotthilf/baa58da5c3dd1f69fae9))
2. Add `Repack.CodeSigningPlugin` to the list of plugins in `webpack.config` and configure it with path to the private key and name of the output file.
```js
new Repack.plugins.CodeSigningPlugin({
        privateKeyPath: './private.pem',
        outputPath: path.join(dirname, 'build/outputs', platform, 'remote'),
        excludedChunks: [ 'local_chunk' ]
}),
```
3. Configure the ScriptManager. Inside of `ScriptManager.shared.addResolver` do the following:
  a. Add `verifyScriptSignature` key set to `true` and pass the resolved `token` from the previous step to the config returned from `ScriptManager.shared.addResolver`
```js
ScriptManager.shared.addResolver(async (scriptId, caller) => {
  // ...
  
  let url;
  let token;
  if (caller === 'main') {
    url = Script.getDevServerURL(scriptId);
  } else {
    url = resolveURL(scriptId, caller);
  }
  
  if (!url) {
    return undefined;
  }
  
  return {
    url,
    verifyScriptSignature: 'strict', // or 'lax' or 'off'
    query: { platform: Platform.OS },
  };
});
```

### Test plan

- UI tests
- tested on super-app-showcase

### DIFF
--- a/packages/repack/android/src/main/java/com/callstack/repack/CodeSigningUtils.kt
+++ b/packages/repack/android/src/main/java/com/callstack/repack/CodeSigningUtils.kt
@@ -6,6 +6,7 @@ import com.nimbusds.jose.JWSVerifier
 import com.nimbusds.jose.crypto.RSASSAVerifier
 import com.nimbusds.jwt.SignedJWT
 import java.math.BigInteger
+import java.nio.charset.Charset
 import java.security.KeyFactory
 import java.security.MessageDigest
 import java.security.NoSuchAlgorithmException
@@ -16,7 +17,7 @@ import java.security.spec.X509EncodedKeySpec
 
 class CodeSigningUtils {
     companion object {
-        private fun getHash(string: String): ByteArray? {
+        private fun getHash(bytes: ByteArray): ByteArray? {
             var digest: MessageDigest? = null
             try {
                 digest = MessageDigest.getInstance("SHA-256")
@@ -24,14 +25,14 @@ class CodeSigningUtils {
                 e1.printStackTrace()
             }
             digest?.reset()
-            return digest?.digest(string.toByteArray())
+            return digest?.digest(bytes)
         }
 
         private fun bin2hex(data: ByteArray): String {
             return java.lang.String.format("%0" + data.size * 2 + "x", BigInteger(1, data))
         }
 
-        private fun computeHash(content: String?): String? {
+        private fun computeHash(content: ByteArray?): String? {
             if (content == null) {
                 return null
             }
@@ -82,7 +83,7 @@ class CodeSigningUtils {
             return null
         }
 
-        fun verifyBundle(context: Context, token: String?, fileContent: String?) {
+        fun verifyBundle(context: Context, token: String?, fileContent: ByteArray?) {
             if (token == null) {
                 throw Exception("The bundle verification failed because no token for the bundle was found.")
             }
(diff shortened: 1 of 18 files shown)


packages/repack/android/src/main/java/com/callstack/repack/CodeSigningUtils.kt
```diff
@@ -104,5 +105,27 @@ class CodeSigningUtils {
                 throw Exception("The bundle verification failed because the bundle hash is invalid.")
             }
         }
+
+        fun extractBundleAndToken(fileContent: ByteArray): Pair<ByteArray, String?> {
+            val signatureSize = 1280
+            val startingSequence = "/* RCSSB */"
```
RafikiTiki, line 111: Please, for the sake of future generations, add a comment what the RCSSB stands for :)

packages/repack/src/webpack/plugins/__tests__/CodeSigningPlugin.test.ts
```diff
@@ -59,171 +55,175 @@ const injectThisCompilationHookMock = (
...
+    expect(bundle1.byteLength).toBeGreaterThan(1280);
+
+    const bundle2 = (fs.writeFile as jest.Mock).mock.calls[1][1] as Buffer;
+    expect(bundle2.toString().match(BUNDLE_WITH_JWT_REGEX)).toBeTruthy();
+    expect(bundle2.byteLength).toBeGreaterThan(1280);
   });
 
-  it('produces code-signing-mapping file with valid JWTs', () => {
-    const assets = {
+  it('produces code-signing-mapping file with valid JWTs', async () => {
```
RafikiTiki, line 113:
```suggestion
  it('produces code-signed bundles with valid JWTs', async () => {
```

RafikiTiki: NIT

RafikiTiki (approved):
Tested it locally, everything works as expected. Great work @jbroma 🚀

DivyatejaChe:
HI @jbroma  how did you resolve the token in the latest implementation of Code Signing?

PR #356, opened 2023-05-18
feat: repack-init

jbroma:
### Summary

This PR introduces `repack-init`, a utility npm package that automates the process of integrating the `Repack` bundler into React Native projects.

Here's a summary of the functionality `repack-init` provides:

- **Package Manager Detection:** Identifies the package manager (npm or yarn) being used in the project.
- **React Native Verification:** Checks whether React Native is installed in the project, and what version is being used.
- **Dependency Addition:** Utilizes the established package manager to add necessary dependencies to the React-Native project.
- **Configuration of React Native Config:** Checks for the existence of `react-native.config.js` and adds or modifies the necessary fields (commands).
- **Webpack Configuration:** Creates and configures `webpack.config.js` based on user-provided flags (--mjs or --cjs).
- **iOS Modification:** Updates the `project.pbxproj` file in the iOS directory.
- **Android Modification:** Updates the `build.gradle` file in the Android directory based on specific React Native version requirements.

This tool aims to simplify the setup of `Repack` in React Native projects, ensuring developers can quickly start using `Repack`'s features.

### Limitations:

Minimum required `node` version: 16

### Usage
```
Usage: @callstack/repack-init [options]

Options:
  -e, --entry    Path to main entry point for the React-Native project
                                                  [string] [default: "index.js"]
  -f, --format   Format of the webpack.config file
                               [string] [choices: "mjs", "cjs"] [default: "mjs"]
  -v, --verbose  Enables verbose logging              [boolean] [default: false]
      --version  Show version number                                   [boolean]
      --help     Show help                                             [boolean]
```

### Example outputs
```
✔  Dependencies installed
✔  Created webpack.config.mjs from template
✔  Created react-native.config.js
✔  Added "webpack-bundle" as bundleCommand to android/app/build.gradle
✔  Added "webpack-bundle" as BUNDLE_COMMAND to build phase shellScript in ios/test1.xcodeproj/project.pbxproj

🎉 Setup complete. Thanks for using Re.Pack!
```

### DIFF
--- a/packages/init/README.md
+++ b/packages/init/README.md
@@ -1,0 +1,32 @@
+<p align="center">
+  <img src="https://raw.githubusercontent.com/callstack/repack/HEAD/logo.png">
+</p>
+<p align="center">
+A Webpack-based toolkit to build your React Native application with full support of Webpack ecosystem.
+</p>
+
+---
+
+[![Build Status][build-badge]][build]
+[![Version][version-badge]][version]
+[![MIT License][license-badge]][license]
+[![PRs Welcome][prs-welcome-badge]][prs-welcome]
+[![Code of Conduct][coc-badge]][coc]
+
+`@callstack/repack-init` is a tool that automates the integration of the `@callstack/repack` into React-Native projects`.
+
+Check out our website at https://re-pack.netlify.app/ for more info and documentation or out GitHub: https://github.com/callstack/repack.
+
+<!-- badges -->
+
+[callstack-readme-with-love]: https://callstack.com/?utm_source=github.com&utm_medium=referral&utm_campaign=react-native-paper&utm_term=readme-with-love
+[build-badge]: https://img.shields.io/github/workflow/status/callstack/repack/CI/main?style=flat-square
+[build]: https://github.com/callstack/repack/actions/workflows/main.yml
+[version-badge]: https://img.shields.io/npm/v/@callstack/repack-debugger-app?style=flat-square
+[version]: https://www.npmjs.com/package/@callstack/repack-debugger-app
+[license-badge]: https://img.shields.io/npm/l/@callstack/repack-debugger-app?style=flat-square
+[license]: https://github.com/callstack/repack/blob/master/LICENSE
+[prs-welcome-badge]: https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square
+[prs-welcome]: ./CONTRIBUTING.md
+[coc-badge]: https://img.shields.io/badge/code%20of-conduct-ff69b4.svg?style=flat-square
+[coc]: https://github.com/callstack/repack/blob/master/CODE_OF_CONDUCT.md
--- a/packages/init/README.md
+++ b/packages/init/README.md
@@ -15,6 +15,26 @@ A Webpack-based toolkit to build your React Native application with full support
 
 `@callstack/repack-init` is a tool that automates the integration of the `@callstack/repack` into React-Native projects`.
 
+## Usage
+
+```sh
+npx @callstack/repack-init [options]
+```
+
+## Options
+
+- `-e, --entry` Path to main entry point for the React-Native project. Defaults to: `index.js`
+
+- `-f, --format` Format of the webpack.config file. Available choices: `"mjs"`, `"cjs"` Defaults to: `"mjs"`
+
+- `-v, --verbose` Enables verbose logging. Defaults to: `false`
+
+- `--version` Show version number.
+
+- `--help` Show help.
+
+---
+
 Check out our website at https://re-pack.netlify.app/ for more info and documentation or out GitHub: https://github.com/callstack/repack.
 
 <!-- badges -->
--- a/packages/init/src/tasks/addDependencies.ts
+++ b/packages/init/src/tasks/addDependencies.ts
@@ -1,3 +1,4 @@
+import type { PM } from 'detect-package-manager';
 import { execa } from 'execa';
 import ora from 'ora';
 
@@ -13,7 +14,7 @@ const dependencies = [
  *
  * @param packageManager yarn or npm
  */
-export default async function addDependencies(packageManager: 'yarn' | 'npm') {
+export default async function addDependencies(packageManager: PM) {
   let installCommand: string;
 
   if (packageManager === 'yarn') {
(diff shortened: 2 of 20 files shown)


packages/init/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+---
+
+[![Build Status][build-badge]][build]
+[![Version][version-badge]][version]
+[![MIT License][license-badge]][license]
+[![PRs Welcome][prs-welcome-badge]][prs-welcome]
+[![Code of Conduct][coc-badge]][coc]
+
+`@callstack/repack-init` is a tool that automates the integration of the `@callstack/repack` into React-Native projects`.
```
RafikiTiki, line 16: Let's add basic usage section to the Readme

packages/init/src/tasks/addDependencies.ts
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    spinner = ora(
+      ` Installing Re.Pack dependencies using ${packageManager}`
+    ).start();
+    await execa(command, { stdio: 'pipe', shell: true });
+    spinner.stop();
+    logger.success('Dependencies installed');
+  } catch (error) {
+    spinner?.stop();
+    logger.error(`Failed to install Re.Pack dependencies`);
```
TMisiukiewicz, line 40: as an alternative, you can also use `spinner.succeed()` and `spinner.fail()`, IMO it looks better than using another logger but it's up to you :)

packages/init/src/tasks/addDependencies.ts
```diff
@@ -0,0 +1,43 @@
...
+export default async function addDependencies(packageManager: 'yarn' | 'npm') {
+  let installCommand: string;
+
+  if (packageManager === 'yarn') {
+    installCommand = 'add';
+  } else {
+    installCommand = 'install';
+  }
+  const deps = dependencies.join(' ');
+  const command = `${packageManager} ${installCommand} -D ${deps}`;
```
RafikiTiki, line 27: Question: does `-D` flag work for npm?

packages/init/src/tasks/checkReactNative.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Checks whether react-native is installed and returns it's version
+ *
+ * @param cwd
+ * @returns React-Native version
+ */
+export default function checkReactNative(rootDir: string): string {
+  const packageJsonPath = path.join(rootDir, 'package.json');
+  const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
```
TMisiukiewicz, line 15: maybe it's also worth checking if package.json exists first, and wrapping JSON.parse in try/catch statement since it may throw an error if JSON file is malformed

packages/init/src/tasks/checkPackageManager.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export default function checkPackageManager(cwd: string) {
+  const yarnLockPath = isProjectUsingYarn(cwd);
+
+  if (yarnLockPath) {
+    logger.info('Using yarn as package manager');
+    return 'yarn';
+  }
+  logger.info('Using npm as package manager');
+  return 'npm';
```
RafikiTiki, line 23: PNPM is another possibility here, we can add an additional check for it:

```ts
findUpSync('pnpm-lock.yaml', {cwd})
```

Or we can use an existing package to detect package manager (first hit from google, have a lot of downloads but last release was almost 2 years ago): https://www.npmjs.com/package/detect-package-manager

packages/init/src/tasks/addDependencies.ts
jbroma: yes, it's mentioned in the docs: https://docs.npmjs.com/cli/v8/commands/npm-install

packages/init/src/tasks/checkReactNative.ts
jbroma: This will be caught by try/catch clause in the `index.js`, displaying the error from `fs.readFileSync` there.

packages/init/src/tasks/checkPackageManager.ts
jbroma: I've kind of omitted `pnpm` as it doesn't work well in React-Native ecosystem (yet).

I'll check out the `detect-package-manager` and see whether it makes sense here. 
I've resorted to using method of detecting package manager almost identical to that of RN CLI.

jbroma: In theory if someone is using pnpm and it works for his usecase, I see no reason to not incldue it as one of the options and utilize that. It's needed only for the step of installing dependencies anyways

packages/init/src/tasks/modifyIOS.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  if (pbxShellScriptBuildPhaseIndex === -1) {
+    throw new Error(
+      'Could not find PBXShellScriptBuildPhase section in project.pbxproj\n' +
+        fallbackMessage
+    );
+  }
+
+  const shellScriptIndex = config.indexOf(
+    'shellScript =',
```
RafikiTiki, line 36: Are you sure the `Bundle React Native code and images` in which we must add the bundle command will always be the first in the `PBXShellScriptBuildPhase` section?

RafikiTiki: Wouldn't looking for `react-native-xcode.sh` and inserting `export BUNDLE_COMMAND=webpack-bundle` in a line above be safer?

jbroma: Great point about the order in that section! I'll make sure to include that. 

Regarding the 2nd comment - I would prefer to stick to the project itself instead of modifying something that comes from `node_modules`, which could potentially get overwritten with each `yarn install` in the project

TMisiukiewicz:
maybe it's worth checking [this](https://www.npmjs.com/package/xcode) lib for `pbxproj` parsing instead of relying on `indexOf`

packages/init/src/tasks/modifyIOS.ts
jbroma: Rewrote the whole modifyIOS part using `xcode` package for modyfing `project.pbxproj` files

RafikiTiki (approved):

RafikiTiki (approved):

PR #366, opened 2023-05-24
chore: bump node version on CI to 18 & 20

jbroma:
### Summary

Bumped node version on CI from `14` & `16` to `18` & `20`


thymikee (approved):
16 is technically still supported, so I'd add it

PR #368, opened 2023-05-26
fix(repack): keypress events not working after using inquirer

kerm1it:
### Summary

I used repack in my custom cli where it is using inquirer,  I find `keypress` events isn't working.

Finally, I see it in https://github.com/SBoudrias/Inquirer.js/issues/662 and add `process.stdin.resume()` to `bindKeypressInput` method to fix it.


kerm1it:
any progress ?

RafikiTiki:
Hey @kerm1it, sorry for a long time without a response – I somehow missed the notification. I'm unfamiliar with the `inquirer` library, and I want to make sure that the change you introduced, to make it work together with Repack, isn't causing regressions for people who are not using the `inquirer`. Did you check that? The issue you linked in the description doesn't provide a reason why this change helps, and additionally, [this comment](https://github.com/SBoudrias/Inquirer.js/issues/662#issuecomment-571344568) suggests calling the `process.stdin.resume()` after the inquirer usage and before other keypress listeners – so in your case this could potentially live in the custom CLI you're developing.

kerm1it:
In my case it works perfect，so I think we can keep it as it is for stability.

we can reopen it if anyone encounters this problem again.

jbroma:
This doesn't feel like something that should be included in Re.Pack. As @RafikiTiki mentioned, have you tried using `process.stdin.resume()` after the use of `inquirer` in your app?

RafikiTiki:
We're not using `inquirer` in any of our projects, and I feel this issue should be addresed in the CLI code rather than in the Repack.

Closing this for now, we might revisit it if there are more issues in the future that are not fixable by the end users.

PR #377, opened 2023-06-13
Add cli-progress to the repack cli

### LINKED ISSUE #369: I want to use progress instead of showing the bundle percentagein a separate line
I want to use progress instead of showing the bundle percentage in a separate line. The problem is that I can't find this number that I can use in progress.
Screenshot 1402-03-08 at 11 01 02

emel95:
Closes #369 

I've been using Repack for quite a while and although the functionalities of this project impressed me, I wasn't happy with the way progress of bundling process was shown in the terminal. So, I rolled up my sleeves and tried to make it neater and as informative as before for the developers who want to utilize this package.

The package that I used for this matter is "cli-progress" and changed the "consoleReporters.js" file. Feel free to change the way progress is shown on the Terminal by changing the presets bundled with the mentioned package and let me know if any changes from my side was required.

### DIFF
--- a/packages/repack/package.json
+++ b/packages/repack/package.json
@@ -60,6 +60,7 @@
   "dependencies": {
     "@callstack/repack-dev-server": "^1.0.1",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.7",
+    "cli-progress": "^3.12.0",
     "colorette": "^1.2.2",
     "dedent": "^0.7.0",
     "escape-string-regexp": "^4.0.0",
--- a/packages/repack/package.json
+++ b/packages/repack/package.json
@@ -91,6 +91,7 @@
     "@callstack/eslint-config": "^11.0.0",
     "@react-native-community/cli": "5.0.1-alpha.1",
     "@react-native-community/cli-types": "5.0.1-alpha.1",
+    "@types/cli-progress": "^3.11.0",
     "@types/dedent": "^0.7.0",
     "@types/fs-extra": "^9.0.11",
     "@types/jest": "^28.1.1",
--- a/packages/repack/src/logging/reporters/ConsoleReporter.ts
+++ b/packages/repack/src/logging/reporters/ConsoleReporter.ts
@@ -2,6 +2,7 @@ import util from 'util';
 import colorette from 'colorette';
 import throttle from 'lodash.throttle';
 import type { LogEntry, LogType, Reporter } from '../types';
+import cliProgress from 'cli-progress';
 
 export interface ConsoleReporterConfig {
   asJson?: boolean;
@@ -33,7 +34,7 @@ export class ConsoleReporter implements Reporter {
 }
 
 class JsonConsoleReporter implements Reporter {
-  constructor(private config: ConsoleReporterConfig) {}
+  constructor(private config: ConsoleReporterConfig) { }
 
   process(log: LogEntry) {
     console.log(JSON.stringify(log));
@@ -71,8 +72,14 @@ const FALLBACK_SYMBOLS: Record<LogType | 'progress', string> = {
 
 class InteractiveConsoleReporter implements Reporter {
   private requestBuffer: Record<string, Object> = {};
+  bar = new cliProgress.SingleBar({
+    format: 'Compiling {platform}: ' + '|{bar}|' + ' %{percentage}' + ' ',
+    hideCursor: true
+  }, cliProgress.Presets.shades_grey);
 
-  constructor(private config: ConsoleReporterConfig) {}
+  constructor(private config: ConsoleReporterConfig) {
+    this.bar.start(100, 0);
+   }
 
   process(log: LogEntry) {
     // Do not log anything in silent mode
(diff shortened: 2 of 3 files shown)


packages/repack/package.json
```diff
@@ -60,6 +60,7 @@
   "dependencies": {
     "@callstack/repack-dev-server": "^1.0.1",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.7",
+    "cli-progress": "^3.12.0",
```
szymonrybczak, line 63: Please install `@types/cli-progress` in `devDependencies` because the library code is not compiling.

packages/repack/src/logging/reporters/ConsoleReporter.ts
```diff
@@ -33,7 +34,7 @@ export class ConsoleReporter implements Reporter {
 }
 
 class JsonConsoleReporter implements Reporter {
-  constructor(private config: ConsoleReporterConfig) {}
+  constructor(private config: ConsoleReporterConfig) { }
```
szymonrybczak, line 37: Why did you make this change? Make sure to correctly setup eslint and prettier.

packages/repack/src/logging/reporters/ConsoleReporter.ts
```diff
@@ -94,8 +101,7 @@ class InteractiveConsoleReporter implements Reporter {
     const normalizedLog = this.normalizeLog(log);
     if (normalizedLog) {
       process.stdout.write(
-        `${
-          IS_SYMBOL_SUPPORTED ? SYMBOLS[log.type] : FALLBACK_SYMBOLS[log.type]
+        `${IS_SYMBOL_SUPPORTED ? SYMBOLS[log.type] : FALLBACK_SYMBOLS[log.type]
```
szymonrybczak, line 104: Same as above. Please make sure that you have correctly setup litner.

packages/repack/src/logging/reporters/ConsoleReporter.ts
```diff
@@ -71,8 +72,14 @@ const FALLBACK_SYMBOLS: Record<LogType | 'progress', string> = {
 
 class InteractiveConsoleReporter implements Reporter {
   private requestBuffer: Record<string, Object> = {};
+  bar = new cliProgress.SingleBar({
+    format: 'Compiling {platform}: ' + '|{bar}|' + ' %{percentage}' + ' ',
```
szymonrybczak, line 76: Make sure to handle the initial state. Right now when the progress is 0% there's label with the placeholder "Compiling **{platform}**". Please do not show this placeholder.
CleanShot 2023-06-15 at 14 09 09@2x

szymonrybczak:
💡 Do we want to have the progress bar in the same format as other logs? Because right know this progress bar is without any timestamp. cc. @RafikiTiki 
CleanShot 2023-06-15 at 14 16 06@2x

szymonrybczak:
@emel95 any updates? 🙏

emel95:
@szymonrybczak  
hi
I have fixed the issues you mentioned. I hope they are okay now. If there are any further problems, please let me know so that I can address them. Thank you.

jbroma:
Hi @emel95,

First of all, thanks for making an effort to contribute to Re.Pack!

I've reviewed the PR and found the following issues:
- to support bundling for both platforms at the same time, you probably want to use multi-bar instead of a single-bar, otherwise it will keep replacing itself and the output gets messy
- every log now has additional `\n` and there are lots of unnecessary empty lines
- we probably don't want to display the bar when there is no compilation taking place at the start and only start displaying it when the actual compilation is underway 
- the bar get's redrawn and the old one stays in place, would be nice to somehow fix this behaviour and remove the old bar before drawing the new one, and maybe keep it always at the bottom of the output from Re.Pack?

This is how it looked in my terminal:
image

jbroma (changes requested):
Hi @emel95,

First of all, thanks for making an effort to contribute to Re.Pack!

I've reviewed the PR and found the following issues:
- to support bundling for both platforms at the same time, you probably want to use multi-bar instead of a single-bar, otherwise it will keep replacing itself and the output gets messy
- every log now has additional `\n` and there are lots of unnecessary empty lines
- we probably don't want to display the bar when there is no compilation taking place at the start and only start displaying it when the actual compilation is underway 
- the bar get's redrawn and the old one stays in place, would be nice to somehow fix this behaviour and remove the old bar before drawing the new one, and maybe keep it always at the bottom of the output from Re.Pack?

This is how it looked in my terminal:
image

emel95:
hi @jbroma
Thank you for your complete explanation, I changed the things you said, please check again

emel95:
hi @jbroma
any updates?🙏

jbroma:
Hi @emel95,

some of the issues I mentioned above are still not resolved, try bundling 2 apps at once and see how the progress bars just keep overwriting themselves. 

P.S. I realise this not a trivial issue to fix

emel95:
hi
I will try to start these mentioned items in a few days and fix them as soon as possible

PR #378, opened 2023-06-16
Add ChunksToHermesBytecodePlugin

mikeduminy:
### Summary

At Klarna we've been using Re.pack for a while now, and wish to contribute back our implementation of a `ChunksToHermesBytecodePlugin`. This PR is based on an in-house plugin developed by @oblador.

The plugin itself re-implements tranformations found in each platform (which assume a single main chunk):

>For iOS the logic is copied from [react-native-xcode.sh](https://github.com/facebook/react-native/blob/f38fc9ba8681622f7cfdb586753e50c596946929/packages/react-native/scripts/react-native-xcode.sh#L166-L187)
>For Android the logic is copied from [BundleHermesCTask.kt](https://github.com/facebook/react-native/blob/f38fc9ba8681622f7cfdb586753e50c596946929/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/tasks/BundleHermesCTask.kt#L93-L111) (with defaults in [ReactExtension.kt](https://github.com/facebook/react-native/blob/f38fc9ba8681622f7cfdb586753e50c596946929/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/ReactExtension.kt#L116-L117))

> Side note: maybe these should be extracted into the react-native CLI, or scripts, instead of having separate implementations?

### Test plan

- [x] Add to TesterApp
- [x] Verify iOS (bundle, bundle and run in Release, start and run in Dev)
- [x] Verify Android (bundle, bundle and run in Release, start and run in Dev)
- [x] Verify it works as expected for the simple case

### ~~Open~~ Resolved* questions
1. [Resolved] Should this transformation happen inside the `OutputPlugin` (or `AssetProcessor`s)? 
    > - There's a lot of reusable logic in those and it feels like a more natural home.
    
    Resolution: No, the `OutputPlugin` is already doing too much.
2. [Resolved] Should we transform all chunks in this plugin, or ignore the main chunk (letting react-native transform that using hermes) and transform the rest?
    > - PRO: Letting react-native transform the main chunk means we don't have to recommend that people opt-out of this (less work, more maintainable)
    > - CON: It means special handling within Re.pack to ignore the main chunk and risks the transformation algorithm diverging over time (and composition of source maps). This might be okay since the main chunk is usually the most important to have transformed, so if our implementation gets out of sync with RN it might not be too bad.
    > - There is no super clean way of doing this on Android. See [comment](https://github.com/callstack/repack/pull/378/files#r1233363369) in code.
    
    Resolution: By default we let RN transform the main chunk but the API of the plugin should allow users to also transform the entry chunk if they want.
3. [Resolved] Alternative to (2): What if the `ChunksToHermesBytecodePlugin` accepted a `transformMainChunk: boolean` property?
    > - This would allow the most flexibility
    
    Resolution: See (2)

> Side note: maybe we should add more fine-grained control to the configuration `react-native` where you can opt out of the automatic hermes transformation of the bundle, but still use Hermes as the runtime.

TODO:
- [x] Resolve open questions
- [x] Figure out how to make it work with remote chunks
- [x] Add sensible defaults and messaging for misuse of the plugin
- [x] Add tests for the plugin
- [x] Update API docs
- [x] Update usage docs


mikeduminy:
I wonder if this should be added to the output plugin as an optional step. It needs to use a lot of the same logic to determine local vs remote chunks 🤔

packages/repack/src/webpack/plugins/OutputPlugin.ts
```diff
@@ -339,8 +339,8 @@ export class OutputPlugin implements WebpackPlugin {
             outputPath,
             bundleOutput: bundleFilename,
             bundleOutputDir: bundlePath,
-            sourcemapOutput: sourceMapFilename!,
-            assetsDest: assetsPath!,
```
mikeduminy, line 343: Removed these because code-smell 😁 (can revert this if needed)

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/composeSourceMaps.ts
```diff
@@ -0,0 +1,37 @@
+import fs from 'fs/promises';
+import path from 'path';
+import execa from 'execa';
+
+/**
+ * {@link composeSourceMaps} options.
```
mikeduminy, line 6: Not sure if I've documented this correctly. I would appreciate some input 🙏

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/ChunksToHermesBytecodePlugin.ts
```diff
@@ -0,0 +1,224 @@
...
+              ? bundleFilename
+              : path.join(
+                  platform === 'ios' ? assetsOutputDir : bundleOutputDir,
+                  assetPath
+                );
+
+            const bundleExists = await fileExists(bundleOutputPath);
+            if (!bundleExists) {
+              logger.error(`Bundle does not exist: ${bundleOutputPath}`);
+              // TODO: for now ignore this, consider throwing an error when remote
```
mikeduminy, line 175: We should solve running these transformations for remote assets. This is partially why I'd like to combine this logic with the asset processing that happens in the `OutputPlugin`.

For now this code just logs and skips the remote asset

packages/TesterApp/ios/TesterApp.xcodeproj/project.pbxproj
```diff
@@ -264,7 +264,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "set -e\n\nWITH_ENVIRONMENT=\"../node_modules/react-native/scripts/xcode/with-environment.sh\"\nexport BUNDLE_COMMAND=webpack-bundle\nREACT_NATIVE_XCODE=\"../node_modules/react-native/scripts/react-native-xcode.sh\"\n\n/bin/sh -c \"$WITH_ENVIRONMENT $REACT_NATIVE_XCODE\"\n";
+			shellScript = "set -e\n\n# hermes compilation is handled in webpack\nexport USE_HERMES=false\n\nWITH_ENVIRONMENT=\"../node_modules/react-native/scripts/xcode/with-environment.sh\"\nexport BUNDLE_COMMAND=webpack-bundle\nREACT_NATIVE_XCODE=\"../node_modules/react-native/scripts/react-native-xcode.sh\"\n\n/bin/sh -c \"$WITH_ENVIRONMENT $REACT_NATIVE_XCODE\"\n";
```
mikeduminy, line 267: Adds the following to the build phase:

```sh
# hermes compilation is handled in webpack
export USE_HERMES=false
```

This skips the main chunk hermes compilation performed in react-native-xcode.sh

packages/TesterApp/android/gradle.properties
```diff
@@ -41,4 +41,7 @@ newArchEnabled=false
 
 # Use this property to enable or disable the Hermes JS engine.
 # If set to false, you will be using JSC instead.
-hermesEnabled=true
+# NOTE: This is **NOT** set to opt-out of the automatic Hermes transformation
+# performed by the react gradle plugin. Instead, this transformation happens
+# during the webpack build.
+# hermesEnabled=false
```
mikeduminy, line 47: This happens to skip the main chunk hermes compilation performed in [BundleHermesCTask.kt](https://github.com/facebook/react-native/blob/main/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/tasks/BundleHermesCTask.kt#L92-L113). Setting this to `true` would enable it, and setting it to `false` would end up excluding the hermes runtime from the build.

Another option is setting this to `false` (matching the iOS change) and adding setting [enableSoCleanup](https://github.com/facebook/react-native/blob/main/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/ReactExtension.kt#L81) to `false`. This should result in a skipped transformation but allows us to still include the hermes runtime.

jbroma:
Hi @mikeduminy 

Regarding the open questions:
1. Let's keep the transformation in the separate plugin as it is because: 
- `OutputPlugin` is something we noticed adds some convenience for the user, but introduces a lot of complexity when releasing new plugins, in the future we are probably going to deprecate it and remove it from the `RePack plugin`.
- `AssetsProcessor` is something that deals with assets other than chunks so it's not the right place to do this transformation

2. I think we want to leave the responsibility of transforming the main chunk to the react-native itself. It's hard to only disable the transformation without disabling the hermes engine and it might be quite confusing for the end user. 

3. Good idea, only suggestion I might have is to name it 'entryChunk' instead of 'mainChunk', please tell me what you think

What I suggest we should do starting with this plugin (and apply the same pattern for the rest of the plugins in the future): 
1. Add `inputPath` option to tell the plugin where to look for input
2. Add `outputPath` option to tell the plugin where to place the transformed chunks. If the path is the same as `inputPath` it will overwrite the existing bundles with the new transformed ones.
3. Add `transformEntryChunk` optional option (by default set to `false`) or in most cases we are going to be only omitting `index.bundle` (the name can be obtained through webpack configuration object available inside the plugin) 
4. Add `tranformIgnore` optional option to provide a list of additional ignored paths

jbroma:
@mikeduminy 

take a look here and give this a shot, I had something like this in mind: https://github.com/callstack/repack/commit/fd060760f964a705fea9e1fab117800bc2892474

compare branches [here](https://github.com/mikeduminy/repack/compare/add-hermes-plugin...callstack:repack:feat/hermes-plugin)

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/transformBundleToHermesBytecode.ts
```diff
@@ -0,0 +1,37 @@
...
+    hermesCLIPath,
+    [
+      '-w', // Silence warnings else buffer overflows
+      '-O', // Enable optimizations
+      '-emit-binary',
+      '-out',
+      outputFile,
+      useSourceMaps ? '-output-source-map' : '',
+      filePath,
+    ].filter(Boolean)
```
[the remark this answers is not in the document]
mikeduminy: It filters out the empty string possibility on L27

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/transformBundleToHermesBytecode.ts
```diff
@@ -0,0 +1,37 @@
...
+  filePath: string;
+}
+
+export const transformBundleToHermesBytecode = async ({
+  hermesCLIPath,
+  useSourceMaps,
+  filePath,
+}: TransformBundleToHermesBytecodeOptions) => {
+  const outputFile = filePath + '.hbc';
+  const hermesSourceMapPath = filePath + '.hbc.map';
```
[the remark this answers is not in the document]
mikeduminy: This approach is lifted from [BundleHermesCTask.kt](https://github.com/facebook/react-native/blob/f38fc9ba8681622f7cfdb586753e50c596946929/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/tasks/BundleHermesCTask.kt#L93-L111) which should work on all platforms, but happy to change it if needed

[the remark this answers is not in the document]
oblador: It was a long time ago that I worked, but if memory serves me I tried to simplify and make this part less ugly but somehow that broke source maps. In the end I opted for doing eeeexactly like the RN scripts in terms of naming etc and only then it worked.

jbroma:
I've managed to use `assetEmitted` hook instead of `afterEmit` so it runs before `OutputPlugin` and simplifies the logic of the plugin greatly.

All the transformations are now done inside `build/generated` directory which is the source from which `OutputPlugin` copies the bundles to their target directories.

README.md
```diff
@@ -52,13 +53,12 @@ Re.Pack uses Webpack 5 and React Native CLI's plugin system to allow you to bund
   - `GET /api/:platform/assets` - assets emitted from compilation.
   - `GET /api/:platform/stats` - data about compilation.
 - [x] [Module Federation](https://medium.com/swlh/webpack-5-module-federation-a-game-changer-to-javascript-architecture-bcdd30e02669) support
+- [x] Code signing and verification
```
mikeduminy, line 56: I saw this was merged but not sure if you want it to be in the official feature list

jbroma:
Preview of the docs available here: https://6494804b67314a519583e33b--re-pack.netlify.app/docs/configuration/plugins/chunks-to-hermes

website/docs/configuration/plugins/chunks-to-hermes.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+To use `ChunksToHermesBytecodePlugin`, you need to instantiate it and add it to your webpack configuration's `plugins` array. Here's a simple example:
+
+```js
+// webpack.config.mjs
+import * as Repack from '@callstack/repack';
+
+// ...
+plugins: [
+  // Must appear after `Repack.RepackPlugin`
```
mikeduminy, line 44: This should no longer be necessary since we are using a different hook

website/docs/configuration/plugins/chunks-to-hermes.mdx
```diff
@@ -0,0 +1,59 @@
...
+  new Repack.ChunksToHermesBytecodePlugin({
+    enabled: mode === 'production',
```
mikeduminy, line 46: If users can ever run `webpack-serve` with mode set to production then we should add to this: `&& !devServer`

website/docs/configuration/plugins/chunks-to-hermes.mdx
jbroma: True, nice catch!

website/docs/configuration/plugins/chunks-to-hermes.mdx
jbroma: I think I've never used `webpack-serve`, could you describe this in more detail? It seems like advanced use-case for Re.Pack anyways. 

Although we might just replace `mode === production` with `!devServer`, how does that sound?

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/composeSourceMaps.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Removes original source map files.
+ */
+export const composeSourceMaps = async ({
+  reactNativePath,
+  packagerMapPath,
+  compilerMapPath,
+}: ComposeSourceMapsOptions) => {
+  const composedSourceMapPath = packagerMapPath + '.composed';
+
+  await execa('node', [
```
thymikee, line 35: There's no error handling for that exec call. We might want to handle it gracefully

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/getHermesCLIPath.ts
```diff
@@ -0,0 +1,30 @@
...
+  switch (os.platform()) {
+    case 'darwin':
+      return 'osx-bin';
+    case 'linux':
+      return 'linux64-bin';
+    case 'win32':
+      return 'win64-bin';
+    default:
+      throw new Error(
+        'OS not recognized. Please set hermesCLIPath to the path of a working Hermes compiler.'
```
thymikee, line 17: I have issues understanding this error message. The suggestion is to change hermesCLIPath but we only check on platform OS. Is this anyhow related?

jbroma: `hermesCLIPath` is an optional option for the plugin. By default the plugin tries to find the compiler on it's own

thymikee: yeah, but how is it related to `os.platform` failing to match supported dev platforms?

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/composeSourceMaps.ts
jbroma: if there is an error when executing that command,  `execa` should throw and that should stop the plugin altogether. Not sure if we want to handle it any other way tbh

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/getHermesCLIPath.ts
jbroma: I suppose we could move the error to the `getHermesCLIPath` function below

packages/repack/src/webpack/plugins/ChunksToHermesBytecodePlugin/utils/composeSourceMaps.ts
thymikee: What we do in RN CLI for example we like to add some extra context around the error. e.g. "Failed to compose source maps. Here you can find details: {error.toString()}". We also strip some stack traces, but that's not always great idea 😅

PR #382, opened 2023-06-20
fix: remote-assets url includes full publicPath now

jbroma:
### Summary

Fixed an issue for remote-assets where only the basename from `publicPath` was included.

Example of the problem:
if `publicPath` was set to
```sh
http://localhost:9999/remote-assets
```
then final URL would look like this:
```sh
# remote-assets is removed from the path
http://localhost:9999/assets/image.png
```
instead of
```sh
http://localhost:9999/remote-assets/assets/image.png
```

### DIFF
--- a/.changeset/sour-meals-cough.md
+++ b/.changeset/sour-meals-cough.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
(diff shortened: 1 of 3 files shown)


RafikiTiki (approved):
🚀

PR #401, opened 2023-07-12
fix: update nimbus-jose-jwt android dependency

jbroma:
### Summary

Upgraded `com.nimbusds:nimbus-jose-jwt` to the newest version which is not listed as having security vulnerabilities

### Test plan

Tested locally on the super-app-showcase which utilises code-signing extensively


RafikiTiki (approved):
LGTM! 🚀

PR #405, opened 2023-07-26
fix(repack-init): monorepo & version range support

jbroma:
### Summary

- fixed an issue where repack-init would fail to use proper package manager in monorepo environment
- fixed an issue where react-native version had a semver range like `^0.71.3` instead of a plain version, which caused the whole repack-init to fail with an error: `Invalid Version: ^0.71.3`.
- updated function definition docs

### Test plan

executable tested locally


RafikiTiki (approved):
Neat!

PR #409, opened 2023-07-31
feat: add enabled flag to CodeSigningPlugin

jbroma:
### Summary

Added `enabled` flag to `CodeSigningPlugin` to make it possible to disable the plugin when running in development environment. This flag defaults to `true` now to prevent a breaking change.

### Test plan

Added unit test for the flag


RafikiTiki (approved):
All good!

PR #421, opened 2023-09-20
feat(repack): builtin support for react-native deep imports for Module Federation

jbroma:
### Summary

**Problem Statement:**

In our current federated setup for React Native applications, a persistent issue emerged around the handling of shared dependencies, specifically with respect to deep imports from the 'react-native' library. These deep imports disrupted the efficient sharing of modules across federated boundaries. Furthermore, due to the challenges posed by these deep imports, our `assetsLoader` couldn't inline scaled assets in federated modules, leading to broken functionalities in some federated scenarios.

References:
[StackOverflow discussion on deep imports with Module Federation](https://stackoverflow.com/questions/65636979/wp5-module-federation-sharing-deep-imports)
[Webpack's resolveMatchedConfigs](https://github.com/webpack/webpack/blob/main/lib/sharing/resolveMatchedConfigs.js#L77-L79)

**Solution:**

This PR updates the `ModuleFederationPlugin` introducing mechanisms to better handle these shared dependencies, and by extension, fixing the `assetsLoader` issue.

**Changes Introduced:**

1. **Enhanced Configuration Options**:
   - Added `reactNativeDeepImports?: boolean` to `ModuleFederationPluginConfig` to allow the option to enable or disable adding React Native deep imports to shared dependencies.

2. **Deep Import Handling**:
   - Incorporated logic to adjust shared dependencies to account for deep imports from 'react-native' (e.g., `react-native/Libraries/Utilities/PixelRatio`). This was achieved using an undocumented feature of webpack's `ModuleFederationPlugin`.
   - Added a mechanism to ensure deep imports from the `react-native` library and packages from the `@react-native` group are correctly shared and resolved.

---

With this setup, the PR description begins by setting the context and outlining the problem, then presents the solution and the steps taken to achieve it.
### Test plan

- [x] Added unit tests for added functionality inside `ModuleFederationPlugin.test.ts`

### DIFF
--- a/packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
+++ b/packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
@@ -150,7 +150,6 @@ describe('ModuleFederationPlugin', () => {
 
   it('should not add deep imports to existing shared dependencies when react-native is not present', () => {
     new ModuleFederationPlugin({
-      reactNativeDeepImports: false,
       shared: {
         react: Federated.SHARED_REACT,
       },
--- a/packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
+++ b/packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
@@ -1,5 +1,6 @@
 import { container } from 'webpack';
 import { ModuleFederationPlugin } from '../ModuleFederationPlugin';
+import { Federated } from '../../federated';
 
 jest.mock('webpack', () => ({
   container: {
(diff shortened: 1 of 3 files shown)


packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
```diff
@@ -97,4 +98,109 @@ describe('ModuleFederationPlugin', () => {
...
+    new ModuleFederationPlugin({
+      shared: {
+        react: Federated.SHARED_REACT,
+        'react-native': Federated.SHARED_REACT_NATIVE,
+      },
+    }).apply(jest.fn() as any);
+    const config = (container.ModuleFederationPlugin as jest.Mock).mock
+      .calls[0][0];
+    expect(config.shared).toHaveProperty('react-native/');
+    expect(config.shared).toHaveProperty('@react-native/');
```
thymikee, line 134: Same assertions are made for `new ModuleFederationPlugin({})` in "should add default shared dependencies" test. How is adding `shared` different?

thymikee (approved):

asherLZR:
Do we know of any instances where we might want to toggle `reactNativeDeepImports: false`? It seems RN internals like `react-native/Libraries/Image/AssetRegistry` and `react-native/Libraries/Utilities/codegenNativeComponent` which handle local images and native components rely on it always being shared/singleton.

jbroma:
> Do we know of any instances where we might want to toggle `reactNativeDeepImports: false`? It seems RN internals like `react-native/Libraries/Image/AssetRegistry` and `react-native/Libraries/Utilities/codegenNativeComponent` which handle local images and native components rely on it always being shared/singleton.

I'm including this as an option to opt out of this solution as from previous experiences, there might be a use-case where that's going to cause issues.

packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
```diff
@@ -97,4 +98,109 @@ describe('ModuleFederationPlugin', () => {
...
+    }).apply(jest.fn() as any);
+    const config = (container.ModuleFederationPlugin as jest.Mock).mock
+      .calls[0][0];
+    expect(config.shared).not.toHaveProperty('react-native/');
+    expect(config.shared).not.toHaveProperty('@react-native/');
+  });
+
+  it('should not add deep imports to existing shared dependencies when react-native is not present', () => {
+    new ModuleFederationPlugin({
+      reactNativeDeepImports: false,
```
RafikiTiki, line 153: Wouldn't setting this flag to `false` prevent adding deep imports regardless of whether `react-native` is present in the sharedDeps?

RafikiTiki (approved):
Great investigation and RN-specific resolution with the undocumented webpack API.

packages/repack/src/webpack/plugins/__tests__/ModuleFederationPlugin.test.ts
jbroma: great find! ❤️  Fixed!

PR #441, opened 2023-10-10
fix(dev-server): fix Failed to open stack frame in editor error

hexboy:
### Summary

This PR fixes the `[DevServer] Failed to open stack frame in editor ` error.
image

### Test plan
image

This error occurs when attempting to open a file in an editor and using the dev tools plugin to get file info from the `request.body` by JSON.parse. However, the issue arises because the `request.body` is an object rather than a string, resulting in an error.

Here is an example of the request:
 `curl 'http://localhost:8081/open-stack-frame' \
-X POST \
-H 'Connection: Keep-Alive' \
-H 'User-Agent: okhttp/4.9.2' \
-H 'Content-Type: application/json' \
--data-raw '{"file":"webpack:///node_modules/react-native/Libraries/LogBox/Data/LogBoxData.js","lineNumber":195}'`

To avoid this error, it's important to verify the type of `request.body` before attempting to parse it.

### DIFF
--- a/.changeset/gold-bananas-think.md
+++ b/.changeset/gold-bananas-think.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack-dev-server": patch
+---
+
+fix(dev-server): fix Failed to open stack frame in editor error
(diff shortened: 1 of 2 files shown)


jbroma (approved):

jbroma (approved):
Hey @hexboy,

Thanks for your contribution! I've verified this fixes the issue, LGTM 🚀

PR #457, opened 2023-10-27
feat: experimental lazy-compilation support

jbroma:
### Summary

Enable experimental lazy compilation support. This is done via `react-native-event-source` package which serves as a polyfill for `EventSource` which is not present in `react-native` environment.

#### Checklist

- [x] - enable lazy compilation in DevelopmentPlugin
- [x] - add DX errors related to lazy compilation
- [x] - add a guide to docs about lazy compilation

### Test plan

- [x] tested locally on TesterApp

### DIFF
--- a/.changeset/stale-panthers-taste.md
+++ b/.changeset/stale-panthers-taste.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+Support lazy-compilation for dynamic imports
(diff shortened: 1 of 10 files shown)


jbroma:
Docs preview: https://653f98faa8057505cc80046a--re-pack.netlify.app/docs/configuration/guides/lazy-compilation

thymikee:
The docs look good. Maybe one more thing that I would expect as a user would be code example – what code patterns do I need to use to benefit from lazy compilation. So something like:

You may find this useful if you have a lot of code like this:
```js
const Component = React.lazy(() => await import('./component'));
return (
  <Supense>
    <Component />
  </Supense>
);
```

PR #463, opened 2023-11-16
chore: upgrade to yarn v4

jbroma:
### Summary

Upgraded `yarn` in the repository to version `4.0.2`

Lockfile format has slightly changed, but the versions are preserved


jbroma:
Looks like this needs to be postponed after we remove support for node 16

thymikee:
Let's create a `v4` branch and start merging stuff there

PR #464, opened 2023-11-16
chore: upgrade TypeScript & Eslint & Typedoc

jbroma:
### Summary

Old versions of `typescript` are blocking fastify upgrade, there it's time to update. To support `typescript@5.x`, `eslint-confg-callstack` needed to be updated as well. `Eslint` and `TypeDoc` were upgraded as they became incompatible as well.

- [x] - upgrade `typescript` in all packages
- [x] - upgrade `eslint` in all packages
- [x] - upgrade `eslint-config-callstack`  in all packages
- [x] - upgrade `typedoc` & `typedoc-markdown-plugin` in dev-server & repack packages
- [x] - fix TypeScript issues
- [x] - fix Eslint issues

### Test plan

- [x] - tests pass
- [x] - lint pass
- [x] - typescript pass


packages/repack/src/webpack/Compiler.ts
```diff
@@ -184,7 +184,7 @@ export class Compiler extends EventEmitter {
       this.progressSenders[platform] = this.progressSenders[platform] ?? [];
       this.progressSenders[platform].push(sendProgress);
     }
-    return new Promise<Asset>((resolve, reject) => {
+    return await new Promise<Asset>((resolve, reject) => {
```
jbroma, line 187: this is just here to satisfy the linter, has no effect on the result

packages/repack/src/modules/WebpackHMRClient.ts
```diff
@@ -128,7 +128,8 @@ class HMRClient {
       });
 
       if (!this.upToDate()) {
-        this.checkUpdates(update);
+        // TODO - should we return here?
+        void this.checkUpdates(update);
```
jbroma, line 132: this should probably be handled in separate PR - as this would cause a change in behaviour, needs to be checked whether returning here would still result in correct behaviour

packages/dev-server/.eslintrc.cjs
```diff
@@ -1,5 +1,14 @@
...
   extends: '@callstack/eslint-config/node',
+  parserOptions: {
+    project: true,
+    tsconfigRootDir: __dirname,
+  },
+  rules: {
+    /* These 2 rules do not play nicely with fastify */
+    'require-await': 0,
+    '@typescript-eslint/no-floating-promises': 0,
+  },
```
jbroma, line 11: as the comment mentions, lot's of Fastify API uses `thenables` which incorrectly get picked up as `Promise`, that's why these 2 rules were disabled

packages/repack/src/commands/start.ts
```diff
@@ -172,16 +172,18 @@ export async function start(_: string[], config: Config, args: StartArguments) {
...
+                ([name, asset]) => ({
+                  name,
+                  size: asset.info.size,
+                })
+              )
             ),
-          getCompilationStats: async (platform) =>
-            compiler.statsCache[platform] ?? null,
+          getCompilationStats: (platform) =>
+            Promise.resolve(compiler.statsCache[platform] ?? null),
```
jbroma, line 186: changed `async` to `Promise.resolve` to satisfy the requirements, otherwise TS & eslint have a conflict

jbroma (commented):
Highlights

thymikee (approved):

PR #465, opened 2023-11-17
chore(dev-server): upgrade fastify & related dependencies in 

jbroma:
### Summary

Upgrade `fastify` and related packages to v4 and v4 compatible.

Migration included changing from deprecated `instance.listen()` variadic function to currently supported one with configuration object.

This PR also fixes an issue where `dev-server` was only listening on only IPv4 or IPv6. Now, when running on `localhost`, both interfaces are being used for listening for connections. 

### Test plan

- [x] - tested locally in TesterApp


thymikee (approved):
Cool! Can you link the issues that this upgrade resolves?

jbroma:
> Cool! Can you link the issues that this upgrade resolves?

not sure there are any open issues about that, I've discovered that some time ago - many issue simply say that the Android can't find the bundle so that will indirectly help limit those kind of questions :)

Issue #304, opened 2023-02-06
Any limitation to use Re.Pack with Expo SDK 47 (and greater)?


jfbaro:

jfbaro:
Please. Any tip would be appreciated. Now that we have Expo Dev Client is this still a limitation?

thymikee:
Hey @jfbaro! I don't have immediate answer on that, but I'll put this on our research plan to see if we can have e.g. a mini-app in Module Federation approach developed with Expo client. It's gonna be way trickier if we'd want to use Expo for a host because we build on Webpack.

iambharath87:
@thymikee an example app with expo would be very much welcome. I created an issue earlier with a [sample repo](https://github.com/iambharath87/expo-repack-starter) i was trying. Maybe its helpful.

outsung:
I've created an example app that includes Expo. You can find the repository [here](https://github.com/outsung/expo-repack-monorepo).

I followed the structure provided in the [callstack/repack-examples (module-federation)](https://github.com/callstack/repack-examples/tree/main/module-federation) repository.

#### Modifications Made:

1. **Added a Plugin to Modify IOS Native Code:**

   I added a [plugin](https://github.com/outsung/expo-repack-monorepo/blob/main/apps/host/plugin/withRepack.js) to modify the iOS native code. (Note: Android implementation is pending.)

2. **Used `metro-react-native-babel-preset`:**

   Instead of `babel-preset-expo`, I switched to using `metro-react-native-babel-preset` in the [babel.config.js](https://github.com/outsung/expo-repack-monorepo/blob/main/apps/host/babel.config.js). Ensure you install the same version of `metro-react-native-babel-preset` as specified in `babel-preset-expo`. Multiple versions can cause errors.

3. **Fix for Expo in `webpack.config.mjs`:**

   I replaced `@expo` with `expo` in the [webpack.config.mjs](https://github.com/outsung/expo-repack-monorepo/blob/082633edded401b7f599ecebd8e033d47266b8fd/apps/host/webpack.config.mjs#L157).

Feel free to review the repository for more details.

msgtobala:
@outsung, As a beginner I am not able to understand how to run your repository [expo-repack-monorepo](https://github.com/outsung/expo-repack-monorepo). Please help me on running the repository

outsung:
@msgtobala My answer is late, but are you saying you don't know which part?

PR #470, opened 2023-12-01
fix: `@expo` changed to `expo` in babel-loader config in webpack config templates

outsung:
### Summary

When using expo library or when using expo and repack,
Need to modify it to `expo` instead of `@expo` to work.

Can refer to this [comment](https://github.com/callstack/repack/issues/304#issuecomment-1809708696).

### DIFF
--- a/packages/TesterApp/webpack.config.mjs
+++ b/packages/TesterApp/webpack.config.mjs
@@ -160,7 +160,7 @@ export default (env) => {
             /node_modules(.*[/\\])+@react-native/,
             /node_modules(.*[/\\])+@react-navigation/,
             /node_modules(.*[/\\])+@react-native-community/,
-            /node_modules(.*[/\\])+@expo/,
+            /node_modules(.*[/\\])+expo/,
             /node_modules(.*[/\\])+pretty-format/,
             /node_modules(.*[/\\])+metro/,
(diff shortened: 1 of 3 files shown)


RafikiTiki:
Hey @outsung! First of all – thanks for your contribution!

I think the `@expo` regex was there for a reason – there are still some expo packages that are named `@expo/package-name` (for example: https://www.npmjs.com/package/@expo/vector-icons), and I think they still need to be processed by webpack. Let me ask around and see if we would like to leave the old `@expo` rule along with the new `@expo` one for backward compatibility.

outsung:
@RafikiTiki Thanks for the confirmation!

jbroma:
@RafikiTiki any update on this?

RafikiTiki:
Thanks for pinging me – from my investigation, it looks like only the [@expo/vector-icons]( https://www.npmjs.com/package/@expo/vector-icons) package uses the `@expo` naming convention. I think we're good to remove it from the default webpack config since we're not using this package internally.

If anyone would want to use it, it can be added to the webpack config same as with other 3rd party packages that need to be transpiled by babel.

RafikiTiki (approved):

jbroma (approved):
LGTM

jbroma:
Thanks for your contribution @outsung!

PR #475, opened 2023-12-12
feat: add `X-React-Native-Project-Root` header

szymonrybczak:
### Summary

Inside React Native CLI we're looking for [`X-React-Native-Project-Root`](https://github.com/react-native-community/cli/blob/772880f8f691e9f049441983214c356abaef041a/packages/cli-tools/src/isPackagerRunning.ts#L37) header, to check if bundler is already running for the project.

| before / after |
|--------|
|  CleanShot 2023-12-12 at 11 45 17@2x |
| CleanShot 2023-12-12 at 11 45 56@2x |

### Test plan

Update latest CLI, or update React Native version inside `packages/TesterApp` and start bundler. When running `run-ios` command, "info A dev server is already running for this project on port 8081." should be printed to console.


packages/dev-server/src/createServer.ts
```diff
@@ -91,6 +91,7 @@ export async function createServer(config: Server.Config) {
 
   instance.addHook('onSend', async (request, reply, payload) => {
     reply.header('X-Content-Type-Options', 'nosniff');
+    reply.header('X-React-Native-Project-Root', process.cwd());
```
thymikee, line 94: is this true for monorepos as well? if not, how we can make it configurable?

szymonrybczak: > is process.cwd() correct for monorepos as well?

probably not the best, updated to use `rootDir` from `config` ✅

thymikee (approved):

thymikee:
cc @kelset for RNTA

packages/dev-server/src/createServer.ts
jbroma: `rootDir` looks good, let's leave it at that

jbroma (approved):
LGTM 🚀 

Thanks for handling this @szymonrybczak!

Issue #473, opened 2023-12-07
react-native@0.73.0 @callstack/repack/commands error


dagouzhi:
## Environment
`react-native.config.js`
```
commands: require('@callstack/repack/commands'),
```

```
[!] Invalid `Podfile` file: 859: unexpected token at 'TypeError: Cannot read properties of undefined (reading 'description')

    at Object.<anonymous> (/xxxxx/node_modules/@callstack/repack/commands.js:62:32)
```

shashwat2003:
Hello! Did you find any fixes yet? IMO it will take time to shift everything to KT.

jbroma:
Fix merged, new version of Re.Pack should be available later today 🚀

PR #477, opened 2023-12-12
fix: get commands options and description from new package

szymonrybczak:
### Summary

Fixes https://github.com/callstack/repack/issues/473. With React Native 0.73, [we've migrated](https://github.com/facebook/react-native/pull/38795) `start` and `bundle` command to Core. So right now when resolving path to CLI and trying to require these commands from CLI, script will fail. In this PR I changed the `require.resolve` to look into new package, where these commands are living right now.

### Test plan

1. Create new React Native app with `npx react-native@latest init app`
2. Init Re.Pack with `npx @callstack/reapck-init`
3. Apply changes from this PR
4. `npx react-native config` should be executed correctly.


packages/repack/commands.js
```diff
@@ -1,43 +1,33 @@
...
-  try {
-    cliPath = path.dirname(require.resolve('@react-native-community/cli'));
-  } catch {
-    // NOOP
-  }
-
   try {
     cliPath = path.dirname(
-      require.resolve('react-native/node_modules/@react-native-community/cli')
+      require.resolve('@react-native/community-cli-plugin')
```
thymikee, line 8: This will break Re.Pack working with 0.71 or 0.72, isn't it? Couldn't we look for `"bin"` entry in `react-native` package, which points to the cli?

jbroma:
@szymonrybczak is this still compatible with previous versions? We need to ensure compatibility with at least last 3 versions of RN (0.71, 0.72, 0.73)

packages/repack/commands.js
szymonrybczak: > Couldn't we look for "bin" entry in react-native package, which points to the cli?

I don't think so ;/ Of course we can get path to the `cli.js` and it will be resolved, but how can we get informations about commands from there? 🤔

szymonrybczak: Unfortunately we can't go with the easy path, because right now `start` and `bundle` command are added in `react-native.config.js` so they won't exist under `/commands` dir.

RafikiTiki:
With the backwards compatibility it looks great, thanks!

hosseinmd:
I tested this change by react-native@73, but there is a problem when I build android, main js doesn't compile with repack,
webpack.config.js never run

packages/repack/commands.js
```diff
@@ -25,19 +25,38 @@ function getReactNativeCliPath() {
...
+  try {
+    commands = require(require.resolve(
+      'react-native/react-native.config.js'
+    )).commands;
+  } catch (e) {
+    // NOOP
+  }
+
+  if (!commands) {
+    throw new Error('Cannot resolve path react-native package');
```
jbroma, line 46:
```suggestion
    throw new Error('Cannot resolve path to react-native package');
```

jbroma (changes requested):
Just few details to fix, other than that GJ! 🚀

szymonrybczak:
Suggestsions applied, should be good now! 🙌

jbroma (approved):
LGTM 🚀

PR #479, opened 2023-12-16
fix: make webpack cache work

ambar:
### Summary

See https://webpack.js.org/api/node/#run

> Don't forget to close the compiler, so that low-priority work (like persistent caching) have the opportunity to complete.

```sh
turbo run build && yarn workspace testerapp run bundle:ios

# without cache
webpack 5.75.0 compiled successfully in 14680 ms 
# with cache
webpack 5.75.0 compiled successfully in 1000 ms
```

### DIFF
--- a/.changeset/serious-hairs-thank.md
+++ b/.changeset/serious-hairs-thank.md
@@ -1,0 +1,6 @@
+---
+"testerapp": patch
+"@callstack/repack": patch
+---
+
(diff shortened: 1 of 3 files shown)


RafikiTiki (approved):
Looks good, thanks for the contribution – the `compiler.close` was obviously missed there 🙌

jbroma:
Thank you for your contribution @ambar, great stuff 🚀

PR #481, opened 2023-12-19
chore: upgrade react-native dep and remove deprecated @types/react-native

RafikiTiki:
### Summary

- Upgraded `react-native` dependency to 0.73.1 in the TesterApp
- removed deprecated `@types/react-native` package (https://reactnative.dev/blog/2023/12/06/0.73-debugging-improvements-stable-symlinks#deprecated-typesreact-native).

### Test plan

- Build packages in monorepo
- Run the TesterApp


packages/TesterApp/package.json
```diff
@@ -21,7 +21,7 @@
     "@react-native-async-storage/async-storage": "^1.21.0",
     "lodash.throttle": "^4.1.1",
     "react": "18.2.0",
-    "react-native": "0.73.0",
+    "react-native": "^0.73.1",
```
jbroma, line 24: I think we should stick to single SemVer version instead of range here

packages/repack/package.json
```diff
@@ -99,14 +99,13 @@
...
     "@types/react-dom": "^17.0.7",
-    "@types/react-native": "^0.67.8",
     "@types/shallowequal": "^1.1.1",
     "babel-jest": "^28.1.1",
     "babel-loader": "^8.2.2",
     "eslint": "^7.30.0",
     "headers-polyfill": "^3.0.7",
     "jest": "^28.1.1",
-    "react-native": "^0.64.1",
+    "react-native": "^0.73.1",
```
jbroma, line 108: Let's do the same thing here, and keep them in sync

jbroma (approved):
All good, I've pointed out few things worth discussing, WDYT about them @RafikiTiki?

PR #491, opened 2024-01-30
fix: log errors before exit

hosseinmd:
### Summary
To clarify what are errors exactly, we need to log them before exist.

### DIFF
--- a/.changeset/bright-pillows-argue.md
+++ b/.changeset/bright-pillows-argue.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
+log error before exit when compilation has error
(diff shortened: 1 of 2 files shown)


jbroma:
Hi @hosseinmd, thanks for your contribution!

Can you showcase an example of how this improves error reporting when bundling? Before and after would be ideal, thanks!

hosseinmd:
## Before adding this I didn't have any idea what is error this is my logs

Screenshot 2024-01-30 at 8 10 57 PM


*****

## after adding this

******
Screenshot 2024-01-30 at 8 14 24 PM


Now it is clear what is the exact error.


Font color is different, that is about something else forget it.

jbroma (approved):

jbroma:
@hosseinmd thank you for your contribution! 🚀

PR #495, opened 2024-01-31
New arch migration

troZee:
### Summary
This PR aims to add a turbo module with backward compatibility. Re.Pack current implementation directly uses RN native modules, but with this change, there is a `SPEC` implementation that allows for more efficient communication between native and JavaScript. This change allows for both old and new architecture. All implementation follows the official react native bob builder template. 

### Test plan

- Make sure you are currently using the newArch and the setup conforms to [React Native Doc directives](https://reactnative.dev/docs/next/new-architecture-app-intro) which also means you're on **`RN>=0.68.0`**

NOTE: TesterApp currently meets this requirement if you're using repack TesterApp.

 #### For Android
- You will only need to update your `android/gradle.properties` file as follows:
**`newArchEnabled=true`**

-  and then **`yarn TesterApp:start`**

#### For iOS
- Run pod install with the flags: 
**`RCT_NEW_ARCH_ENABLED=1 pod install`**

-  and then  **`yarn TesterApp:start`**

### DIFF
--- a/packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModule.kt
+++ b/packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModule.kt
@@ -57,22 +57,23 @@ class ScriptManagerModule(reactContext: ReactApplicationContext) : ScriptManager
         if (!config.fetch) {
             // Do nothing, script is already prefetched
             promise.resolve(null);
-        } else {
-            runInBackground {
-                when {
-                    config.url.protocol.startsWith("http") -> {
-                        remoteLoader.prefetch(config, promise)
-                    }
+            return
+        }
+        runInBackground {
+            when {
+                config.url.protocol.startsWith("http") -> {
+                    remoteLoader.prefetch(config, promise)
+                }
 
-                    else -> {
-                        promise.reject(
-                            ScriptLoadingError.UnsupportedScheme.code,
-                            "Scheme in URL: '${config.url}' is not supported"
-                        )
-                    }
+                else -> {
+                    promise.reject(
+                        ScriptLoadingError.UnsupportedScheme.code,
+                        "Scheme in URL: '${config.url}' is not supported"
+                    )
                 }
             }
         }
+
     }
 
     @ReactMethod
--- a/packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModule.kt
+++ b/packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModule.kt
@@ -3,9 +3,10 @@ package com.callstack.repack
 import android.os.Handler
 import com.facebook.react.bridge.*
 
-class ScriptManagerModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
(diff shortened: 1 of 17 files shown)


packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModule.kt
```diff
@@ -35,62 +36,69 @@ class ScriptManagerModule(reactContext: ReactApplicationContext) : ReactContextB
...
-                }
-                else -> {
-                    promise.reject(
+    @ReactMethod
+    override fun prefetchScript(scriptId: String, configMap: ReadableMap, promise: Promise) {
+        val config = ScriptConfig.fromReadableMap(scriptId, configMap)
+        if (!config.fetch) {
+            // Do nothing, script is already prefetched
+            promise.resolve(null);
+        } else {
```
okwasniewski, line 60: How about changing this to early return? It would make it more readable

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt
```diff
@@ -1,17 +1,35 @@
...
-class ScriptManagerPackage : ReactPackage {
-    override fun createNativeModules(reactContext: ReactApplicationContext): List<NativeModule> {
-        return listOf(ScriptManagerModule(reactContext))
+class ScriptManagerPackage : TurboReactPackage() {
+    override fun getModule(name: String, reactContext: ReactApplicationContext): NativeModule? {
+        return if (name == ScriptManagerModule.NAME) {
+            ScriptManagerModule(reactContext)
+        } else {
+            null
+        }
```
okwasniewski, line 16: nit: Not needed `else {}`

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt
```diff
@@ -1,17 +1,35 @@
...
-        return emptyList()
+    override fun getReactModuleInfoProvider(): ReactModuleInfoProvider {
+        return ReactModuleInfoProvider {
+            val moduleInfos: MutableMap<String, ReactModuleInfo> = HashMap()
+            val isTurboModule: Boolean = BuildConfig.IS_NEW_ARCHITECTURE_ENABLED
+            moduleInfos[ScriptManagerModule.NAME] = ReactModuleInfo(
+                ScriptManagerModule.NAME,
+                ScriptManagerModule.NAME,
+                false,  // canOverrideExistingModule
+                false,  // needsEagerInit
```
okwasniewski, line 27: This is interesting, shouldn't repack have eager init enabled?

okwasniewski (commented):
Looks good! Left few questions

atlj (approved):

krozniata (approved):

packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt
troZee: Screenshot 2024-02-01 at 10 02 04

It seems to be needed :D

okwasniewski (approved):

jbroma (approved):
Great stuff, thanks @troZee 🎉

PR #508, opened 2024-02-23
feat: add RN experimental debugger compatibility

RafikiTiki:
### TODO

- [x] Verify if the Flipper debugging experience is working properly – if not – create an issue and address it later
- [x] Consider reverting DevtoolsPlugin removal – it might be needed for the deprecated Remote JS debugging. We need this to work before it's removed from the RN core

### Summary

This PR introduces compatibility with the new Experimental Debugger to the Repack devserver.

Experimental Debugger info:
- https://reactnative.dev/blog/2023/12/06/0.73-debugging-improvements-stable-symlinks#experimental-new-debugger
- https://reactnative.dev/docs/debugging#opening-the-debugger

To test this feature the dev server has to be started with the `--experimental-debugger` flag. To play with it using `TesterApp` simply run `yarn TesterApp:start --experimental-debugger` from the root of the repo.

image

### Additional benefits

This PR also made it possible to remove a custom implementation of the Hermes Inspector Proxy and reuse what `@react-native/dev-middleware` provides – less code to maintain and less breaking changes in the future 🤞 

### Next steps

In the future, we'll likely remove the Remote JS debugger and the whole `debugger-app` package – this will, of course, be a breaking change and we want to wait till [it gets removed from the RN core itself](https://reactnative.dev/blog/2023/12/06/0.73-debugging-improvements-stable-symlinks#remote-javascript-debugging).


jbroma:
- Added `eslint-disable` directive fro `@react-native/dev-middleware` because there is no main field declared there. ([description of the problem](https://github.com/import-js/eslint-plugin-import/issues/2132))
- Added TS path override for `@react-native/dev-middleware` because we use `moduleResolution: 'node'` which doesn't read package exports, and changing this is out of scope of this PR

jbroma:
@szymonrybczak should we use `@react-native/dev-middleware` as a peer dependency in our dev-server? I think it makes sense because it's always included with `react-native` as part of `@react-native/community-cli-plugin`, would love to hear your point of view

thymikee:
FYI, in 0.75 the template will have the dependency on `@react-native-communit/cli`, but likely not directly on the `@react-native/community-cli-plugin`

jbroma:
> Using it as a peer dependency would make it auto-install as a dep when using npm, right? 

that might be the case, let's mark it as optional inside of `peerDependenciesMeta` then I guess?

> Can we treat it as an optional dependency that we require and throw error saying that it's supposed to be there, so most likely make sure you've installed the deps?

We could make it optional and just skip integrating that part with debugger altogether. On the other hand, I'm not sure we need the error handling here since `dev-server` doesn't make much sense outside of `react-native` context so it should be safe to assume `react-native` is there. I would love to hear others opinion on this though. 

> FYI, in 0.75 the template will have the dependency on @react-native-communit/cli, but likely not directly on the @react-native/community-cli-plugin: https://github.com/react-native-community/template/pull/7

I don't see anything there (or related PR) regarding `@react-native/community-cli-plugin`, but if that happens then I guess we could just align to that in the future.

packages/dev-server/package.json
```diff
@@ -59,9 +60,13 @@
     "source-map": "^0.7.4",
     "ws": "^8.7.0"
   },
+  "peerDependencies": {
```
thymikee, line 63: Following up on our discussion
```suggestion
  "peerDependenciesMeta": {
```

thymikee: btw, will this satisfy pnpm too when sync importing `@react-native/dev-middleware`?

packages/repack/src/commands/start.ts
```diff
@@ -67,7 +67,7 @@ export async function start(_: string[], config: Config, args: StartArguments) {
   const { start, stop } = await createServer({
     options: {
       rootDir: cliOptions.config.root,
-      host: args.host,
+      host: args.host?.length ? args.host : 'localhost',
```
jbroma, line 70: let's pick this to a separate PR

jbroma (changes requested):

packages/repack/src/commands/start.ts
RafikiTiki: Rebased on `main`

RafikiTiki:
Flipper compatibility update – I rebased this branch on the latest changes from the `main` branch, which includes RN 0.74.0 in the `TesterApp`. Both Flipper and Experimental Debugger experiences are working as expected now 🎉 

image

RafikiTiki:
Deprecated Remote Debugging #578 update – I reverted removal of part of the `devtoolsPlugin.ts` and used a few middlewares from `@react-native-community/cli-server-api` & replaced vendored code for the debugger-ui with `@react-native-community/cli-debugger-ui`. This enabled me to completely remove the `packages/debugger-app`.

Additionally, I left a TODO comment to remove the internal code/dependencies related to the deprecated remote debugging once it is removed from the React Native core itself.

@thymikee @jbroma, this should be good to be re-reviewed. The last thing to consider is the status of the added dependencies:

```
"@react-native-community/cli-debugger-ui": "^13.6.6",
"@react-native-community/cli-server-api": "^13.6.6",
"@react-native/dev-middleware": "^0.73.8",
```

As mentioned above, the first two will be removed in the future, while `@react-native/dev-middleware` will continue to be used. I've seen the discussion earlier about moving them to the peerDeps. Since debugging is essential to the development experience, it should be provided by repack. Previously we had all of the code implemented internally – I think leaving them as regular dependencies will be a better dev experience rather than marking them optional and providing runtime checks/errors in case they're not installed.

jbroma (commented):
few nits

packages/dev-server/src/createServer.ts
```diff
@@ -59,10 +65,24 @@ export async function createServer(config: Server.Config) {
     },
   });
 
+  const devMiddleware = createDevMiddleware({
+    projectRoot: config.options.rootDir,
+    serverBaseUrl: `http://${config.options.host}:${config.options.port}`,
+    // TODO verify this works properly
+    logger: instance.log,
```
jbroma, line 72: @RafikiTiki did you manage to verify whether this is ok?

jbroma (approved):
1 nit, other than that LGTM 🎉

packages/dev-server/package.json
jbroma: we decided to include it as direct dep for now, we'll revisit this subject when the oldest supported RN version includes dev-middleware

packages/dev-server/package.json
```diff
@@ -70,5 +70,9 @@
     "typedoc": "^0.25.3",
     "typedoc-plugin-markdown": "^3.17.1",
     "typescript": "^5.2.2"
+  },
+  "peerDependencies": {
+    "@react-native-community/cli-debugger-ui": ">=11.3.1",
+    "@react-native-community/cli-server-api": ">=11.3.1"
```
thymikee, line 76: I'd change this to `peerDependenciesMeta`, otherwise npm will install these deps (which are transitive deps of `react-native` currently and don't normally show up in app's dependencies)

packages/dev-server/src/createServer.ts
RafikiTiki: Yes, it does work flawlessly – TODO removed

jbroma (changes requested):
Missing changeset

jbroma (approved):
LGTM 🎉

PR #515, opened 2024-03-05
refactor(repack): use `done` hook inside of `OutputPlugin`

jbroma:
### Summary

Migrated `OutputPlugin` to run only inside of `done` hook - this enables usage of `OutputPlugin` with `rspack` as previous approach utilised features not available in `rspack` yet(?)

### Test plan

- [x] - updated tests pass
- [x] - output is the same as before

### DIFF
--- a/packages/repack/src/webpack/plugins/OutputPlugin.ts
+++ b/packages/repack/src/webpack/plugins/OutputPlugin.ts
@@ -183,6 +183,11 @@ export class OutputPlugin implements WebpackPlugin {
       return;
     }
 
+    const outputPath = compiler.options.output?.path;
+    if (!outputPath) {
+      throw new Error('Cannot infer output path from compilation');
+    }
+
     const logger = compiler.getInfrastructureLogger('RepackOutputPlugin');
 
     const extraAssets = (this.config.extraChunks ?? []).map((spec) =>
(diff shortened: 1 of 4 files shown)


packages/repack/src/webpack/plugins/OutputPlugin.ts
```diff
@@ -213,205 +218,202 @@ export class OutputPlugin implements WebpackPlugin {
...
-          }
+    compiler.hooks.done.tapPromise('RepackOutputPlugin', async (stats) => {
+      const compilationStats = stats.toJson({
+        all: false,
+        assets: true,
+        chunks: true,
+        chunkRelations: true,
+        ids: true,
+      });
+      const chunks = new Map(
```
RafikiTiki, line 246: NIT: consider renaming this to sth like `allCompilationChunks` for readability – this map is used multiple times and I found myself wondering whether it was the "all chunks map" or some more scoped temp variable when building specific (local/remote/shared) chunks lists

RafikiTiki (approved):
LGTM!

jbroma:
one thing of note: there are few cases in this PR where I was forced to use [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), but newest [webpack 5.90.3](https://github.com/webpack/webpack/releases/tag/v5.90.3) introduced some improved typings for `stats.toJson()` so we might be able to fix it in a sensible way in a separate PR.

PR #524, opened 2024-03-18
test:  metro-compat resolution tests

jbroma:
### Summary

This PR introduces `metro-compat` package which houses tests suites designed for checking compatibility with parts of `metro`. 

Right now, tests include resolver tests directly imported from `metro-resolver` from [here](https://github.com/facebook/metro/tree/main/packages/metro-resolver/src/__tests__)

Test environment is modified to accommodate for how `enhanced-resolve` works. Basic principle is this: use unchanged tests and provide compatible `resolve` function for tests. While in `metro-resolver` everything happens via mocked methods for checking whether file exists, our tests create a fresh in-memory filesystem for each test to test resolution against, which yields more ensuring results.

Some tests suite are skipped:
- `assets-test.js` - because `enhanced-resolve`, like node resolution, returns only a single entry, it's impossible to support these test cases. We handle this differently using `AssetResolver` which provides you with first matching entry for development, but all scales are exposed to the runtime
- `index-test.js` - this suite contains some basic tests & failure tests which we could include but most of it contains tests for unsupported features like: 
   - `resolveRequest` - this is essentially a custom resolver or a plugin, it should be documented how to achieve similar functionality in `Re.Pack`
   - `redirectModulePath` - this is equal to `alias` field in `enhanced-resolve`, if end user wants to do this dynamically, then custom resolver/plugin should be used for that - also needs to documented
   - `disableHierarchicalLookup` - can be achieved in `enhanced-resolve` by using `modules` - also needs to be documented

Some test cases are skipped (all come from `package-exports-test.js`):
- `[nonstrict]` tests are skipped because we don't support fallback in form of legacy file resolution
- asset tests with package exports enabled are skipped for the same reason as above (regarding the assets)
- rest of the test skipped are due to inline snapshot not matching up with what `enhanced-resolve` returns - this could be potentially fixed by trapping `toThrowErrorMatchingInlineSnapshot` calls and checking for existence of errors only

Tests will be excluded from CI for now until everything is aligned properly and some issues are brought up with metro team. Alignment of `getResolveOptions` will be done in a separate PR. Documentation will follow in a separate PR as well.

Side-note: this PR also unveiled few issues with package exports implementation in `metro-resolver` which will be investigated more throughly and perhaps become more aligned in the future.

Side-note 2: the way we skip tests without modyfing the test files is rather hacky, we trap `jest.describe`, `jest.it` & `jest.test` and we check if the test name matches, TBD whether there is a better way to do this.

### Checklist
- [x] `metro-resolver` test suite
- [x] setup mocks
- [x] setup skip list 
- [x] exclude tests from turborepo so they are not run with the rest of the tests
- [x] cleanup
- [x] README.md in `metro-compat`


jbroma:
Issue raised with metro: https://github.com/facebook/metro/issues/1236

jbroma:
@RafikiTiki don't mind the enabled tests failing for now, it's expected without reworking implementation

RafikiTiki (approved):

PR #525, opened 2024-03-18
[BREAKING] feat: make resolver preset follow Metro defaults closely

jbroma:
### Summary

BREAKING CHANGE:

`getResolveOptions` is now way more compatible with `metro-resolver` and `@react-native/metro-config`

1. `getResolveOptions` now accepts a second optional parameter called options with the following properties:
   - `enablePackageExports` - defaults to `false`
   - `preferNativePlatform` - defaults to `true`
2. Order of extensions was changed to match the order from `@react-native/metro-config`.
3. Resolution via Package Exports (`exports` field in package.json) is now optional and disabled by default.
   It can now be enabled via `getResolveOptions` options parameter. This change was introduced to match `metro` defaults.
4. Default `conditionNames` are now: `['require', 'import', 'react-native']` and match `@react-native/metro-config` defaults.

### Remarks

Implementation passes every `metro-compat` test with the following exceptions in `package-exports-test`:
- **_should resolve "exports" target directly > without expanding `sourceExts`_** - test should be marked as `[nonstrict]` as it relies on unsupported fallback mechanism
- **_should resolve "exports" target directly > without expanding platform-specific extensions_** - test should be marked as `[nonstrict]` as it relies on unsupported fallback mechanism
- **_subpath exports > should expand array of strings as subpath mapping (root shorthand)_** - this spec is not compliant with Node spec and should be fixed upstream as it's original purpose was to match Node spec.
- _**subpath patterns > should resolve subpath patterns in "exports" matching import specifier**_ - one of the examples in the test is bad - `Bar.js` should not be exposed by the package according to the Node spec, but instead fallback to using less specific specifier. This mechanism is used to restrict imports in packages and should be fixed upstream

### Checklist
- [x] - rework implementation of `getResolveOptions`
- [x] - adjust `JSDoc`
- [x] - ensure `TypeDoc` compiles and displays new descriptions properly
- [x] - add tests from Remarks to skipped ones 

### Test plan

- [x] - pass selected `metro-compat` tests

### DIFF
--- a/packages/repack/src/webpack/utils/getResolveOptions.ts
+++ b/packages/repack/src/webpack/utils/getResolveOptions.ts
@@ -49,21 +49,7 @@ export function getResolveOptions(platform: string, options?: ResolveOptions) {
   const preferNativePlatform = options?.preferNativePlatform ?? true;
   const enablePackageExports = options?.enablePackageExports ?? false;
 
-  let extensions = ['.ts', '.js', '.tsx', '.jsx', '.json'];
-
-  const platformExtensions = [
-    `.${platform}.ts`,
-    `.${platform}.js`,
-    `.${platform}.tsx`,
-    `.${platform}.jsx`,
-  ];
-
-  const nativeExtensions = [
-    '.native.ts',
-    '.native.js',
-    '.native.tsx',
-    '.native.jsx',
-  ];
+  let extensions = ['.js', '.jsx', '.json', '.ts', '.tsx'];
 
   let conditionNames: string[];
   let exportsFields: string[];
@@ -79,11 +65,16 @@ export function getResolveOptions(platform: string, options?: ResolveOptions) {
   } else {
     conditionNames = [];
     exportsFields = [];
-    extensions = [
-      platformExtensions,
-      preferNativePlatform ? nativeExtensions : [],
-      extensions,
-    ].flat();
+    extensions = extensions.flatMap((ext) => {
+      const platformExt = `.${platform}${ext}`;
+      const nativeExt = `.native${ext}`;
+
+      if (preferNativePlatform) {
+        return [platformExt, nativeExt, ext];
+      } else {
+        return [platformExt, ext];
+      }
+    });
   }
 
   /**
--- a/packages/repack/src/webpack/utils/getResolveOptions.ts
+++ b/packages/repack/src/webpack/utils/getResolveOptions.ts
@@ -1,5 +1,5 @@
 /**
- * {@link getResolveOptions} options.
+ * {@link getResolveOptions} additional options.
  */
 export interface ResolveOptions {
   /**
@@ -16,7 +16,8 @@ export interface ResolveOptions {
  * Get Webpack's resolve options to properly resolve JavaScript files:
  * - resolve platform extensions (e.g. `file.ios.js`)
  * - resolve native extensions (e.g. `file.native.js`)
- * - optionally use package exports (`exports` field in `package.json`) instead of main fields (e.g. `main` or `browser` or `react-native`)
+ * - optionally use package exports (`exports` field in `package.json`) instead of
+ *   main fields (e.g. `main` or `browser` or `react-native`)
  *
  * @param platform Target application platform.
  * @param options Additional options that can modify resolution behaviour.
@@ -25,6 +26,7 @@ export interface ResolveOptions {
  * @category Webpack util
  *
  * @example Usage in Webpack config:
+ *
  * ```ts
  * import * as Repack from '@callstack/repack';
  *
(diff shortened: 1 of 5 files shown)


packages/repack/src/webpack/utils/getResolveOptions.ts
```diff
@@ -17,35 +35,83 @@
...
  * };
  * ```
  */
-export function getResolveOptions(platform: string) {
+
+export function getResolveOptions(platform: string, options?: ResolveOptions) {
+  const preferNativePlatform = options?.preferNativePlatform ?? true;
+  const enablePackageExports = options?.enablePackageExports ?? false;
+
+  let extensions = ['.ts', '.js', '.tsx', '.jsx', '.json'];
```
thymikee, line 52: Shouldn't we match default Metro extensions order too?
https://github.com/facebook/metro/blob/7250388527dab827ba86f57b26ca5d8dd0098889/packages/metro-config/src/defaults/defaults.js#L56

jbroma: nice find 🚀 , it's best that we match that as well, it's a rare edge case but might cause issues nonetheless

thymikee:
What are possible migration paths, or differences in resolve behaviors that users could bump into unintentionally? This also shows some promise to remove now unnecessary webpack config overrides, can we provide some examples of what this means for the users in regards what can possibly change in their config?

packages/repack/src/webpack/utils/getResolveOptions.ts
```diff
@@ -17,35 +35,83 @@
...
+     */
+    conditionNames: conditionNames,
+    /**
+     * Reference: Webpack's [configuration.resolve.exportsFields](https://webpack.js.org/configuration/resolve/#resolveexportsfields)
+     */
+    exportsFields: exportsFields,
+    /**
+     * Reference: Webpack's [configuration.resolve.extensions](https://webpack.js.org/configuration/resolve/#resolveextensions)
+     */
+    extensions: extensions,
```
thymikee, line 115: nit:
```suggestion
    conditionNames,
    /**
     * Reference: Webpack's [configuration.resolve.exportsFields](https://webpack.js.org/configuration/resolve/#resolveexportsfields)
     */
    exportsFields,
    /**
     * Reference: Webpack's [configuration.resolve.extensions](https://webpack.js.org/configuration/resolve/#resolveextensions)
     */
    extensions,
```

jbroma: This is there only for proper displaying of the typedoc/rendering API doc page - without this it renders incorrectly afaik.

thymikee (approved):

packages/repack/src/webpack/utils/getResolveOptions.ts
szymonrybczak: also we could re-use https://github.com/facebook/metro/blob/7250388527dab827ba86f57b26ca5d8dd0098889/packages/metro-config/src/defaults/defaults.js#L18-L52 here:
https://github.com/callstack/repack/blob/b0b8613eab82641e91607e0931eafc2360ff42a2/packages/repack/src/webpack/utils/assetExtensions.ts#L2-L43

jbroma:
> What are possible migration paths, or differences in resolve behaviours that users could bump into unintentionally?

The only real difference is that Package Exports support is now disabled by default. There might be packages out there that don't declare `main` field and only use `exports` field, but it's super rare (e.g. `@react-native/dev-middleware`). In this case this could be a breaking change. Until now users needed to override resolution via `alias` to get it to work (or disabling usage of package exports`), or they relied on `exports` entrypoint. 

> This also shows some promise to remove now unnecessary webpack config overrides, can we provide some examples of what this means for the users in regards what can possibly change in their config?

We still need to override the webpack config defaults, nothing changes in this regard. The API of `getResolveOptions` is extended in a non-breaking way, with proper defaults in place 👍

packages/repack/src/webpack/utils/getResolveOptions.ts
jbroma: @szymonrybczak could you elaborate more? do you mean plugging into `@react-native/metro-config` ?

szymonrybczak: ah, I misunderstood first comment. yeah, that's what I meant but that's not possible 😅

RafikiTiki (approved):

packages/repack/src/webpack/utils/getResolveOptions.ts
jbroma: the order of extensions is now matched exactly as in metro 👍

jbroma:
Order of extensions now matches `@react-native/metro-config` 1:1, updated PR description & changeset to reflect that, thanks @thymikee for spotting this 👍

PR #530, opened 2024-03-22
[BREAKING] Remove ReactRefreshPlugin

jbroma:
### Summary

`ReactRefreshPlugin` was unmaintained for a very long time, and all of it's functionality was included inside of `DevelopmentPlugin` which is also part of `RepackPlugin`. I couldn't find a single instance of this plugin being used on GitHub, so it's most likely safe to remove.

### Test plan

n/a


packages/repack/src/webpack/plugins/ReactRefreshPlugin.ts
```diff
@@ -1,95 +0,0 @@
-import webpack from 'webpack';
```
thymikee, line 1: maybe worth to make it a no-op with a warning "This plugin has been deprecated and is now an no-op. Please switch to `DevelopmentPlugin`..."

jbroma: as mentioned above, I'm not sure anyone used it ever, so we should even skip compatibility layer like that

thymikee (approved):

Issue #487, opened 2024-01-09
[RFC] Override `start` command in `react-native.config.js`


szymonrybczak:
## Describe the feature

In React Native Community CLI recently we allowed to override existing commands from `react-native.config.js`: https://github.com/react-native-community/cli/pull/2229.  

## Motivation

When running `run-ios` and `run-android` we're executing `node /path/to/cli start`, so automatically it starts Metro bundler instead of Re.Pack, which may be confusing for non-advanced users.

This also will unblock https://github.com/nrwl/nx/issues/20847

szymonrybczak:
cc @jbroma @tom-sherman @thymikee

jbroma:
@szymonrybczak 

I gave it some thought and I think this will be a good way going forward, this will reduce the overhead when migrating from `metro` and avoid some silly mistakes on users part. It's also worth noting that we the commands used in `Re.Pack` are aligned to that of `metro` so the change so it should be relatively smooth to add. If somebody uses `metro` fro development and `Re.Pack` for production, it can still manually adjusted in `react-native.config.js`.

This is quite a big change and definitely breaking so I'm adding an RFC to the title, and hope to gain more opinion on this matter.

jbroma:
@RafikiTiki @thymikee @tom-sherman some insights from you would be great

thymikee:
As long as we document how to fall back to previous behavior, or expose metro's start command, I'm good with moving forward with this 👍🏼

tom-sherman:
Massively in favour of this default.

I think 90% of developers onboarded hit an issue where they somehow manage to trigger metro.

jbroma:
We should proceeded with this feature when RN 0.74 with CLI 13 lands. I did some brief testing and it seems to work nicely by just changing the name of the commands. 

Actionable items:
- [ ] - Change command names in `packages/repack/commands.js`
- [ ] - Add `legacyCommands` for compatibility with older versions of RN 
- [ ] - Align `@callstack/repack-init` to be able to choose whether to use overriden or separate commands
- [ ] - Update `Getting Started` section in docs to reflect these changes and mention how to use `legacyCommands`

thymikee:
Closed by #537

PR #534, opened 2024-04-03
chore: upgrade tester app to react-native 0.74

jbroma:
### Summary

upgrade to `0.74-rc.6` to enable working on https://github.com/callstack/repack/issues/487

### Test plan

- [x] - ios works
- [x] - android works


jbroma:
needed to disable bridgless as it's enabled by default in RN 0.74 and we don't suport it yet.

PR #537, opened 2024-04-04
[BREAKING] feat: override start & bundle commands

### LINKED ISSUE #93: Launch webpack instead of metro during development
## Describe the feature

After creating a fresh RN project (RN 0.65.1, TypeScript) & installing repack, running a build in Xcode will still start the Metro packager (instead of Webpack). I solved this by patching `packager.sh` in `react-native/scripts` to override the packager command (instead of `react-native start`, setting `PACKAGER_CMD=webpack-start` will launch webpack instead), and modifying the "Start Packager" build phase in Xcode as follows:

`react-native` patch: https://github.com/rahulsmehta/repack-example/blob/master/patches/react-native%2B0.65.1.patch

```bash
export RCT_METRO_PORT="${RCT_METRO_PORT:=8081}"
+ export PACKAGER_CMD="webpack-start"

echo "export RCT_METRO_PORT=${RCT_METRO_PORT}" > "${SRCROOT}/../node_modules/react-native/scripts/.packager.env"
if [ -z "${RCT_NO_LAUNCH_PACKAGER+xxx}" ] ; then
  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then
    if ! curl -s "http://localhost:${RCT_METRO_PORT}/status" | grep -q "packager-status:running" ; then
      echo "Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly"
      exit 2
    fi
  else
    open "$SRCROOT/../node_modules/react-native/scripts/launchPackager.command" || echo "Can't start packager automatically"
  fi
fi

```

(note: this could also be due to error on my part while setting up my example project, so would love any feedback/feel free to close if that's the case)

## Motivation

In the interest of making setup easier, it'd be possible to pull the modified versions of `launchPackager.command`/`packager.sh` into a `scripts/` dir in repack & adding the overrides to the "Start Packager" build phase in the docs. Would there be any interest in accepting a PR for this enhancement?

### LINKED ISSUE #487: [RFC] Override `start` command in `react-native.config.js`

jbroma:
### Summary

Closes #93
Closes #487

Override `start` & `bundle` commands from CLI.

For maximum compatibility, instead of replacing `webpack-start` and `webpack-bundle` commands, the overrides were added on top of old commands. This results in following behaviour:

- for React-Native <0.74 (CLI <13)  - CLI won't apply overrides, webpack commands will work just as before
- for React-Native >=0.74 (CLI >=13) - both overrides and webpack commands work

It's possible to opt out of overrides by simply filtering out commands in `react-native.config.js`:
```js
const commands = require('@callstack/repack/commands');

module.exports = {
  commands: commands.filter((command) => command.name.startsWith('webpack')),
};
```

This also solves the issue with launching metro dev server when running `react-native run-ios` or `react-native run-android` - Re.Pack dev server will be launched instead

### Test plan

- [x] - tested locally
- [ ] - add integration tests to TesterApp suite

### DIFF
--- a/packages/repack/commands.d.ts
+++ b/packages/repack/commands.d.ts
@@ -1,5 +1,17 @@
 const commands: [
   {
+    name: 'bundle';
+    description: string;
+    options: Array<any>;
+    func: typeof import('./dist/commands/bundle').bundle;
+  },
+  {
+    name: 'start';
+    description: string;
+    options: Array<any>;
+    func: typeof import('./dist/commands/start').start;
+  },
+  {
     name: 'webpack-bundle';
     description: string;
     options: Array<any>;
@@ -10,6 +22,6 @@ const commands: [
     description: string;
     options: Array<any>;
     func: typeof import('./dist/commands/start').start;
-  }
+  },
 ];
 export default commands;
--- a/packages/repack/commands.js
+++ b/packages/repack/commands.js
@@ -72,9 +72,9 @@ const webpackConfigOption = {
   },
 };
 
-module.exports = [
+const commands = [
   {
-    name: 'webpack-bundle',
+    name: 'bundle',
     description: bundleCommand.description,
     options: bundleCommand.options.concat(
       {
@@ -96,7 +96,7 @@ module.exports = [
     func: require('./dist/commands/bundle').bundle,
   },
   {
-    name: 'webpack-start',
+    name: 'start',
     options: startCommand.options.concat(
       {
         name: '--verbose',
@@ -124,3 +124,10 @@ module.exports = [
     func: require('./dist/commands/start').start,
   },
 ];
+
+const webpackCommands = commands.map((command) => ({
+  ...command,
+  name: `webpack-${command.name}`,
+}));
+
+module.exports = [...commands, ...webpackCommands];


jbroma:
cc @thymikee @szymonrybczak

szymonrybczak (approved):
LGTM, I think it'd be valuable to add note to docs 👍

szymonrybczak:
For future it'd good to also somehow inform user about options that are not available, because right now we're explicitly using the same options as Metro's `start` and `bundle` command, and some of them aren't implemented (yet).
https://github.com/callstack/repack/blob/270ca1a29bff3dada233b05b96c805453fe5d75d/packages/repack/commands.js#L55-L56

jbroma:
> For future it'd good to also somehow inform user about options that are not available, because right now we're explicitly using the same options as Metro's `start` and `bundle` command, and some of them aren't implemented (yet).
> 
> https://github.com/callstack/repack/blob/270ca1a29bff3dada233b05b96c805453fe5d75d/packages/repack/commands.js#L55-L56

agree 💯, we need to address this in a separate PR though, and most likely after v4 release

thymikee (approved):

PR #539, opened 2024-04-04
feat: OutputPlugin support for empty object as entry

### LINKED ISSUE #361: Feature: OutputPlugin's optional entryName
## Description 
In certain scenarios, there is a requirement to deploy a container separately from any associated app. This necessitates building with an empty object `{}` for the entry configuration. This approach ensures that only the remote entry and related chunks are built. However, this causes Re.Pack's OutputPlugin to throw an exception.

To address this issue, the current workaround involves individually expanding the Re.Pack plugins instead of utilizing Repack.RepackPlugin. This workaround allows for manually passing the entryName to the OutputPlugin, thereby resolving the exception.

## Expected behaviour
The OutputPlugin should handle an empty entry object without throwing an exception

jbroma:
### Summary

Closes #361
 
When supplied with empty object as entry, `OutputPlugin` would fail because we didn't allow for such scenario. Since this scenario is a valid webpack use case (documented [here](https://webpack.js.org/concepts/entry-points/#object-syntax)), the constraint of `entryChunk` being present was removed. 

This feature is especially useful when using `ModuleFederation` - if we only care about the federated container + chunks, then all of our entrypoints declarations come from the `ModuleFederationPlugin`. 

### Test plan

- [x] - TesterApp tests suite pass
- [x] - tested locally with super-app-showcase

### DIFF
--- a/.changeset/early-radios-grow.md
+++ b/.changeset/early-radios-grow.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+OutputPlugin now supports configuration with empty object as compilation entry.
--- a/packages/repack/src/webpack/plugins/OutputPlugin.ts
+++ b/packages/repack/src/webpack/plugins/OutputPlugin.ts
@@ -293,12 +293,6 @@ export class OutputPlugin implements WebpackPlugin {
         }
       }
 
-      if (!entryChunk) {
-        throw new Error(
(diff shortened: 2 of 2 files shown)


thymikee (approved):
do we have any test for that scenario?

RafikiTiki (approved):

PR #552, opened 2024-04-15
fix(TesterApp): resolving symlinks in `resolveNodePackage.gradle` on Windows

szymonrybczak:
### Summary

When I was trying to run `TesterApp` on Windows I saw that paths to symlinked packages weren't resolve correctly, for some reason `.canonicalPath` on Windows doesn't resolve symlink. In this Pull Request I've replace `.canonicalPath` with workaround functions that seems to work on both platforms. 

### Test plan

1. Clone repository
3. Go to `packages/TesterApp`
4. Try running `pnpm android` on Windows

### DIFF
--- a/packages/TesterApp/android/resolveNodePackage.gradle
+++ b/packages/TesterApp/android/resolveNodePackage.gradle
@@ -2,10 +2,20 @@
 // TODO remove in the future when this lands in RN
 
 ext.resolveNodePackage = { String packageName, File startDir ->
-    def candidate = new File(startDir, "node_modules/$packageName")
+    def candidate
+
+    def osName = System.getProperty("os.name").toLowerCase()
+    if (osName.contains("win")) {
+        // for Windows OS
+        candidate = new File(startDir, "node_modules\\$packageName")
+    } else {
+        // for Unix-likely OS (MacOS/Linux)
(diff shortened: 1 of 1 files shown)


szymonrybczak:
@robhogan just FYI we're using here [part](https://github.com/facebook/react-native/pull/42823/files#diff-172086449aec9cb35555f00058ee8cc93714f36aefb1020177efd9406d410f9dR23) from your PR in Core, I know that this Pull Request was reverted some time ago but in case you'll try to work on this once again in future let's keep in mind that `.canonicalPath` is not working in the same way on Windows as on macOS.

jbroma (approved):
LGTM 🎉 Thanks for working on this @szymonrybczak!

robhogan:
> @robhogan just FYI we're using here [part](https://github.com/facebook/react-native/pull/42823/files#diff-172086449aec9cb35555f00058ee8cc93714f36aefb1020177efd9406d410f9dR23) from your PR in Core, I know that this Pull Request was reverted some time ago but in case you'll try to work on this once again in future let's keep in mind that `.canonicalPath` is not working in the same way on Windows as on macOS.

Huh, good to know, and thanks for the heads up!

PR #557, opened 2024-04-16
feat: specify custom version of Re.Pack to install with `@callstack/repack-init`

jbroma:
### Summary

add `--custom-version` flag to `@callstack/repack-init`. This allows for installing other version than the latest one available on the NPM.

Note: error handling is left to the package manager of choice.

### Test plan

- [x] - tested locally


packages/init/src/bin.ts
```diff
@@ -11,10 +11,15 @@ const info = require('../package.json');
 
 const argv = yargs(hideBin(process.argv))
   .usage(`Usage: ${info.name} [options]`)
+  .option('custom-version', {
```
thymikee, line 14: can we change it to just `--version` or will it intercept the version of the program?

jbroma: yeah, `--version` was already taken

thymikee: so how about `--repack-version` or `--exact-version`? I'm thinking about long-term using custom and it feels ambiguous

thymikee (approved):
Some docs around available options would be nice

jbroma:
> Some docs around available options would be nice

oh, good point, I forgot to update the README.md inside of `init`. But separate docs for the init should be added too at some point, good place for a contrib 👍

PR #565, opened 2024-04-18
chore: add 404 redirects to new website

jbroma:
### Summary

when route won't be matched on filesystem, show 404

this required updating `rspress` to `1.18.3` which featured a fix for proper handling of 404s


jbroma:
this required fixing a bug in `@rspress/core`, upstream PR submitted, using patch locally for now

PR #567, opened 2024-04-19
fix(dev-server): should traversal all connected devices when sending event

kerm1it:
### Summary

- When more than two devices are connected, one Android and one iOS, the second connected device will never receive the hot update notification, because the second device is different from the first connected device during traversal, and the loop will be exited directly.  We should continue to traversal next device when the platform of current device is different from  target device.

- Fix delete device don't work when disconnected

### DIFF
--- a/.changeset/hmr-client-prove.md
+++ b/.changeset/hmr-client-prove.md
@@ -1,0 +1,7 @@
+---
+"@callstack/repack-dev-server": patch
+---
+
+should traversal all connected devices when sending event.
+
+delete device don't work when disconnected.
--- a/packages/dev-server/src/plugins/wss/servers/WebSocketHMRServer.ts
+++ b/packages/dev-server/src/plugins/wss/servers/WebSocketHMRServer.ts
@@ -80,17 +80,22 @@ export class WebSocketHMRServer extends WebSocketServer {
     }
 
     const clientId = `client#${this.nextClientId++}`;
-    this.clients.set({ clientId, platform }, socket);
 
-    this.fastify.log.info({ msg: 'HMR client connected', clientId, platform });
+    const client = {
+      clientId,
+      platform,
+    };
+
+    this.clients.set(client, socket);
+
+    this.fastify.log.info({ msg: 'HMR client connected', ...client });
 
     const onClose = () => {
       this.fastify.log.info({
         msg: 'HMR client disconnected',
-        clientId,
-        platform,
+        ...client,
       });
-      this.clients.delete({ clientId, platform });
+      this.clients.delete(client);
     };
 
     socket.addEventListener('error', onClose);
--- a/packages/dev-server/src/plugins/wss/servers/WebSocketHMRServer.ts
+++ b/packages/dev-server/src/plugins/wss/servers/WebSocketHMRServer.ts
@@ -46,7 +46,7 @@ export class WebSocketHMRServer extends WebSocketServer {
         key.platform !== platform ||
         !(clientIds ?? [key.clientId]).includes(key.clientId)
       ) {
-        return;
+        continue;
       }
 
       try {


jbroma:
Hi @kerm1it, first of all great find 🎉 

I found this bug a while ago while working on the rspack prototype, I believe continue worked for webpack nicely, but it didn't for rspack for some reason. It's not something to be concerned about right now anyways.

Can you please test different configurations of platforms and number of devices connected to the dev server to verify that this solution doesn't cause an infinite loop case? (I think that's what I've encountered last time)

kerm1it:
> Can you please test different configurations of platforms and number of devices connected to the dev server to verify that this solution doesn't cause an infinite loop case? (I think that's what I've encountered last time)

I have tested it and connected an Android and iOS device respectively. When changing the code, the hot update of both devices can take effect, and there is no infinite loop.

kerm1it:
Why does an infinite loop occur?

jbroma:
> > Can you please test different configurations of platforms and number of devices connected to the dev server to verify that this solution doesn't cause an infinite loop case? (I think that's what I've encountered last time)
> 
> I have tested it and connected an Android and iOS device respectively. When changing the code, the hot update of both devices can take effect, and there is no infinite loop.

can you please try with multiple ios/android emulators as well ?

kerm1it:
> > > Can you please test different configurations of platforms and number of devices connected to the dev server to verify that this solution doesn't cause an infinite loop case? (I think that's what I've encountered last time)
> > 
> > 
> > I have tested it and connected an Android and iOS device respectively. When changing the code, the hot update of both devices can take effect, and there is no infinite loop.
> 
> can you please try with multiple ios/android emulators as well ?

I also tested connecting two Android and iOS devices at the same time, for a total of four devices, and no problems were found.

jbroma (dismissed):
LGTM 🚀

kerm1it:
I do seem to have an infinite loop problem. 😂

jbroma:
@kerm1it from my research I remember that I ended up with conclusion that our WS servers don't manage dead connections. This causes the `this.clients` field to keep growing in size indefinitely  - I think this is something we could adjust as well if we want to handle this nicely,

kerm1it:
> @kerm1it from my research I remember that I ended up with conclusion that our WS servers don't manage dead connections. This causes the `this.clients` field to keep growing in size indefinitely - I think this is something we could adjust as well if we want to handle this nicely,

I find this problem and fixed it.

kerm1it:
I found that the infinite loop problem I was having was:

When multiple devices are connected, and then another device is connected, and a hot update is requested, an error message is reported: "[HMRClient] Cannot find update - full reload needed", and then it will be reloaded, resulting in a cycle of disconnection and reconnection.

jbroma:
@kerm1it changes look solid 👍 I assume from the context that fixing `onClose` solves the infinite loop problem? One case I'm only worried about is whether `onClose` is reliable enough - I think it might not get called sometime and we might end up with infinite loop anyways, WDYT?

jbroma:
Seems like this isn't going to be such an easy fix. We need to get to the bottom and find the root cause here, perhaps an error in how the flow design. I appreciate all your work so far on this @kerm1it 🙏

kerm1it:
I think what I did is not the root cause of the infinite loop problem. The root cause is that when hot reload is connected, the error message is reported: "[HMRClient] Cannot find update - full reload needed", and then reload app , finally causing hot reload to be in an infinite loop of disconnection and connection. The current approach simply circumvents this problem by allowing only the first client to receive a hot update notification.

So I think the fixes I made exposed the existing problem, and we will find a way to solve it later. Moreover, it is not very common to have many clients connected at the same time.

jbroma:
hey @kerm1it I agree with you 💯, this change only unveiled a problem that was there for a long time. Since we are headed towards a stable release of 4.0.0, we need to address the issue with the loop, I'll do some research myself and see how we can deal with that 👍 

This is a definite pick for 4.0.0 so needs to solved somehow

jbroma:
@kerm1it I've created a diagram of the flow showing the behaviour when the infinite loop happens:

multi-hmr-loop-dark

As of now, I think that inside of `start` command we keep a reference to `lastStats` but that could come from either platform, so that's most likely why it fails.

jbroma:
@kerm1it yep, that was it, I can't reproduce the infinite loop anymore 🎉 I'll submit a separate PR so we can proceed with this one here

kerm1it:
> @kerm1it yep, that was it - `lastStats` were the culprit of the problem, I can't reproduce the infinite loop anymore 🎉 I'll submit a separate PR so we can proceed with this one here

Cool!

kerm1it:
> @kerm1it yep, that was it - `lastStats` were the culprit of the problem, I can't reproduce the infinite loop anymore 🎉 I'll submit a separate PR so we can proceed with this one here

Can you modify it based on my PR? Bringing these fixes with me

jbroma:
@kerm1it #588 is merged, please do the following so we can proceed with this PR:
1. update your branch with newest main, 
2. verify that the loop is no longer the issue
3. add a changeset

kerm1it:
@jbroma  I tested several times on 2 IOS and 2 Android devices. The infinite loop problem has not recurred and hot reloading can work normally.

jbroma (approved):
LGTM 🎉

PR #574, opened 2024-04-22
fix: initalize compiler progress senders eagerly

jbroma:
### Summary

sometimes Compiler workers send progress faster than anticipated, and the progress senders are not properly initialized. This results in compiler crashing and requires user to restart the dev server. 

### Test plan
- [x] - test locally on MacOS
- [x] - test locally on Windows


packages/repack/src/webpack/Compiler.ts
```diff
@@ -158,39 +158,47 @@ export class Compiler extends EventEmitter {
...
-    }
+    if (fileFromCache) return fileFromCache;
+
+    this.addProgressSender(platform, sendProgress);
 
     // Spawn new worker if not already running
     if (!this.workers[platform]) {
       this.workers[platform] = this.spawnWorker(platform);
     } else if (!this.isCompilationInProgress[platform]) {
+      this.removeProgressSender(platform, sendProgress);
```
RafikiTiki, line 189: Is it possible to have a registered progress sender if the compilation is **not** in progress?

RafikiTiki (approved):
Other than the one smol question LGTM

packages/repack/src/webpack/Compiler.ts
RafikiTiki: Oh, right, that makes sense! 🎯

PR #582, opened 2024-04-24
feat: use `localhost` as defined default hostname for the dev-server

jbroma:
### Summary

- RN CLI defaults to `''` as a default hostname.
- Fastify treats `''` as `localhost` but doesn't inform the user that it's listening on both interfaces at once.

Using a defined default allows for displaying both interfaces when starting the server:
```
ℹ [11:38:23.495Z][DevServer] Server listening at http://[::1]:8081
ℹ [11:38:23.499Z][DevServer] Server listening at http://127.0.0.1:8081
```

### Test plan
n/a

### Notes

cherry-picked from #508

### DIFF
--- a/.changeset/fair-experts-wave.md
+++ b/.changeset/fair-experts-wave.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+use `localhost` as a defined default for the dev-server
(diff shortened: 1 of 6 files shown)


thymikee (approved):
Should we change the default in the RNC CLI then? cc @szymonrybczak

szymonrybczak:
> Should we change the default in the RNC CLI then? cc @szymonrybczak

https://github.com/facebook/react-native/pull/44244 It was already done inside `start`'s command actual implementation, created PR to move it to command's config level so that Re.Pack can also access it :)

PR #629, opened 2024-05-26
chore: migrate TesterApp to `react-native-test-app`

szymonrybczak:
### TODO:
- [x] Make sure that Release mode works
- [x] Specify rest options through `app.json` e.g. New Architecture, build modes, schemas.

### Summary

In this Pull Request I've migrated `TesterApp` to `react-native-test-app`.

CleanShot 2024-05-26 at 20 05 38

### Test plan

App should work in the same way as before.


packages/TesterApp/android/settings.gradle
```diff
@@ -1,12 +1,21 @@
-apply from: "./resolveNodePackage.gradle"
```
szymonrybczak, line 1: I know that this is RePack's hack for PNPM to work with React Native apps, since we're right now migrating to RNTA - can we reuse work we can potentially reuse work from `rnx-kit` 🤔

packages/TesterApp/android/gradle.properties
```diff
@@ -34,8 +39,15 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=false
```
szymonrybczak, line 37: any reason why is it `false`?

jbroma: we can enable that, the only thing Re.Pack doesn't support right now is bridgeless mode - from what I remember, on iOS new arch is currently enabled

packages/TesterApp/android/settings.gradle
jbroma: if you can make it work without this that's even better, not sure what else is here to discuss :D

packages/TesterApp/app.json
```diff
@@ -1,4 +1,18 @@
...
+  ],
+  "resources": {
+    "android": [
+      "build/"
+    ],
+    "ios": [
+      "build/"
+    ]
+  }
+}
```
thymikee, line 18: these are RNTA or Expo specific config options?

packages/TesterApp/android/gradle.properties
```diff
@@ -7,22 +7,27 @@
...
+#org.gradle.parallel=true
 
 # AndroidX package structure to make it clearer which packages are bundled with the
 # Android operating system, and which are packaged with your app's APK
 # https://developer.android.com/topic/libraries/support-library/androidx-rn
 android.useAndroidX=true
 # Automatically convert third-party libraries to use AndroidX
 android.enableJetifier=true
+# Jetifier randomly fails on these libraries
+android.jetifier.ignorelist=hermes-android
```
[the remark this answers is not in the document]
jbroma: this is inside of TesterApp, if we can disable jetifier then lets do it 🚀

packages/TesterApp/app.json
jbroma: RNTA

packages/TesterApp/android/gradle.properties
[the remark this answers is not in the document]
szymonrybczak: if all libs are migrated what's blocking us from removing it from the template? :) probably backward compatability, i have no idea how long we should keep it 🤔

https://github.com/react-native-community/template/blob/cd5e21b702ef5c08c74796d19dd72bdb05775d17/template/android/gradle.properties#L24-L25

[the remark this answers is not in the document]
thymikee: For reference: https://github.com/react-native-community/cli/pull/1662

jbroma:
Needed to add a bootstrap HOC as a workaround for development - apparently anything done before registering the main component causes native modules initialization to fail for some reason?

tido64:
> Needed to add a bootstrap HOC as a workaround for development - apparently anything done before registering the main component causes native modules initialization to fail for some reason?

@jbroma: Do you have repro steps? I reverted badea9d but it seems to work fine?

image

packages/TesterApp/react-native.config.js
```diff
@@ -1,3 +1,20 @@
...
+    });
+  } catch (_) {
+    return undefined;
+  }
+})();
+
 module.exports = {
+  ...(project ? { project } : undefined),
   commands: require('@callstack/repack/commands'),
 };
```
tido64, line 20: Since you're not publishing TesterApp, you can simplify the config:
```suggestion
const { configureProjects } = require('react-native-test-app');

module.exports = {
  project: configureProjects({
    android: {
      sourceDir: 'android',
    },
    ios: {
      sourceDir: 'ios',
    },
  }),
  commands: require('@callstack/repack/commands'),
};
```

jbroma:
@tido64 I've narrowed it down to a more specific scenario, here are steps to repro:

```bash
cd packages/TesterApp
# start a fresh instance of dev-server
pnpm start
# in a different pane/window
pnpm android
````
First load should cause the issue as shown above (the one with AsyncStorage being null)

File: packages/TesterApp/ios/Podfile [hunk 6]
```diff
@@ -1,41 +1,15 @@
...
 
-linkage = ENV['USE_FRAMEWORKS']
-if linkage != nil
-  Pod::UI.puts "Configuring Pod with #{linkage}ally linked Frameworks".green
-  use_frameworks! :linkage => linkage.to_sym
-end
+options = {
+  :bridgeless_enabled => false,
+  :fabric_enabled => true,
+  :hermes_enabled => true,
```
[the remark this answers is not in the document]
jbroma: I think it's setting it to true, copied this config from bridgeless PR in RNTA, so maybe it was there just as an example 😅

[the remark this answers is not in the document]
tido64: Some context here: https://github.com/microsoft/react-native-test-app/issues/1797#issuecomment-2035260695

szymonrybczak:
Added script:
→ `pnpm run ios:release` - should bundle and app in Release mode
→ `pnpm run android:release` - should bundle and app with `./assembleRelease`

packages/TesterApp/scripts/release.js
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    execSync(command, { stdio: 'inherit' });
+  } catch (error) {
+    console.error(`Error executing command: ${command}`, error);
+    process.exit(1);
+  }
+}
+
+function buildIOS() {
+  runCommand('pnpm bundle:ios');
```
jbroma, line 13: let's clean the build output dir for ios here - only `build/generated` is cleaned everytime we build

packages/TesterApp/scripts/release.js
```diff
@@ -0,0 +1,35 @@
...
+  process.chdir('./ios');
+  runCommand('bundle exec pod install');
+  runCommand('npx react-native run-ios --mode "Release"');
+}
+
+function buildAndroid() {
+  console.log('Building Android...');
```
jbroma, line 20: let's clean the build output dir for android here as well (and perhaps also remove the console log)

jbroma (approved):
GJ! small nits, except for that I think we finally got this ready to be merged 💪

packages/TesterApp/scripts/release.js
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    execSync(command, { stdio: 'inherit' });
+  } catch (error) {
+    console.error(`Error executing command: ${command}`, error);
+    process.exit(1);
+  }
+}
+
+function buildIOS() {
+  runCommand('rm -rf build/generated');
```
jbroma, line 13: uhm, let's ensure the dir before we run `rm -rf`, or maybe force is not necessary here?

jbroma: btw this can be moved outside of the function to the startup code down below

jbroma (changes requested):

packages/TesterApp/scripts/release.js
tido64: Can we use `fs.rm` instead of invoking a shell command?

jbroma (approved):
LGTM 🎉

PR #630, opened 2024-05-28
fix: hermes plugin node version

jbroma:
### Summary

use `execa.node` to reuse the node version from env instead of calling `node` - this mitigates the issue where sometimes node can't be found when building through Xcode

### Test plan

- [x] - iOS builds
- [x] - Android builds


szymonrybczak (approved):
👍

Issue #623, opened 2024-05-22
ChunkManagerPackage: No value passed for parameter 'p6'


krzysztofzaluski:
## Environment

RePack: 4.0.0
React Native: 0.71.7

## Description

I get an error while building the application with the "react-native run-android --no-packager" command:
```
/node_modules/@callstack/repack/android/src/main/java/com/callstack/repack/ChunkManagerPackage.kt:31:13 No value passed for parameter 'p6'
```

jbroma:
Hey @krzysztofzaluski

p6 is `isTurboModule`, it comes from this gradle function:
```groovy
def isNewArchitectureEnabled() {
  return rootProject.hasProperty("newArchEnabled") && rootProject.getProperty("newArchEnabled") == "true"
}
```

I'm guessing you don't have a variable named `newArchEnabled` in your `<project>/android/gradle.properties`. Adding that there should fix this issue. Nevertheless, we should fix this by providing a fallback value of `false` in this case.

krzysztofzaluski:
hej @jbroma 

I got this line in my `android/gradle.properties` file and it's set to `false`:

```
newArchEnabled=false
```

krzysztofzaluski:
@jbroma 
I started snooping around in the codes and noticed something like this:
`ReactModuleInfo.class`:
```
    public ReactModuleInfo(String name, String className, boolean canOverrideExistingModule, boolean needsEagerInit, boolean hasConstants, boolean isCxxModule, boolean isTurboModule) {
        this.mName = name;
        this.mClassName = className;
        this.mCanOverrideExistingModule = canOverrideExistingModule;
        this.mNeedsEagerInit = needsEagerInit;
        this.mHasConstants = hasConstants;
        this.mIsCxxModule = isCxxModule;
        this.mIsTurboModule = isTurboModule;
    }
```

it has 7 params but you are passing 6 params here `getReactModuleInfoProvider()`:
```
            moduleInfos[ScriptManagerModule.NAME] = ReactModuleInfo(
                ScriptManagerModule.NAME,
                ScriptManagerModule.NAME,
                false,  // canOverrideExistingModule
                true,  // needsEagerInit
                false,  // isCxxModule
                isTurboModule // isTurboModule
            )
```

so there is an error: `No value passed for parameter 'isTurboModule'`

krzysztofzaluski:
@jbroma ok, but what value should be passed there as `hasConstants`?

jbroma:
[...]

PR #633, opened 2024-05-29
feat: support bridgeless

jbroma:
### Summary

add support for bridgeless in new arch for RN >= 0.74 with compat for all other variants. RN 0.74 introduced way more compatibility layers to handle bridgeless on new arch which was not possible on 0.73. Since there is no way to get into the runtime on bridgeless on 0.73, it is excluded from support (it was not turned on by default before 0.74 so it's a rare scenario). This seems to be rather common amongst other libraries having similar issues.

### Notes

- CPP Native Turbomodule would be nice but we still need to support old arch, at least for now
- Implemented using `CallInvoker` API on both platforms
- Android uses `BridgelessCatalystInstance` since it was introduced in `0.74` - ~~thanks to that we don't need to fork the code~~ needed to fork the code on Android because of `FrameworkAPI` requirement that causes compilation error 😔

### Test plan

| Platform                  | 0.74 | 0.73 | 0.72 |
|---------------------------|------|------|------|
| Android OldArch           | ✅ | ✅ |   ✅    |
| Android NewArch           | ✅ | ✅ |  ✅     |
| Android NewArch Bridgeless| ✅ |    ❌  |  -   |
| iOS OldArch               | ✅ | ✅ |   ✅   |
| iOS NewArch               | ✅ | ✅ |   ✅    |
| iOS NewArch Bridgeless    | ✅ |   ❌   |    -  |


jbroma:
Everything looks green, except for the RN 0.73 bridgeless which is expected. 

I've also fixed multiple issues on Android:
- Kotlin `jvmTarget` incosistencies
- used backwards compatible `ReactModuleInfo` constructor (issue from #623)
- marked `newArchEnabled` as cache invalidating

jbroma:
Canary build for testing available here: https://www.npmjs.com/package/@callstack/repack/v/4.0.1-canary-20240603172122

packages/TesterApp/ios/TesterApp/AppDelegate.mm
```diff
@@ -26,10 +26,9 @@ - (NSURL *)bundleURL
   return [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
 #endif
 }
-// Bridgeless mode is enabled by default in RN>=0.74 but we don't support it yet.
 - (BOOL)bridgelessEnabled
 {
-    return NO;
+    return YES;
 }
```
okwasniewski, line 32: I think this can be removed, right? As far as I understand there is no need to support new arch without bridgeless on 0.74 now

packages/repack/android/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+set(CMAKE_VERBOSE_MAKEFILE ON)
+set(CMAKE_CXX_STANDARD 17)
+
+# Add callstack-repack sources
+add_library(
+        ${PACKAGE_NAME}
+        SHARED
+        src/main/cpp/OnLoad.cpp
+        src/main/cpp/NativeScriptLoader.h
+        src/main/cpp/NativeScriptLoader.cpp
```
okwasniewski, line 15: nit: I think you can provide the path to the directory which should be linked

packages/repack/android/src/main/cpp/NativeScriptLoader.cpp
```diff
@@ -0,0 +1,46 @@
+#include <fbjni/fbjni.h>
+#include "NativeScriptLoader.h"
+
+void NativeScriptLoader::registerNatives() {
+    registerHybrid({
+                           makeNativeMethod("evaluateJavascriptAsync",
+                                            NativeScriptLoader::evaluateJavascriptAsync)});
+}
```
okwasniewski, line 8: I think there is something wrong with formatting

packages/repack/android/src/main/cpp/NativeScriptLoader.cpp
```diff
@@ -0,0 +1,46 @@
...
+        jni::alias_ref<JArrayByte> code,
+        jni::alias_ref<JString> url,
+        jni::alias_ref<JObject::javaobject> promise
+) {
+    // promise gets passed into callback for invokeAsync,
+    // we need to retain the reference longer than the scope of this function
+    auto promiseRef = make_global(promise);
+    // no type for Promise, extract the methods manually
+    static const auto resolve = promiseRef->getClass()->getMethod<void(jobject)>("resolve");
+    static const auto reject = promiseRef->getClass()->getMethod<void(jstring, jstring)>("reject");
```
okwasniewski, line 23: nit: minor perf improvement, you could call `promiseRef->getClass()` once

packages/repack/ios/ScriptConfig.h
```diff
@@ -17,16 +17,16 @@
...
+- (nonnull ScriptConfig *)initWithScript:(nonnull NSString*)scriptId
+                         withURL:(nonnull NSURL*)url
+                      withMethod:(nonnull NSString*)method
+                       withQuery:(nullable NSString*)query
+                       withFetch:(BOOL)fetch
+                    withAbsolute:(BOOL)absolute
+                     withHeaders:(nullable NSDictionary*)headers
+                        withBody:(nullable NSData*)body
+                     withTimeout:(nonnull NSNumber*)timeout
+       withVerifyScriptSignature:(nonnull NSString*)verifyScriptSignature;
```
okwasniewski, line 29: You can actually only specify what's nullable by using `NS_ASSUME_NONNULL_BEGIN` (where it assumes everything is nonnull by default) and you only specify nullable stuff. I thinki having all of these annotations makes the code less readable. 

Here is a reference: https://developer.apple.com/documentation/swift/designating-nullability-in-objective-c-apis

packages/repack/ios/ScriptManager.mm
```diff
@@ -2,23 +2,26 @@
 #import "ScriptConfig.h"
 #import "ErrorCodes.h"
 
+// make it work with use_frameworks!
```
okwasniewski, line 5: Maybe it's worth to create a GH issue and leave a todo?

packages/repack/ios/ScriptManager.mm
```diff
@@ -257,12 +251,55 @@ - (void)executeFromFilesystem:(RCTCxxBridge *)bridge
...
+    // we obtain the ptr here, and use it inside of invokeAsync callback
+    // TODO: Consider doing this during initialization step once to fail fast
+    facebook::jsi::Runtime *runtime = [self getJavaScriptRuntimePointer];
+    if(!runtime) {
+        reject(RuntimeUnavailableError, @"Can't access React Native runtime", nil);
+        return;
+    }
+    
+    std::string source(static_cast<const char*>([code bytes]), [code length]);
+    std::string sourceUrl([[url absoluteString] UTF8String]);
```
okwasniewski, line 291: nit: I think more modern way to init stuff is with `{}` `std::string sourceUrl { [[url absoluteString] UTF8String] };`

okwasniewski (commented):
Great work! I've left few minor comments

packages/TesterApp/ios/TesterApp/AppDelegate.mm
jbroma: yeah absolutely, just left it there for visibility so we can switch on demand for some edge cases

packages/repack/ios/ScriptManager.mm
jbroma: I've just added an explicit comment here, the fix was done with different PR :)

packages/repack/ios/ScriptManager.mm
jbroma: oh yeah sure! apparently this (list initializer) disables the implicit conversion so the compiler can catch more errors this way! nice find!

jbroma:
cpp & objc is now reformatted with `clang-format` 🎉

okwasniewski (approved):

PR #637, opened 2024-06-04
chore: add clang-format

jbroma:
### Summary

added clang-format for formatting objc and cpp code in repack
format config comes from reanimated lib

### Test plan

n/a


thymikee:
Nothing changed in the code, was this run or everything was already formatted?

jbroma:
just added clang-format here, no reformatting was done

PR #639, opened 2024-06-07
fix: native build config for Android

jbroma:
### Summary

- [x] fixed CMake config
- [x] removed buildTools version from config - was causing a warning and other libs don't use this field anymore

### Test plan

- [x] - TesterApp android builds


szymonrybczak (approved):
Tested and app works correctly now! 🚀

PR #641, opened 2024-06-09
fix: native module typing

### LINKED ISSUE #494: Improve types for `prefetchScript` and `loadScript`
Here you can find a typescript spec used by react-native-codegen: https://github.com/callstack/repack/pull/495/files#diff-a063c248295c855435089dd4f53efe44eb82c9b2a8b70470fb638ce5d3d037e4R4 

In this spec, we use `Object` which is not meaningful for the developer. Please go to the native implementation here https://github.com/callstack/repack/blob/b0b8613eab82641e91607e0931eafc2360ff42a2/packages/repack/android/src/main/java/com/callstack/repack/ChunkManagerModule.kt#L24 and replace all `Objects` with the appropriate type. 

_Originally posted by @troZee in https://github.com/callstack/repack/pull/344#discussion_r1173627173_

jbroma:
### Summary

Closes #494 

### Test plan

- [x] - android new arch builds
- [ ] - ios new arch builds

### DIFF
--- a/packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
@@ -1,19 +1,19 @@
 import type { TurboModule } from 'react-native';
 import { TurboModuleRegistry } from 'react-native';
 
-export const enum NormalizedScriptLocatorMethod {
+export const enum NormalizedScriptLocatorHTTPMethod {
   GET = 'GET',
   POST = 'POST',
 }
 
-export const enum NormalizedScriptLocatorVerifyScriptSignature {
+export const enum NormalizedScriptLocatorSignatureVerificationMode {
   STRICT = 'strict',
   LAX = 'lax',
   OFF = 'off',
 }
 
 export interface NormalizedScriptLocator {
-  method: NormalizedScriptLocatorMethod;
+  method: NormalizedScriptLocatorHTTPMethod;
   url: string;
   fetch: boolean;
   timeout: number;
@@ -21,7 +21,7 @@ export interface NormalizedScriptLocator {
   query: string | null;
   headers: { [key: string]: string } | null;
   body: string | null;
-  verifyScriptSignature: NormalizedScriptLocatorVerifyScriptSignature | null;
+  verifyScriptSignature: NormalizedScriptLocatorSignatureVerificationMode | null;
 }
 
 export interface Spec extends TurboModule {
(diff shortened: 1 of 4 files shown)


jbroma:
It seems that for Android, the generated type stays as `ReadableMap`, but for iOS the codegen creates a struct for `config` argument. I'm not sure whether it's worth it to implement that, when we still have to keep compatibility for the old arch - we would have to fork a lot of code just to use the typings from codegen.

cc @troZee

packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
```diff
@@ -1,11 +1,36 @@
...
+  OFF = 'off',
+}
+
+export interface NormalizedScriptLocator {
+  method: NormalizedScriptLocatorHTTPMethod;
+  url: string;
+  fetch: boolean;
+  timeout: number;
+  absolute: boolean;
+  query: string | null;
```
troZee, line 21: [q]: Do you know if it is intended to use null here?

jbroma: codegen supports nullable values, I've went the way of explicitly declaring nulls rather than using optional values here - and query can be empty

packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
```diff
@@ -1,11 +1,36 @@
 import type { TurboModule } from 'react-native';
 import { TurboModuleRegistry } from 'react-native';
 
-// TODO FIX TYPES https://github.com/callstack/repack/issues/494
+export const enum NormalizedScriptLocatorHTTPMethod {
+  GET = 'GET',
+  POST = 'POST',
+}
+
+export const enum NormalizedScriptLocatorSignatureVerificationMode {
```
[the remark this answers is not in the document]
jbroma: yeah, string literals are not supported by codegen but enums are 👍
it gets converted to NSString in this case which is what we want, this is the output of codegen for ios:

```objc
namespace JS {
  namespace NativeScriptManager {
    struct ScriptConfig {
      NSString *method() const;
      NSString *url() const;
      bool fetch() const;
      double timeout() const;
      bool absolute() const;
      NSString *query() const;
      id<NSObject> _Nullable headers() const;
      NSString *body() const;
      NSString *verifyScriptSignature() const;

      ScriptConfig(NSDictionary *const v) : _v(v) {}
    private:
      NSDictionary *_v;
    };
  }
}
```

got the idea for that here: https://github.com/mrousavy/react-native-mmkv/blob/cae0b9c75c7c3f3a9e71da551551e1214f520c39/package/src/NativeMmkv.ts#L10

troZee (approved):

Issue #538, opened 2024-04-04
Adjust the description of `start` and `bundle` commands


jbroma:
## Describe the feature

Adjust the description of commands so that they don't include unsupported flags

## Motivation

Currently, we copy the the command descriptions from CLI commands and add our own flags. We should also adjust the copied descriptions, so that the `--help` output is up to date.

## Related Issues

proposed by @szymonrybczak  in #537

szymonrybczak:
Is this issue still valid? What option do we miss inside Re.Pack CLI commands? 🤔

jbroma:
It's still valid, hard to tell what's missing, haven't looked at that yet - we simply need to go through every flag and check if it makes sense in Re.Pack and/or add missing ones (specific to Re.Pack)

barttom:
@jbroma @szymonrybczak 
Hey guys!
I've checked all options in the start command and a few in the bundle command. Looks like every option works fine so I;m confused, about how I can investigate that

barttom:
@jbroma @szymonrybczak  when You have a moment :) 
https://github.com/callstack/repack/pull/656

PR #656, opened 2024-07-01
refactor: remove unsuported commands from doc

barttom:
### Summary
closes https://github.com/callstack/repack/issues/538

Keep in the doc only options that we support in `start` and `bundle` command

### Test plan

[x] Test locally

image


packages/repack/commands.js
```diff
@@ -1,5 +1,6 @@
 const path = require('path');
 const { createRequire } = require('module');
+const { startCommandOptions } = require('./src/commands/utils/commandsOptions');
```
jbroma, line 3: when the package is built, this will still point to src and target TS file and fail after evaluation - it should point to `dist` instead, where JS will be stored. But then we will have the same problem in dev - I think we just need to inline the commandsOptions instead of having a separate file for that

jbroma (changes requested):

jbroma:
cc @szymonrybczak WDYT about having an explicit set of rules instead of copying from the CLI?

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
 }
 
+const startCommandOptions = [
+  {
+    name: '--cert <path>',
+    description: 'Path to custom SSL cert',
+  },
+  {
+    name: '--host <string>',
+    description: '(default: "")',
```
jbroma, line 60: needs fixing - base this on desc from CLI

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+  },
+  {
+    name: '--http',
```
jbroma, line 63: I think this should be `https`

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    description: 'Disables interactive mode',
+  },
+  {
+    name: '--silent',
+    description: 'Silents all logs to the console/stdout',
+  },
+  {
+    name: '--experimental-debugger',
+    description:
+      "[Experimental] Enable the new debugger experience and 'j' to debug. This enables the new frontend experience only: connection reliability and some basic features are unstable in this release.",
```
jbroma, line 85: not sure we support `j` to debug here, can we double check on this?

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    name: '--port <number>',
+    description: 'The port number that runs the server on',
+  },
+  {
+    name: '--no-interactive',
+    description: 'Disables interactive mode',
+  },
+  {
+    name: '--silent',
+    description: 'Silents all logs to the console/stdout',
```
jbroma, line 80: is this working as expected?

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    description: 'Enables verbose logging',
+  },
+  {
+    name: '--json <statsFile>',
+    description: 'Stores stats in a file.',
+    parse: (val) => path.resolve(val),
+  },
+  {
+    name: '--stats <preset>',
+    description: 'It instructs Webpack on how to treat the stats e.g. normal',
```
jbroma, line 160: it would be nice to point user to a place where he can find possible presets - maybe we can also list them in the description here?

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+  },
+  {
+    name: '--log-file <string>',
+    description: 'Enables file logging to specified file',
+  },
+];
+
+const bundleCommandOptions = [
+  {
+    name: '--assets-dest <string>',
```
jbroma, line 111: lookin at other commands, we should use something like `path` instead of `string` here

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    description:
+      'Directory name where to store assets referenced in the bundle',
+  },
+  {
+    name: '--entry-file <path>',
+    description:
+      'Path to the root JS file, either absolute or relative to JS root',
+  },
+  {
+    name: '--minify [boolean]',
```
jbroma, line 121: i think `[boolean]` can be omitted here - every boolean flag works like this AFAIK

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    description:
+      'Allows overriding whether bundle is minified. This defaults to false if dev is true, and true if dev is false. Disabling minification can be useful for speeding up production builds for testing purposes.',
+  },
+  {
+    name: '--dev [boolean]',
+    description:
+      'If false, warnings are disabled and the bundle is minified (default: true)',
```
jbroma, line 128: is this the desc from CLI? can be improved

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+  },
+  {
+    name: '--bundle-output <string>',
```
jbroma, line 131: let's replace string with `path` or `filename`

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    description:
+      'File name where to store the resulting bundle, ex. /tmp/groups.bundle',
+  },
+
+  {
+    name: '--sourcemap-output <string>',
```
jbroma, line 137: let's replace string with `path` or `filename`

jbroma (changes requested):
mostly nits, let's make it as consistent as possible 🙏

packages/repack/commands.js
szymonrybczak, line 60: can we also for default values use `default` key, same as [here](https://github.com/szymonrybczak/react-native/blob/28357a3eb6fec8699b3d95212d160a9ea2892d4d/packages/community-cli-plugin/src/commands/bundle/index.js#L42)?  `commander` will automatically format these

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+  },
+  {
+    name: '--experimental-debugger',
+    description:
+      "[Experimental] Enable the new debugger experience and 'j' to debug. This enables the new frontend experience only: connection reliability and some basic features are unstable in this release.",
+  },
+  {
+    name: '--verbose',
+    description: 'Enables verbose logging',
+  },
```
szymonrybczak, line 90:
```suggestion

```

`--verbose` is automatically added by RNC CLI, right now `--verobse is presented two times in the `--help` output.

packages/repack/commands.js
```diff
@@ -50,6 +50,117 @@ function getCommands() {
...
+    description: 'Either "ios" or "android" (default: "ios")',
+  },
+  {
+    name: '--reset-cache',
+    description: 'Removes cached files (default: false)',
+  },
+  {
+    name: '--verbose',
+    description: 'Enables verbose logging',
+  },
```
szymonrybczak, line 152:
```suggestion

```

https://github.com/callstack/repack/pull/656/files#r1661095636

packages/repack/commands.js
barttom: looks like we just support experimental debugger without handling 'j' key

packages/repack/commands.js
barttom: yup, no logs from server with this switch
image

packages/repack/commands.js
szymonrybczak: so maybe let's add support for handling `j` in separate PR, I think it won't be a big effort but will sync behaviour with Metro dev server 👍

jbroma: ok, let's remove the 'j' to debug for now, and we will add that in a separate PR

packages/repack/commands.js
barttom: Could You precise, how to improve that?

jbroma: - remove the `[boolean]` in the name
- desc could be worded like: `Disables development warnings and production optimisations (default: true)`

perhaps we can put the whole thing through GPT to enhance the descriptions? WDYT?

barttom: I think this is confusing
```
{
    name: '--dev',
    description: 'Disables development warnings and production optimisations',
    default: true,
  },
```
it suggests that `--dev` flag disables dev warnings, but to achieve that, user should set `--dev false`
so maybe we should keep the param in the name?
{
    name: '--dev [boolean]' ,
    description: 'Disables development warnings and production optimisations',
    default: true,
  },

jbroma: oh yeah absolutely 💯, should be `Enables development warnings and disables production optimisations`

barttom:
@jbroma @szymonrybczak comments addressed :)

jbroma (changes requested):
all looks great 🚀, last few questions down below.

please also do the following:
- rename the PR to follow the convention, and change from feat to refactor
- add a `patch` changeset describing the changes

packages/repack/commands.js
```diff
@@ -50,6 +50,120 @@ function getCommands() {
...
+  {
+    name: '--minify',
+    description:
+      'Allows overriding whether bundle is minified. This defaults to false if dev is true, and true if dev is false. Disabling minification can be useful for speeding up production builds for testing purposes.',
+  },
+  {
+    name: '--dev [boolean]',
+    description:
+      'Enables development warnings and disables production optimisations',
+    default: true,
```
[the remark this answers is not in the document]
barttom: yup

packages/repack/commands.js
```diff
@@ -50,6 +50,120 @@ function getCommands() {
...
+    description:
+      'It instructs Webpack on how to treat the stats:\n' +
+      "'errors-only'  - only output when errors happen\n" +
+      "'errors-warnings' - only output errors and warnings happen\n" +
+      "'minimal' - only output when errors or new compilation happen\n" +
+      "'none' - output nothing\n" +
+      "'normal' - standard output\n" +
+      "'verbose' - output everything\n" +
+      "'detailed' - output everything except chunkModules and chunkRootModules\n" +
+      "'summary' - output webpack version, warnings count and errors count",
```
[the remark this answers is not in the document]
barttom: image

jbroma (approved):
LGTM, thanks for taking care of this @barttom 🎉

szymonrybczak (approved):
:shipit:

PR #661, opened 2024-07-03
docs: start and bundle commands documentation pages

barttom:
### Summary
Added  documentation pages for `start` and `bundle` commands

### Test plan
[x] tested locally

https://github.com/callstack/repack/assets/7682108/258249cf-1fc4-40e4-b418-c812065c2ad7


jbroma (changes requested):
the content looks good! let's just reorganise it slightly - let's combine start and bundle commands into one page named 'repack' that will live alongside of `repack-init`.

Also, if you want to ensure order of pages, you can add a `_meta.json` like in `website/src/4.x/docs/code-splitting/_meta.json`. The name of doc is kept as a part of URL so let's avoid numbering them 👍

barttom:
> the content looks good! let's just reorganise it slightly - let's combine start and bundle commands into one page named 'repack' that will live alongside of `repack-init`.
> 
> Also, if you want to ensure order of pages, you can add a `_meta.json` like in `website/src/4.x/docs/code-splitting/_meta.json`. The name of doc is kept as a part of URL so let's avoid numbering them 👍

Don't You think this is too much information in one page? Also nesting navigation is not a best practice for UX, we should avoid them if it possible. So I think is better to keep each command separately. 

`start` and `bundle` are also cli command and the context of usage is explained in the description, so I'm not sure if nesting is necessary in this case.

To highlight 'init' command I;ve order it first, so I think is simple and clear way to show difference between context of particular command.

jbroma:
> > the content looks good! let's just reorganise it slightly - let's combine start and bundle commands into one page named 'repack' that will live alongside of `repack-init`.
> > Also, if you want to ensure order of pages, you can add a `_meta.json` like in `website/src/4.x/docs/code-splitting/_meta.json`. The name of doc is kept as a part of URL so let's avoid numbering them 👍
> 
> Don't You think this is too much information in one page? Also nesting navigation is not a best practice for UX, we should avoid them if it possible. So I think is better to keep each command separately.
> 
> `start` and `bundle` are also cli command and the context of usage is explained in the description, so I'm not sure if nesting is necessary in this case.
> 
> To highlight 'init' command I;ve order it first, so I think is simple and clear way to show difference between context of particular command.

hmmm yeah, I wasnt particulary conviced about that as well, WDYT about adding a divider to separate `repack-init` from the rest? you can do this by adding this, like in `website/src/4.x/docs/_meta.json`:
```json
{
    "type": "divider"
}
```

barttom:
@jbroma good idea. It keeps flat navigation and highlights different context. I've pushed changes
image

jbroma (approved):
looks great, LGTM! Thank you for your contribution @barttom 🎉🚀

PR #662, opened 2024-07-03
chore: fixed filtering in website commands

barttom:
### Summary
Fixed wrong filtering in website commands
image
 

### Test plan
[x] Tested locally


jbroma (approved):
thanks, nice find!

Issue #654, opened 2024-06-25
How to locate cause of Error: ENOENT: no such file or directory


krzysztofzaluski:
I've noticed that if there is an error in the component code or some configuration file, such as missing a library in `package.json` and if I'm using it in the component or config file, then it always throws me an error:

```
[Error: ENOENT: no such file or directory, open '/Users/krzysztofzaluski/Documents/react-native/Project/build/generated/android/index.bundle'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/krzysztofzaluski/Documents/react-native/Project/build/generated/android/index.bundle'
}
```

so the problem is that the error message doesn't say anything specific that could help to locate the real issue, like the code line where the old removed import is still present, etc.

are there some ways how to track such a cases and locate such a files with issues?

krzysztofzaluski:
to complete the question I would like to add that my solution e.g. for missing imports is to scan the project with eslint with tsc but maybe there are still some cases where this error `Error: ENOENT: no such file or directory ` can appear and maybe there is a chance for the console to display a more precise problem?

jbroma:
Hi @krzysztofzaluski,

thanks for pinpointing this, we're not reporting compilation errors when running the dev server (as in metro, which display an LogBox error for that), so this would need to be implemented to spot this early on. 

When bundling for production this is most likely visible in the output - can you confirm this is the case?

aw-capgemini:
Hi @jbroma I'm encountering the same issue after upgrading from repack v3 to v4  when bundling for production

### Description
I encountered an error while running the `pnpm bundle:ios` command using Repack. The error message indicates that a required file is missing.

### Error Message
[Error: ENOENT: no such file or directory, open '/Users/cgemployee/Projects/rahul/Digico/packages/host/build/generated/ios/index.bundle'] {
errno: -2,
code: 'ENOENT',
syscall: 'open',
path: '/Users/cgemployee/Projects/rahul/Digico/packages/host/build/generated/ios/index.bundle'
}
ELIFECYCLE  Command failed with exit code 2.


### Steps to Reproduce
1. Run the command `pnpm bundle:ios`
2. Observe the error message

### Expected Behavior
The bundling process should complete successfully and generate the `index.bundle` file.

### Environment
- **Repack Version**: v4
- **React Native Version**: 0.72.0
- **Node Version**: 19
- **Operating System**: macOS

### Additional Context
[...]

PR #671, opened 2024-07-23
Fixing issue where entry chunk is not added to local chunks if another chunk is processed first.

sleeper-luke:
### Summary

After upgrading to repack 4, I noticed that after webpack generated and built my app, despite webpack completing successfully android builds were no longer copying the main javascript bundle to `android/app/build/generated/assets/`. I believe this is the same symptom as #654. 

After digging into the code, it turns out that when the OutputPlugin was processing chunks, on android it would just so happen to process one of my remote chunks first that had a dependency on the entry chunk. Since that dependency existed, the entryChunk would get added to `sharedChunks` before it had a chance to be added to `localChunks`, preventing it from being copied to the output folder properly.

### Test plan

- [ ] Have an app with remote chunks that reference the main entry chunk.
- [ ] Run `webpack-bundle` on android to build an android release build.
- [ ] Ensure the destination android folder (`android/app/build/generated/assets/` in my case) contains the main javascript bundle.
- [ ] (Verify other configurations continue to work)

### DIFF
--- a/packages/repack/src/webpack/plugins/OutputPlugin.ts
+++ b/packages/repack/src/webpack/plugins/OutputPlugin.ts
@@ -263,6 +263,7 @@ export class OutputPlugin implements WebpackPlugin {
       const entryChunk = compilationStats.chunks!.find((chunk) => {
         return chunk.initial && chunk.names?.includes(entryChunkName);
       });
+      localChunks.push(entryChunk!);
 
       for (const chunk of compilationStats.chunks!) {
         // Do not process shared chunks right now.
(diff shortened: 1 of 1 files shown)


jbroma:
Hi @sleeper-luke, sorry for the delay and thanks for your contribution!

I took a look at this PR quite a few times but couldn't determine whether this was the correct going forward. 

I went ahead and refactored the logic of Output Plugin's chunk categorisation algorithm in #678. 
This should serve a nice starting point for explicitly defining the spec for what gets copied and what not.

The PR there includes a [canary release](https://www.npmjs.com/package/@callstack/repack/v/4.2.0-canary-20240802125253) which you can install and test: 

`npm install @callstack/repack@4.2.0-canary-20240802125253`

If you are still facing the issue from the PR, let me know whether this fixes it 🙏

jbroma:
Hi @sleeper-luke, we've just released [Re.Pack 4.2.0](https://github.com/callstack/repack/releases/tag/%40callstack%2Frepack%404.2.0), please see if it solves your issues with OutputPlugin

jbroma:
Closing since the PR is stale, and most likely already fixed in current stable versions

PR #674, opened 2024-07-26
feat: ModuleFederation 2.0 support

### LINKED ISSUE #518: Adopt the new federation runtime
## Describe the feature

I have recently redesigned module federation and introduced the concept of runtime plugins, while they have many uses - i think the new runtime could be a prime candidate for repack as i recall we had to do some special things for ChunkManager, this could likely be moved into a runtime plugin and not require a full webpack plugin for federation to work (i do the same for node.js now too) 

## Motivation

More flexability to end user and most likely a cleaner implementation of federation for RN apps

jbroma:
### Summary

Closes #518

- [x] - introduce separate tester for MF2
- [x] - rename `ModuleFederationPlugin` to `ModuleFederationPluginV1`
- [x] - keep `ModuleFederationPlugin` as wrapper for MF1 with wrapper to make this a non-breaking change
- [x] - add `ModuleFederationPluginV2`
- [x] - add `@deprecated` to all federated utils

### Test plan

- [x] - unit tests for MF2 plugin
- [x] - tester federation V1 works
- [x] - tester federation V2 works


apps/tester-federation/package.json
```diff
@@ -9,13 +9,15 @@
     "start:miniapp": "react-native webpack-start --webpackConfig webpack.config.mini-app.mjs --port 8082"
   },
   "dependencies": {
-    "react": "18.2.0",
-    "react-native": "0.74.3",
+    "@callstack/repack": "workspace:*",
+    "@module-federation/enhanced": "^0.3.2",
+    "@module-federation/runtime": "^0.3.2",
```
ScriptedAlchemy, line 14: You don’t need runtime installed unless. Enhanced has it and injects. Unless you’re using runtime outside of the build scope.

apps/tester-federation/src/host/screens/MiniAppScreen.tsx
```diff
@@ -1,10 +1,7 @@
 import React from 'react';
-import { Federated } from '@callstack/repack/client';
 import { ActivityIndicator, StyleSheet, View } from 'react-native';
 
-const MiniAppNavigator = React.lazy(() =>
-  Federated.importModule('MiniApp', './MiniAppNavigator')
-);
+const MiniAppNavigator = React.lazy(() => import('MiniApp/MiniAppNavigator'));
```
ScriptedAlchemy, line 4: Yusss good riddance custom loader in userland! This is what I was hoping to see.

patches/@module-federation__sdk@0.3.2.patch
```diff
@@ -0,0 +1,78 @@
+diff --git a/dist/index.cjs.js b/dist/index.cjs.js
+index 5ac66502f7e3230335fbab7bcf67ecb9614f938f..31ccc081db9aa2b6d20f7a66e5cd67f42f8cc1fe 100644
+--- a/dist/index.cjs.js
++++ b/dist/index.cjs.js
+@@ -604,66 +604,14 @@ function createScriptNode(url, cb, attrs, createScriptHook) {
+             url = hookResult.url;
+         }
+     }
+-    let urlObj;
```
ScriptedAlchemy, line 9: No need to patch. You can PR my repo and add a react-native.ts to sdk. Then do whatever you need to do that can detect RN vs node and we can ship it under official.

apps/tester-federation/package.json
jbroma: Oh yeah, probably just a leftover before I realised I don't need to do runtime shared deps since in webpack MF entry was first 👌

patches/@module-federation__sdk@0.3.2.patch
jbroma: Wanted to showcase what was getting in the way. I would love to remove our custom MF plugin in the future and rely just on MF core 🤞

I can't think of a good way to differentiate RN from other envs though 🤔 one idea I had was to enhance the createScript hook, so we could patch the logic entirely. But I would rather keep the logic in MF altogether like you did with node

jbroma: figured most things out, PR here: https://github.com/module-federation/core/pull/2812

jbroma:
in v4 users will need to configure `getPublicPath` in plugin config. Until now, the publicPath was set to `noop://` in prod, and it would be a breaking change to modify this now.

In v5,  Re.Pack will use `publicPath` and this will further reduce the config needed.

apps/tester-federation-v2/rspack.config.host-app.mjs
```diff
@@ -0,0 +1,167 @@
...
+        },
+        dts: false,
+        runtimePlugins: [
+          resolve('@callstack/repack/federation-runtime-plugin'),
+          path.resolve(dirname, 'src', 'utils', 'runtime-debug.ts'),
+        ],
+        shared: {
+          react: {
+            singleton: true,
+            eager: true,
```
ScriptedAlchemy, line 131: Why do they have to be eager?

jbroma: `react-native` and `react` need to be present for initialisation of the environment done in `react-native/Libraries/Core/InitializeCore.js` - Hermes engine doesn't have event loop on its own -> no microtasks -> no timers like `setImmediate` or `setTimeout`.

ScriptedAlchemy: So I'm guessing there's no promise either? Like wouldn't it help if the entrypoint itself returned a promise so that entry setup itself is asynchronous? Like my hoisted runtime experiment does.

ScriptedAlchemy: I wonder if lynx supports event loops. 🤔 
This is something I'll probably want to bake into our vm if possible

jbroma: > So I'm guessing there's no promise either? Like wouldn't it help if the entrypoint itself returned a promise so that entry setup itself is asynchronous? Like my hoisted runtime experiment does.

yeah that would be ideal, but that would need more tinkering. I think I tried with hoisted runtime once but didn't pursue it much further. The biggest limiting factor is maintenance of custom initialisation logic. 

On the other side, @robik once had an idea of creating an extra layer between RN Core and Hermes called RN Runtime which could be a place for stuff like this. Not sure whats the status of that, maybe he can share something here as well :)

thymikee (approved):
LGTM :shipit:

apps/tester-federation-v2/src/utils/runtime-debug.ts
```diff
@@ -0,0 +1,34 @@
...
+  //     console.log('loadShare:', args);
+  //   },
+  //   async beforeLoadShare(args) {
+  //     console.log('beforeloadShare:', args);
+  //     return args;
+  //   },
+  // createScript(args) {
+  //   console.log('createScript:', args);
+  //   return args;
+  // },
```
thymikee, line 31: dead code?

jbroma: yeah, nice find 👍

apps/tester-federation-v2/rspack.config.host-app.mjs
robik: Yeah, I think this would solve this issue. The RFC is in pre-published state for now, should be up for discussions soon :)

PR #679, opened 2024-08-02
fix: infinite loop when traversing the chunk graph in `OutputPlugin`

### LINKED ISSUE #672: RangeError: Maximum call stack size exceeded
## Description

Hi, I'm encountering an issue with the OutputPlugin in the @callstack/repack library. During the build process, I'm receiving the following error:

`RangeError: Maximum call stack size exceeded`

## Reproducible Demo

setup like a docs and use command below to build bundle
```
react-native webpack-bundle --platform ios --entry-file index.js --dev false
```

**Actual Behavior:**
The build process fails with the `RangeError: Maximum call stack size exceeded error.`

**Analysis:**
I believe the issue may be related to the getAllInitialChunks function in the OutputPlugin.js file, which appears to call itself recursively without a proper termination condition, leading to a stack overflow. Here's the relevant part of the OutputPlugin.js

```
const getAllInitialChunks = (chunk, chunks) => {
  if (!chunk.parents?.length) return [chunk];
  return chunk.parents.flatMap(parent => {
    return getAllInitialChunks(chunks.get(parent), chunks);
  });
};
```
I modified the function to include a termination condition using a Set to track visited chunks, which seems to mitigate the issue:

```
const getAllInitialChunks = (chunk, chunks, visited = new Set()) => {
  if (visited.has(chunk)) {
    return [];
  }
  visited.add(chunk);

  if (!chunk.parents?.length) {
    return [chunk];
  }

  return chunk.parents.flatMap(parent => {
    return getAllInitialChunks(chunks.get(parent), chunks, visited);
  });
};

```
## Environment
@callstack/repack version: [4.1.1]
React Native version: 0.74.3
Node version: v20
OS: MacOS

Thank you for your assistance!

jbroma:
### Summary

Closes #672

Thank you @linhh-phv for suggesting the fix!

### Test plan

- [x] - tested on `tester-federation` where the bug occured when bundling for prod

### DIFF
--- a/packages/repack/src/webpack/plugins/OutputPlugin.ts
+++ b/packages/repack/src/webpack/plugins/OutputPlugin.ts
@@ -4,6 +4,7 @@ import { Rule, WebpackPlugin } from '../../types';
 import { AssetsCopyProcessor } from './utils/AssetsCopyProcessor';
 import { AuxiliaryAssetsCopyProcessor } from './utils/AuxiliaryAssetsCopyProcessor';
 
+type ChunkId = Exclude<webpack.StatsChunk['id'], undefined>;
 /**
  * Matching options to check if given {@link DestinationConfig} should be used.
  */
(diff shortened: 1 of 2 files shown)


packages/repack/src/webpack/plugins/OutputPlugin.ts
```diff
@@ -227,19 +227,22 @@ export class OutputPlugin implements WebpackPlugin {
 
     const getAllInitialChunks = (
       chunk: webpack.StatsChunk,
-      chunks: Map<string | number, webpack.StatsChunk>
+      chunks: Map<string | number, webpack.StatsChunk>,
+      visited = new Set<string | number>()
```
okwasniewski, line 231: nit: How about defining key as a separate type? 

```ts
type Key = string | number
```

okwasniewski (approved):

jbroma:
Thanks for the review @okwasniewski 🙇

hosseinmd:
I can confirm This change is working, please release a new patch version.

PR #680, opened 2024-08-03
fix: bundle caching after invalidating scripts

### LINKED ISSUE #553: Bug in `invalidateScripts` leading to app crash
## Environment

React Native android (but I believe the bug is not specific to RN android)

## Description

Calling `invalidateScripts` for a chunk is causing a crash when next attempting to use that chunk. On debugging it, the chunk file is getting deleted, but the entry in sqlite db is not getting cleared. It looks like there is a [bug in `invalidateScripts` method](https://github.com/callstack/repack/blob/main/packages/repack/src/modules/ScriptManager/ScriptManager.ts#L414 ). `resolveScripts` method appends caller to the `scriptId` to get the cacheKey (```const cacheKey = `${scriptId}_${caller ?? 'unknown'}`;```), while `invalidateScripts` directly uses `scriptId` as the cacheKey. 

## Reproducible Demo

1. Have a remote code split chunk.
2. Call `invalidateScripts` with that chunk's scriptId
3. Try using that chunk in UI, it leads to crash. `com.facebook.react.common.JavascriptException: ChunkLoadError: Loading chunk <scriptId> failed`

### LINKED ISSUE #660: Stuck while loading bundles on second app launch with cache flag enabled
## Environment

Localhost dev

## Description

Hello I've tried to enable caching in my project following the [caching and versioning](https://re-pack.dev/docs/code-splitting/caching-versioning) guide. The app runs with all the remotes loaded and working the first time I launch it. However, if I close the app and launch it again, it gets stuck at the launch screen.

I initially thought it was an issue with the project code, but the same problem occurs when I try it with the sample project [super-app-showcase](https://github.com/callstack/super-app-showcase).

image

I've tried to debug and it seems like it stuck at `Federated.importModule()`.

I didn't get this issue if I set the cache flag to false, but then the app downloads the remote bundles every time I need to access them, which is not optimal.

## Reproducible Demo

1. Clone and setup [super-app-showcase](https://github.com/callstack/super-app-showcase)
2. In
	`packages/booking/index.js`
	`packages/dashboard/index.js`
	`packages/host/index.js`
	`packages/shell/index.js`
	`packages/shopping/index.js`

	Add
	```
	import AsyncStorage from '@react-native-async-storage/async-storage';

	...

	ScriptManager.shared.setStorage(AsyncStorage);

	...
	```
	and set `cache: true`

jbroma:
### Summary

Closes #553

- fixed `invalidateScripts` behaviour when trying to invalidate all scripts (when passing undefined or `[]` as argument)
- fixed hang in native ScriptManager on iOS when trying to execute file that doesn't exist
- aligned Android errors to match iOS when executing the same flow

### Test plan

- [x] - local test on `tester-federation`
- [x] - verify canary fixes linked issue

### DIFF
--- a/.changeset/blue-dolls-know.md
+++ b/.changeset/blue-dolls-know.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+fix bundle caching after invalidating scripts
--- a/apps/tester-federation/app.json
+++ b/apps/tester-federation/app.json
@@ -23,5 +23,9 @@
   },
   "ios": {
     "bundleIdentifier": "com.tester.federation"
+  },
+  "resources": {
+    "android": [],
+    "ios": []
   }
 }
--- a/apps/tester-federation/index.js
+++ b/apps/tester-federation/index.js
@@ -1,8 +1,11 @@
 import { AppRegistry, Platform } from 'react-native';
 import { ScriptManager, Federated } from '@callstack/repack/client';
+import AsyncStorage from '@react-native-async-storage/async-storage';
 
 import App from './src/host/App';
 import { components } from './app.json';
+
+ScriptManager.shared.setStorage(AsyncStorage);
 
 const resolveURL = Federated.createURLResolver({
   containers: { MiniApp: 'http://localhost:8082/[name][ext]' },
@@ -12,6 +15,7 @@ ScriptManager.shared.addResolver(async (scriptId, caller) => {
   return {
     url: resolveURL(scriptId, caller),
     query: { platform: Platform.OS },
+    cache: process.env.MF_CACHE,
   };
 });
 
--- a/apps/tester-federation/ios/Podfile.lock
+++ b/apps/tester-federation/ios/Podfile.lock
@@ -1280,6 +1280,27 @@ PODS:
   - ReactTestApp-DevSupport (3.8.9):
     - React-Core
     - React-jsi
+  - RNCAsyncStorage (1.23.1):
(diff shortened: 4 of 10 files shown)


jbroma:
Canary for testing: https://www.npmjs.com/package/@callstack/repack/v/4.2.0-canary-20240803153933

hafiz-azman-cd:
Hello @jbroma I was testing the canary lib with [super-app-showcase](https://github.com/callstack/super-app-showcase). I manage to get it running and enabled the caching based on [here](https://re-pack.dev/docs/code-splitting/caching-versioning). I wasn't sure what to expect with the cache enabled. I still see the app requesting the bundle files from the dev server instead of loading from AsyncStorage when the app is killed and relaunched. I debugged the AsyncStorage and confirmed that the bundle files are present in their key-value pairs when the cache is enabled.

What I did in super-app-showcase:
1. `pnpm remove -r @callstack/repack --save-dev`
2. update `devDependencies.json` > @callstack/repack > version change to `@callstack/repack@4.2.0-canary-20240803153933`
3. `pnpm add -r @callstack/repack@4.2.0-canary-20240803153933 -D`
4. `pnpm pods`
5. Update booking, dashboard, shell and shopping to enable cache as in [here](https://re-pack.dev/docs/code-splitting/caching-versioning)
5. `pnpm start` + `pnpm run:host:ios`

jbroma:
Doesn't fix #660 but fixes #553, merging as is

PR #683, opened 2024-08-08
fix: bundle naming collisions when using device filesystem cache

### LINKED ISSUE #660: Stuck while loading bundles on second app launch with cache flag enabled

jbroma:
### Summary

Closes #660 

To prevent naming collisions on the device filesystem we're now using `uniqueId` as a filename.

`Script.locator.uniqueId` is generated by prepending `caller` to the `scriptId`.
If the `caller` is unknown, then `uniqueId` is equal to the `scriptId`.

### Notes

This PR changes the cache key from `v3` to `v4` so all previous cache won't be accessible with the next release that includes this PR

### Test plan

- [x] - local test on federation tester
- [x] - added unit tests to ScriptManager suite
- [x] - verify the change works in super app showcase scenario
- [x] - (optionally) verify canary release fixes the issue

### DIFF
--- a/.changeset/brown-needles-pull.md
+++ b/.changeset/brown-needles-pull.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+Fix bundle naming collisions when using device filesystem cache
--- a/packages/repack/android/src/main/java/com/callstack/repack/RemoteScriptLoader.kt
+++ b/packages/repack/android/src/main/java/com/callstack/repack/RemoteScriptLoader.kt
@@ -17,8 +17,8 @@ class RemoteScriptLoader(reactContext: ReactContext) : NativeScriptLoader(reactC
     private val scriptsDirName = "scripts"
     private val client = OkHttpClient()
 
-    private fun getScriptFilePath(id: String): String {
-        return "${scriptsDirName}/$id.script.bundle"
+    private fun getScriptFilePath(scriptUniqueId: String): String {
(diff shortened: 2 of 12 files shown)


jbroma:
Canary release for testing: https://www.npmjs.com/package/@callstack/repack/v/4.2.0-canary-20240808200412

Issue #665, opened 2024-07-09
`sourceURL` in raw React Native stack frames is inconsitent


jbroma:
## Description

Adding  bridgeless support introduced few regressions in terms of `sourceURL` (passed to `evaluateJavascript` from `jsi/jsi.h`) - this argument is used to tell the JS engine where the evaluated code comes from.

1. Native module - iOS:
- `sourceUrl` passed to `evaluateJavascript` in `executeFromFilesystem` from `packages/repack/ios/ScriptManager.mm` is incorrect -> we should use `url` from the `ScriptConfig`

2. Native module - Android:
- `sourceUrl` passed to `evaluate` in `execute` from `packages/repack/android/src/main/java/com/callstack/repack/RemoteScriptLoader.kt` is incorrect -> we should use `url` from the `ScriptConfig`

3. Possible inconsitentencies on Android:
- `sourceUrl` passed to `evaluate` in `execute` from `packages/repack/android/src/main/java/com/callstack/repack/FileSystemScriptLoader.kt` is inconsistent with `iOS` - depending on `config.absolute` we pass different values -> to align with iOS we should most likely use `config.url` in both cases

At this point it's hard to tell what should be passed in both cases - passing `sourceUrl` starting with `http://` makes sense in development and the whole symbolication flow is centred around that , but does it make sense in production when evaluating remote bundles?

## Thoughts on the matter

- we should consider passing just the name of the bundle (enhanced with `uniqueName` if needed - like in a super app setup) to match the source-map nicely - (this could be altered with `config.absolute`)
- we need to ensure that when something is loaded from cache ends up with the same name - TBD whether this is the case
- we need to ensure that local chunks also follow the same convention - the location only matters for the resolver and not the source-maps
- there is a chance that solving this could enable sentry to work with multiple source-maps out of the box, since there is no official support for that but Sentry allows uploading multiple files.

## Reproducible Demo

- can be reproduced in TesterApp by printing out the filenames from stack frames in `process` method from `packages/dev-server/src/plugins/symbolicate/Symbolicator.ts`. Note: filesystem loading flow can only be observed in `release`

## Additional notes

`sourceURL` is used when creating an error stack frame in React Native. When you want to symbolicate a stack trace, you lookup the source through `sourceURL` and then from the source file you can obtain the name of the source map, from the comment at the very bottom (added through `SourceMapDevToolPlugin`):
```
...
[...]

PR #711, opened 2024-08-26
fix: `sourceUrl` passed to eval for local and remote chunks

jbroma:
### Summary

Closes https://github.com/callstack/repack/issues/665

### Test plan

n/a


jbroma:
needs more changes - mostly default name for output chunks so that everything stays in sync and can be properly parsed, postponed to v5

jbroma:
merging as is, will most likely require an alignment in the future

PR #734, opened 2024-09-05
feat: Add `retry` and `retryDelay` Properties to Enhance Download Method Reliability

hexboy:
### Summary

This PR introduces two new properties, `retry` and `retryDelay`, to the `ScriptConfig` within the download method. These properties aim to improve the download process by handling network failures more robustly. 

- **`retry`**: Specifies the number of attempts to retry the download in case of failure.
- **`retryDelay`**: Defines the delay (in milliseconds) between each retry attempt.

This feature aims to minimize the impact of transient network failures and improve the reliability of downloads in unstable network conditions.

### DIFF
--- a/packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
@@ -18,8 +18,6 @@ export interface NormalizedScriptLocator {
   url: string;
   fetch: boolean;
   timeout: number;
-  retry: number | undefined;
-  retryDelay: number | undefined;
   absolute: boolean;
   query: string | undefined;
   headers: { [key: string]: string } | undefined;
--- a/packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
@@ -18,6 +18,8 @@ export interface NormalizedScriptLocator {
   url: string;
   fetch: boolean;
   timeout: number;
+  retry: number | undefined;
+  retryDelay: number | undefined;
   absolute: boolean;
   query: string | undefined;
   headers: { [key: string]: string } | undefined;
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -18,6 +18,14 @@ const CACHE_ENV = __DEV__ ? 'debug' : 'release';
 
 const CACHE_KEY = [CACHE_NAME, CACHE_VERSION, CACHE_ENV].join('.');
 
+const LOADING_ERROR_CODES = [
+  // android
+  'NetworkFailure',
+  'RequestFailure',
+  // ios
+  'ScriptDownloadFailure',
+];
+
 /* Options for resolver when adding it to a `ScriptManager`. */
 export interface ResolverOptions {
   /**
@@ -355,7 +363,8 @@ export class ScriptManager extends EventEmitter {
         return; // Successfully loaded the script, exit the loop
       } catch (error) {
         attempts--;
-        if (attempts > 0) {
+        const { code } = error as Error & { code: string };
+        if (attempts > 0 && LOADING_ERROR_CODES.includes(code)) {
           if (retryDelay > 0) {
             await new Promise((resolve) => setTimeout(resolve, retryDelay));
           }
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -316,7 +316,7 @@ export class ScriptManager extends EventEmitter {
 
     try {
       this.emit('loading', script.toObject());
-      await this.nativeScriptManager.loadScript(scriptId, script.locator);
+      await this.loadScriptWithRetry(scriptId, script.locator);
       this.emit('loaded', script.toObject());
     } catch (error) {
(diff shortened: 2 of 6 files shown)


packages/repack/android/src/main/java/com/callstack/repack/RetryInterceptor.kt
```diff
@@ -0,0 +1,39 @@
...
+                }
+            } catch (e: IOException) {
+                lastException = e
+            }
+
+            retryCount++
+
+            if (retryCount <= maxRetries) {
+                // Delay before retrying
+                Thread.sleep(retryDelayMillis.toLong())
```
jbroma, line 32: putting a thread to sleep might block some other work, it would be best to schedule the the retry rather than sleep

jbroma (changes requested):
hi @hexboy, thanks for another contribution! 

I've reviewed the PR and left some comments there. On a side-note, I think that we could do a lot of this on JS-side only - all we need is to check if the error from native `loadScript` is related to fetching data - WDYT?

jbroma:
hey @hexboy, do you want to work on this PR or should we take this over? It's definitely something that we want to have in the next major (5.0) release. Either way your contribution is very much appreciated!

hexboy:
Hey @jbroma, I'll start working on rewriting the entire PR in JavaScript. I estimate it will take one or two days to complete.

Since I don't see any existing native unit tests in the project, and there may be a need to add a new native module or package, I'm not planning to write native tests for this.

jbroma:
> Hey @jbroma, I'll start working on rewriting the entire PR in JavaScript. I estimate it will take one or two days to complete.

glad to hear that!

> Since I don't see any existing native unit tests in the project, and there may be a need to add a new native module or package, I'm not planning to write native tests for this.

yeah, let's not worry about native tests for now 👍

hexboy:
Hey @jbroma, I just rewrote the entire PR in JavaScript. Please check it again.

packages/repack/src/modules/ScriptManager/ScriptManager.ts
```diff
@@ -329,6 +329,43 @@ export class ScriptManager extends EventEmitter {
...
+
+    while (attempts > 0) {
+      try {
+        await this.nativeScriptManager.loadScript(scriptId, locator);
+        return; // Successfully loaded the script, exit the loop
+      } catch (error) {
+        attempts--;
+        if (attempts > 0) {
+          if (retryDelay > 0) {
+            await waitMs(retryDelay);
```
hosseinmd, line 360: Can it be done just for network calls?
How we could detect it is a network issue?

jbroma: this is a valid concern, we should only identify the Network related errors here and handle them, otherwise we should fail as fast possible 👍

jbroma, line 360: no need to make this a separate function here, lets just inline this

packages/repack/src/modules/ScriptManager/ScriptManager.ts
```diff
@@ -329,6 +329,43 @@ export class ScriptManager extends EventEmitter {
...
+   * If the initial attempt fails, it retries the specified number of times
+   * with an optional delay between retries.
+   *
+   * @param {string} scriptId - The ID of the script to load.
+   * @param {NormalizedScriptLocator} locator - An NormalizedScriptLocator containing retry configuration.
+   * @param {number} [locator.retry=0] - The number of retry attempts.
+   * @param {number} [locator.retryDelay=0] - The delay in milliseconds between retries.
+   * @throws {Error} Throws an error if all retry attempts fail.
+   */
+  async loadScriptWithRetry(
```
jbroma, line 345: lets make it private/protected, this is an internal method of `loadScript` and we don't need to expose it for the end user.

jbroma (changes requested):
thanks for rewriting this in JS! Looks good overall, few nits there and there, take a look 👇

packages/repack/src/modules/ScriptManager/ScriptManager.ts
hexboy: I'll work on that.

hexboy: it's Done.

packages/repack/src/modules/ScriptManager/NativeScriptManager.ts
```diff
@@ -18,6 +18,8 @@ export interface NormalizedScriptLocator {
   url: string;
   fetch: boolean;
   timeout: number;
+  retry: number | undefined;
+  retryDelay: number | undefined;
```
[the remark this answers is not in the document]
hexboy: Done.

packages/repack/src/modules/ScriptManager/ScriptManager.ts
hexboy: Done.

packages/repack/src/modules/ScriptManager/ScriptManager.ts
hexboy: Done.

jbroma (commented):
All looks good now 🎉 Let's fix the failing tests by not testing the protected method in the tests and instead, test `loadScript` directly there since it's exposed as a public interface 👍

jbroma (approved):
LGTM 🎉

jbroma:
Thank you for working on this @hexboy, much appreciated!

PR #744, opened 2024-09-11
docs: align wording with Rspack support

thymikee:
### Summary

Aligning the wording for Rspack support. Also changed the React Native CLI to React Native Community CLI. There's still v5.x docs missing, but let's address that separately.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -12,7 +12,7 @@ for an additional context on configuration, parameters and types.
 
 There are 2 ways to look at the content of the Re.Pack:
 
-- by command that are exposed to React Native CLI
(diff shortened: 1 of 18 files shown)


website/src/public/diffs/repack_v1-v2.diff
jbroma, line 1: this an archived file, let's avoid changing it

website/src/public/diffs/repack_v2-v3.diff
jbroma, line 1: this an archived file, let's avoid changing it

jbroma (changes requested):
2 issues then :shipit:

jbroma (approved):
all good 🎉 will merge after v5 branch

PR #745, opened 2024-09-12
fix: throw an error when bundle download was unsuccessful on iOS

ra1nj:
### Summary

On iOS ,if download bundle was not successful, throw an error.


jbroma (approved):
LGTM, thanks for another contribution @ra1nj 🎉

PR #753, opened 2024-09-19
feat: support React Native 0.76

artus9033:
### Summary

- [x] - align CMakeLists to account for merged RN SO libs
- [x] - updated both tester apps to `0.76-rc.6`
- [x] - updated codegen rules to use hermes parser
- [x] - updated `flow-remove-types` to include necessary fixes for RN 76
- [x] - add missing fusebox related method to `DevClient`  

### Test plan

- [x] - tester app works with old/new arch
- [x] - tester federation works with old/new arch


packages/repack/src/modules/DevServerClient.ts
jbroma, line 1: let's move this to a separate PR for support of RN 0.76

packages/repack/android/gradle.properties
jbroma, line 1: let's move this to a separate PR for support of RN 0.76

packages/repack/android/build.gradle
jbroma, line 1: let's move this to a separate PR for support of RN 0.76

packages/repack/android/CMakeLists.txt
jbroma, line 1: let's move this to a separate PR for support of RN 0.76

apps/tester-federation/package.json
```diff
@@ -20,16 +20,19 @@
...
-    "@react-native/babel-preset": "0.74.83",
-    "@react-native/eslint-config": "0.74.83",
-    "@react-native/typescript-config": "0.74.83",
+    "@babel/core": "^7.25.2",
+    "@react-native/babel-preset": "0.76.0-rc.1",
+    "@react-native/eslint-config": "0.76.0-rc.1",
+    "@react-native/typescript-config": "0.76.0-rc.1",
+    "@react-native-community/cli": "15.0.0-alpha.2",
+    "@react-native-community/cli-platform-android": "15.0.0-alpha.2",
+    "@react-native-community/cli-platform-ios": "15.0.0-alpha.2",
```
jbroma, line 35: do we need to add all these 3 packages explicitly? 

cc @szymonrybczak

apps/tester-app/package.json
```diff
@@ -16,16 +16,21 @@
...
-    "react-native": "0.74.3",
-    "react-native-svg": "15.3.0"
+    "react": "18.3.1",
+    "react-native": "0.76.0-rc.1",
+    "react-native-svg": "15.8.0-rc.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.23.9",
+    "@babel/core": "^7.25.2",
+    "@babel/plugin-syntax-typescript": "^7.25.4",
```
jbroma, line 25: is this needed?

apps/tester-app/package.json
```diff
@@ -16,16 +16,21 @@
...
     "@callstack/repack": "workspace:*",
-    "@react-native/babel-preset": "0.74.83",
-    "@react-native/eslint-config": "0.74.83",
-    "@react-native/typescript-config": "0.74.83",
+    "@react-native-community/cli": "15.0.0-alpha.2",
+    "@react-native-community/cli-platform-android": "15.0.0-alpha.2",
+    "@react-native-community/cli-platform-ios": "15.0.0-alpha.2",
+    "@react-native/babel-plugin-codegen": "^0.75.3",
```
jbroma, line 30: isn't this included in the babel-preset?

jbroma (changes requested):
@artus9033 thanks for the work on this! fantastic progress 🎉 

~~we need to put 0.76 changes in Re.Pack in a separate PR~~

Let's make sure all of this is backwards compatible with 0.75 and 0.74 - did you manage to verify this already?

Other than that it looks solid 🎉

jbroma (changes requested):
@artus9033 thanks for the work on this! fantastic progress 🎉 

we need to put 0.76 changes in Re.Pack in a separate PR and make sure all of this is backwards compatible with 0.75 and 0.74 - did you manage to verify this already?

Other than that it looks solid 🎉

apps/tester-federation/package.json
artus9033: The RN upgrade helper specifies this explicitly, but now I see that RNTA already has them: https://github.com/microsoft/react-native-test-app/blob/a89b47711dd2ddd5a92d853267204af644a76d66/package.json#L171 so most likely we can skip this

apps/tester-app/package.json
artus9033: Yes, apparently from now on, it is 
image

apps/tester-app/package.json
artus9033: No, same thing as above - without it, it crashes. I agree that it should be unneeded since the RN babel preset [already has it](https://github.com/facebook/react-native/blob/v0.76.0-rc.1/packages/react-native-babel-preset/package.json), however still it is not found without this explicit dependency:
image

apps/tester-federation/package.json
jbroma: these are just resolutions, not dependencies though

apps/tester-app/package.json
jbroma: is this comming from upgrade helper? or just because rnsvg doesnt work without it?

apps/tester-federation/package.json
artus9033: Removed in 2cdf84773fbe164e37ec0a53125d8b09d0b85b24

apps/tester-app/__tests__/bundle.test.ts
```diff
@@ -15,7 +15,10 @@ const RELATIVE_REACT_NATIVE_PATH = path.relative(
 const REACT_NATIVE_ANDROID_ASSET_PATH = RELATIVE_REACT_NATIVE_PATH.replaceAll(
   path.sep,
   '_'
-).replaceAll(/[-.@+]/g, '');
+)
+  .replaceAll(/[-.@+]/g, '')
+  // when a module is patched with pnpm, its node_modules' path changes
+  .replaceAll(/patch_hash=/g, 'patch_hash');
```
jbroma, line 21: what about adding just '=' to the regex above?

jbroma:
flow-types still being removed incorrectly, this PR to `flow-remove-types` should fix it:
https://github.com/facebook/flow/pull/9213

jbroma:
- [ ] - one more flow issue that needs to be tackled: https://github.com/facebook/flow/pull/9216
- RNScreens on Android still seems to be failing build, works on iOS though

jbroma:
patched RNScreens using changes from this PR: https://github.com/software-mansion/react-native-screens/pull/2415, after that it builds correctly for Android 👍

all thats left is the flow issue and we should be g2g on this

jbroma (approved):

jbroma:
Thanks for your work on this @artus9033 🎉

jnbt:
@jbroma I might be mistaken, these changes have not been back ported to 4.x, or?

jbroma:
@jnbt you are correct, 4.4.0 was just released to fix that, sorry to keep you waiting so long

PR #755, opened 2024-09-20
fix: disable package imports by default

jbroma:
### Summary

metro doesn't support import fields at all, aligning for compat.

if a user wants to allow for importsFields, they can do so through `resolve.importsFields` and restoring the default value of `['imports']`

### Test plan

n/a

### DIFF
--- a/.changeset/lucky-paws-shop.md
+++ b/.changeset/lucky-paws-shop.md
@@ -1,0 +1,5 @@
+---
+'@callstack/repack': patch
+---
+
+Disable package imports by default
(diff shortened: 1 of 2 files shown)


thymikee:
Any chance we could have a docs page covering all the custom webpack config we do by default and a reason for doing so (either RN specific thing or Metro quirk)? Sounds like it could be helpful for mobile + web projects reusing as much of their configs as possible

jbroma:
> Any chance we could have a docs page covering all the custom webpack config we do by default and a reason for doing so (either RN specific thing or Metro quirk)? Sounds like it could be helpful for mobile + web projects reusing as much of their configs as possible

Yeah, should be covered in module resolution docs 👍

Issue #749, opened 2024-09-16
Error: Script file does not exist, but the file exists.


hosseinmd:
### Describe the bug

I have a project with mode federation and remote bundles

I am getting this error after bundle downloaded successfully (I checked network)
`Error: Script file does not exist: /data/user/0/com.a.dev/files/scripts/user.script.bundle`
I check the file path and I realized it exists.

This error happening randomly. Some time all work, some time happenning for other containers.

### System Info

```shell
Android 13
react-native 75
New Architecture
```

### Re.Pack Version

4.3.3

### Reproduction

n/a

### Steps to reproduce

Create a react-native new architecture project with few containers. Make them remote federation.

artus9033:
Hi @hosseinmd, could you post your repack configuration or even better, provide an MRE to reproduce this problem? I checked locally with RN 0.75.3, newest Re.Pack and it works on my side.

hosseinmd:
Hi, it is hard to reproduce. But I will find a way.

I found the problem. It is because a when module script is starting to load another request was call but the `config.fetch==false` so second request want to read file from file system before than first request write code in that file.

https://github.com/callstack/repack/blob/90faeeb7d6be9ddf5aa74c9552df01ec58d5372c/packages/repack/android/src/main/java/com/callstack/repack/ScriptManagerModule.kt#L33

It needs to wait until an ongoing process to `loadScript` finish, or prevent to load a script twice.

hosseinmd:
In this PR I fixed the issue.
https://github.com/callstack/repack/pull/756

PR #756, opened 2024-09-21
fix: prevent to load script again

hosseinmd:
https://github.com/callstack/repack/issues/749
### Summary
In this PR, I prevented `loadScript` to load scripts is loaded before

### Test plan
This is happening in remote federation with huge size of script and too many modules


jbroma (changes requested):
Hey @hosseinmd, sorry for the late reply but I was unavailable most of the week.

The concept introduced in this PR looks great, we definitely want this in the next major (5.0) release!

Can you please add some tests for this scenarios? We need to document this behaviour as well as we can

I've also noticed that we're using the same variable 
`sriptsPromises` for both prefetching and loading - while in theoery these should not overlap, I would go with creating a separate variables to avoid potential collision in rare edge cases. 

We also need to address housekeeping of the promises that have already been fulfilled - we should remove the entires from `scriptsPromises` that are not relevant anymore.

Last thing that comes to my mind: perhaps we could return a promise in `prefetchScript` and `loadScript` instead of executing loadProcess again after awaiting the previous promise?

hosseinmd:
> Can you please add some tests for this scenarios? We need to document this behaviour as well as we can

Yes, I will do it

hosseinmd:
> I've also noticed that we're using the same variable
> `sriptsPromises` for both prefetching and loading - while in theoery these should not overlap, I would go with creating a separate variables to avoid potential collision in rare edge cases.

`loadScript` and `prefetchScript` are using same `resolveScript`. If we separate those, the same bug https://github.com/callstack/repack/issues/749 possible to be happening.

hosseinmd:
> We also need to address housekeeping of the promises that have already been fulfilled - we should remove the entires from `scriptsPromises` that are not relevant anymore.

Ok, I will do it.

hosseinmd:
> > I've also noticed that we're using the same variable
> > `sriptsPromises` for both prefetching and loading - while in theoery these should not overlap, I would go with creating a separate variables to avoid potential collision in rare edge cases.
> 
> `loadScript` and `prefetchScript` are using same `resolveScript`. If we separate those, the same bug #749 possible to be happening.

i added a test for that. https://github.com/callstack/repack/pull/756/commits/969d6d8819498ffe95ba797c414ffabcb4810358

hosseinmd:
> > I've also noticed that we're using the same variable
> > `sriptsPromises` for both prefetching and loading - while in theoery these should not overlap, I would go with creating a separate variables to avoid potential collision in rare edge cases.
> 
> `loadScript` and `prefetchScript` are using same `resolveScript`. If we separate those, the same bug #749 possible to be happening.

In this https://github.com/callstack/repack/pull/756/commits/ee3a58c5f3f2d9d2a9ae44c47d3d0fe4b72fbdb6 commit loadScript will excute script if promise is for prefetch

jbroma:
@hosseinmd can you merge newest main and resolve the conflicts? Then I will proceed with another review, thanks!

hosseinmd:
Done

jbroma:
> Done

Nice job! I'll try to review it later today

packages/repack/src/modules/ScriptManager/__tests__/Federated.test.ts
jbroma, line 1: can we move this to a a file with ScriptManager tests? almost all of the setup feels repeated here.

packages/repack/src/modules/ScriptManager/__tests__/Federated.test.ts
```diff
@@ -0,0 +1,145 @@
...
+beforeEach(() => {
+  ScriptManager.init();
+});
+
+afterEach(() => {
+  globalThis.__webpack_require__.repack.shared.scriptManager = undefined;
+});
+
+describe('Federated', () => {
+  it('should wait loadScript with same scriptId to finished ', async () => {
```
jbroma, line 60:
```suggestion
  it('should await loadScript with same scriptId to finish ', async () => {
```

packages/repack/src/modules/ScriptManager/ScriptManager.ts
```diff
@@ -92,6 +92,10 @@ export class ScriptManager extends EventEmitter {
   }
 
   protected cache: Cache = {};
+  protected scriptsPromises: Record<
+    string,
+    (Promise<void> & { isPrefetch?: true }) | undefined
+  > = {};
```
jbroma, line 98: lets move this type to a separate line

jbroma (changes requested):
lets clean this up a little bit, other than these few things, nice job!

hosseinmd:
Done

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
```diff
@@ -2,7 +2,13 @@ import { Script } from '../Script';
...
 
 jest.mock('../NativeScriptManager', () => ({
-  loadScript: jest.fn(),
+  loadScript: jest.fn((locator) =>
+    locator.fetch
+      ? new Promise<void>((resolve) => {
+          setTimeout(() => resolve(), 10);
+        })
+      : Promise.resolve()
+  ),
```
jbroma, line 11: can we only apply this mock implementation to the selected tests that we need? I think this can be achieved with `jest.spyOn`

jbroma (changes requested):
all looks good, just one last thing 🙏

hosseinmd:
Last commits rebased

jbroma (approved):
LGMT 🎉

jbroma:
Thank you for your work on this @hosseinmd, this was no easy issue to tackle! 👏 🚀

PR #757, opened 2024-09-26
feat: support experimental RN debugger interaction

artus9033:
### Summary

This PR introduces support for the TUI 'j' interaction and implementations of this interaction for Webpack & Rspack.
It also refactors code in `setupInteractions.ts` to properly warn about possible (currently none) unsupported interactions in the underlying bundler wrapper implementation, as well as replaces a series of `if`-`else` expressions with a `switch` and replacing object arrow functions with object literal methods shorthand syntax.

#### Demo

Live interaction:
demo-repack-exp-debugger

Initial help message:

| With experimental debugger capability | Without experimental debugger capability |
| --------- | --------- |
| Screenshot 2024-09-26 at 18 17 03 | Screenshot 2024-09-26 at 18 23 26 |

### Test plan

- [x] Run without `--experimental-debugger` & pressing 'j' should result in warning: `DevTools require the '--experimental-debugger' flag to be passed to the bundler process`
- [x] Run with `--experimental-debugger` & pressing 'j' should result in the DevTools being opened, both for Webpack & Rspack


thymikee:
demo pls!

artus9033:
> demo pls!

I've just updated the root comment to include a demo GIF.

thymikee:
Do we have a way to display the shortcuts that you can press?

artus9033:
For sure we can, I will have this implemented as well.

artus9033:
Done, also added screenshots ✅

packages/repack/src/commands/common/setupInteractions.ts
```diff
@@ -23,16 +57,83 @@ export function setupInteractions(
...
+        : undefined,
+      postPerformMessage: 'Opening DevTools',
+      helpName: 'Open DevTools',
+      actionUnsupportedExplanation: hasExperimentalDebuggerSupport
+        ? undefined
+        : `${EXPERIMENTAL_DEBUGGER_FLAG} was not passed`,
+    },
+  };
+
+  console.log(blueBright('You can use the following keystrokes:'));
```
jbroma, line 126: use logger instead of console

packages/repack/src/commands/common/setupInteractions.ts
```diff
@@ -23,16 +57,83 @@ export function setupInteractions(
...
+        interaction?.actionUnsupportedExplanation === undefined &&
+        interaction?.action !== undefined,
+      text = `${key}: ${interaction?.helpName}${isSupported ? '' : yellow(` (unsupported${interaction?.actionUnsupportedExplanation ? `, ${interaction.actionUnsupportedExplanation}` : 'by the current bundler'})`)}`;
+
+    console.log(isSupported ? text : italic(text));
+  }
+
+  console.log('');
+  console.log('Press ctrl+c or ctrl+z to quit the dev server');
+  console.log('');
```
jbroma, line 138: use logger instead of console

packages/repack/src/commands/common/setupInteractions.ts
```diff
@@ -23,16 +57,83 @@ export function setupInteractions(
...
+  };
+
+  console.log(blueBright('You can use the following keystrokes:'));
+  for (const [key, interaction] of Object.entries(plainInteractions)) {
+    const isSupported =
+        interaction?.actionUnsupportedExplanation === undefined &&
+        interaction?.action !== undefined,
+      text = `${key}: ${interaction?.helpName}${isSupported ? '' : yellow(` (unsupported${interaction?.actionUnsupportedExplanation ? `, ${interaction.actionUnsupportedExplanation}` : 'by the current bundler'})`)}`;
+
+    console.log(isSupported ? text : italic(text));
```
jbroma, line 133: use logger instead of console

packages/repack/src/commands/rspack/start.ts
```diff
@@ -106,12 +110,17 @@ export async function start(
...
               ctx.broadcastToMessageClients({ method: 'reload' });
             },
-            onOpenDevMenu: () => {
+            onOpenDevMenu() {
               ctx.broadcastToMessageClients({ method: 'devMenu' });
             },
+            onOpenDevTools() {
+              void fetch(`${serverURL}/open-debugger`, {
+                method: 'POST',
+              });
```
jbroma, line 122: I think we should call this through `ctx` which comes from dev server, no need to call this over the network when we have access to dev server in the same process

packages/repack/src/commands/common/setupInteractions.ts
```diff
@@ -1,10 +1,44 @@
 import readline from 'node:readline';
+import { yellow, blueBright, italic } from 'colorette';
```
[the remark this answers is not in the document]
artus9033: Makes sense to me, I will rewrite this

packages/repack/src/commands/common/setupInteractions.ts
```diff
@@ -23,16 +57,83 @@ export function setupInteractions(
...
+        }
+      }
     }
   });
+
+  // since now Re.pack officially supports RN >= 0.73, it is sure that RN
+  // has the capability of the new debugger
+  const hasExperimentalDebuggerSupport = process.argv.includes(
+    EXPERIMENTAL_DEBUGGER_FLAG
+  );
```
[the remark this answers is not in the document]
artus9033: Sure

packages/repack/src/commands/common/setupInteractions.ts
artus9033: As we spoke, the `ctx.logger` does not write anything at this stage.

jbroma: use `logger` but dont pass `ctx.logger` in `start.ts` 👍

jbroma (approved):
LGMT 🚀

jbroma:
this is how final output looks like:

image

PR #763, opened 2024-10-06
feat: add util function to generate module paths

thiagobrez:
### Summary

This PR introduces the `getModulePaths` util function, to generate module path regexes for both classic (npm, yarn) and exotic (pnpm) paths, to be used in the bundler config file `include` rules.

This function will be available for both internal and public use.

### Test plan

Tests to classic and exotic module paths are included, also taking in consideration nested packages and backslashes from Windows systems.


packages/repack/src/utils/getModulePaths.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        '@callstack/react-native-visionos',
+        '@react-native/core',
+        '@babel/core',
+        '@types/react',
+        '@expo/vector-icons',
+      ],
+    },
+  ];
+
+  testCases.forEach(({ category, examples }) => {
```
jbroma, line 31: we can use `jest.each` for this, take a look at `assetsLoader.test.ts`

packages/repack/src/utils/getModulePaths.test.ts
jbroma, line 1: lets move it to `__tests__` dir in the utils to follow the pattern in the repo

jbroma (changes requested):
thanks for the PR! The logic is sound 🎉, we just need to work on the API and simplify the tests a little bit, see the comments below

packages/repack/src/utils/getModulePaths.ts
```diff
@@ -0,0 +1,42 @@
...
+ * import * as Repack from '@callstack/repack';
+ *
+ * return {
+ *   ...,
+ *   module: {
+ *     rules: [
+ *       ...,
+ *       include: [
+ *         ...Repack.getModulePaths('react-native'),
+ *         ...Repack.getModulePaths('some-other-package'),
```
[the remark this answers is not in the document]
thiagobrez: That's totally right. Changed it to accept and array of deps

packages/repack/src/utils/__tests__/getModulePaths.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    '@expo/vector-icons',
+    'socket.io',
+  ];
+
+  const paths = getModulePaths(packages);
+
+  // Classic paths are even-indexed
+  const classicPaths = paths.filter((_, index) => index % 2 === 0);
+  // Exotic paths are odd-indexed
+  const exoticPaths = paths.filter((_, index) => index % 2 !== 0);
```
thiagobrez, line 22: @jbroma Let me know what you think. Now that the `getModulePaths` function accepts an array of deps (and returns a flattened array of regexes) I had to filter classic and exotic paths based on if they're odd or even-indexed.

packages/repack/src/utils/__tests__/getModulePaths.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  packages.forEach((packageName, index) => {
```
thiagobrez, line 24: I had migrated to `it.each` but had to rollback because it doesn't provide an `index` of the iterated element. We need it to determine if the current iterated element is classic or exotic, to be able to test it correctly

thiagobrez:
@jbroma Pushed the requested changes!

Let me know what you think about the new test structure. I separated it into logical tests (classic, exotic, backslashes) and think it's much better.

Although I had to do some massaging in the regex array returned from the function to be able to have some more scalable tests.

packages/repack/src/utils/__tests__/getModulePaths.test.ts
jbroma: hmmm, maybe lets just call `getModulePaths` with an array with 1 element (package) in each test and then we know that 0 will have classicPath and 1 will have exoticPath? If there is a new case in the future then we can just use 2 as next one

packages/repack/src/rules/reactNativeLoadingRules.ts
```diff
@@ -1,26 +1,19 @@
...
-    /node_modules([/\\])+@callstack[/\\]react-native-visionos[/\\]/,
-    // exotic paths (e.g. pnpm)
-    /node_modules(.*[/\\])+react-native@/,
-    /node_modules(.*[/\\])+@react-native\+/,
-    // exotic paths for OOT
-    /node_modules(.*[/\\])+react-native-macos@/,
-    /node_modules(.*[/\\])+react-native-windows@/,
-    /node_modules(.*[/\\])+react-native-tvos@/,
-    /node_modules(.*[/\\])+@callstack\+react-native-visionos@/,
+    ...getModulePaths([
```
jbroma, line 9: `include: getModulePaths([...` to skip the extra spread :)

jbroma (approved):
LGTM 🎉

jbroma:
Thank you for working on this @thiagobrez, very useful stuff 🔥

Issue #770, opened 2024-10-18
Cache value is not valid when script is failed


hosseinmd:
### Describe the bug

In some situation, for example network failed when loading a script, The failed script is cached.

### System Info

```shell
repack 4.3.3
react-native 75.4
```

### Re.Pack Version

4.3.3

### Reproduction

https://github.com/callstack/repack

### Steps to reproduce

Make your app fail to load remote script, then restart the app. Here the failed script doesn't download again, and it is trying to execute from cache, but there is no cache file.

jbroma:
@hosseinmd, I'm finishing the work on MF2 and should release V5 RC very soon 🙏 
I would put this issue on hold until we can verify that it's also happening on V5

hosseinmd:
Please review this https://github.com/callstack/repack/pull/771
I made a good test scenario which can show you the problem

jbroma:
Thank you for handling this @hosseinmd!

Closing since this was fixed by #771. 

Fix available in `5.0.0-rc.0` version of Re.Pack.

PR #771, opened 2024-10-18
fix: don't cache failed script

hosseinmd:
### Summary
https://github.com/callstack/repack/issues/770

The root cause was where cache is saving, script Resolver should not save the cache, it should be saved after script loaded successfully.

### Test plan
I added a test for failing scenario which expected to The failed script doesn't save on cache and also if scriptLoader called again, expected in second time fetch property should be true too.

### DIFF
--- a/packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
+++ b/packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
@@ -768,7 +768,7 @@ describe('ScriptManagerAPI', () => {
     await ScriptManager.shared.loadScript('miniApp');
 
     expect(loadingScriptIsFinished).toEqual(true);
-    spy.mockReset();
+
     spy.mockRestore();
   });
 
@@ -810,7 +810,7 @@ describe('ScriptManagerAPI', () => {
     expect(loadingScriptIsFinished).toEqual(true);
     await ScriptManager.shared.loadScript('miniApp2');
     expect(loadingScript2IsFinished).toEqual(true);
-    spy.mockReset();
+
     spy.mockRestore();
   });
 
@@ -836,7 +836,7 @@ describe('ScriptManagerAPI', () => {
     await ScriptManager.shared.loadScript('miniApp');
 
     expect(prefetchScriptIsFinished).toEqual(true);
-    spy.mockReset();
+
     spy.mockRestore();
   });
 
@@ -844,7 +844,7 @@ describe('ScriptManagerAPI', () => {
     jest.useFakeTimers({ advanceTimers: true });
     const spy = jest.spyOn(NativeScriptManager, 'loadScript');
 
-    spy.mockImplementation(
+    spy.mockRejectedValueOnce(
       (_scriptId: string, _scriptConfig: NormalizedScriptLocator) =>
         Promise.reject({ code: 'NetworkFailed' })
     );
@@ -866,14 +866,11 @@ describe('ScriptManagerAPI', () => {
     // //expected to cache be empty
     expect(Object.keys(cache.data).length).toBe(0);
 
-    await expect(async () =>
-      ScriptManager.shared.loadScript('miniApp')
-    ).rejects.toThrow();
+    await ScriptManager.shared.loadScript('miniApp');
 
     // expected to fetch again
-    // expect(spy.mock.lastCall?.[1].fetch).toBe(true);
+    expect(spy.mock.lastCall?.[1].fetch).toBe(true);
 
-    spy.mockReset();
     spy.mockRestore();
   });
 });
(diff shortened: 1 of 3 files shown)


packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
```diff
@@ -435,6 +437,7 @@ describe('ScriptManagerAPI', () => {
       'src_App_js',
       'main'
     );
+    await ScriptManager.shared.loadScript('src_App_js', 'main');
```
jbroma, line 440: why do we need to call `loadScript` here in test for `resolveScript` ?

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
```diff
@@ -590,7 +597,7 @@ describe('ScriptManagerAPI', () => {
     expect(NativeScriptManager.loadScript).toHaveBeenCalledTimes(1);
     expect(NativeScriptManager.loadScript).toHaveBeenCalledWith(scriptId, {
       absolute: false,
-      fetch: false,
+      fetch: true,
```
jbroma, line 600: why do we need to set `fetch` to true here?

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
```diff
@@ -832,6 +839,43 @@ describe('ScriptManagerAPI', () => {
...
   });
+
+  it('should refetch failed script', async () => {
+    jest.useFakeTimers({ advanceTimers: true });
+    const spy = jest.spyOn(NativeScriptManager, 'loadScript');
+
+    spy.mockImplementation(
+      (_scriptId: string, _scriptConfig: NormalizedScriptLocator) =>
+        Promise.reject({ code: 'NetworkFailed' })
+    );
```
jbroma, line 850: maybe we can get away with `.mockRejectedValueOnce` ?

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
```diff
@@ -832,6 +839,43 @@ describe('ScriptManagerAPI', () => {
...
+
+    // //expected to cache be empty
+    expect(Object.keys(cache.data).length).toBe(0);
+
+    await expect(async () =>
+      ScriptManager.shared.loadScript('miniApp')
+    ).rejects.toThrow();
+
+    // expected to fetch again
+    // expect(spy.mock.lastCall?.[1].fetch).toBe(true);
```
jbroma, line 874: I guess we can delete this if we are not using this anymore?

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
```diff
@@ -832,6 +839,43 @@ describe('ScriptManagerAPI', () => {
...
+
+    spy.mockReset();
+    spy.mockRestore();
```
jbroma, line 877: `mockRestore()` should be sufficient on it's own

https://jestjs.io/docs/mock-function-api#mockfnmockrestore

jbroma (commented):
hey @hosseinmd, sorry for the late reply and thanks again for yet another contribution 🎉. I was on sick leave during last week so wasn't able to get to this any sooner. 

Overall, the changes in code looks good, I just need some explanation for the tests, take a look 👇

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
hosseinmd: It is because we move `saveCache` from `resolveScriopt` to end of `loadScript`. So if we want to sure caching system working correctly, we should call `loadScript` to save cache

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
hosseinmd: This is the main problem This PR solves. If a LoadScript failed, it should be fetched again.
Currently, repack is caching even loadScript failed.

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
hosseinmd: i will try

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
hosseinmd: i will fix it

packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
jbroma: do we need to call `resolveScript` before here? I think we can just do `loadScript` now

hosseinmd: We need value returned by `resolveScript` to `script1` to check fetch

jbroma: ah right, makes sense!

jbroma:
Thank you for your contribution @hosseinmd 👏🎉

PR #786, opened 2024-11-08
[BREAKING] feat: enable React Native DevTools by default

jbroma:
### Summary

- [x] - promote React Native DevTools as the stable and supported solution
- [x] - removed `--experimental-debugger` flag  
- [x] - remove support for remote debugging
- [x] - remove `debugger-ui` from RNC CLI

### Test plan

- [x] - works in testers


thymikee (approved):
LFG

packages/dev-server/package.json
```diff
@@ -33,34 +33,28 @@
...
+    "@react-native-community/cli-server-api": "15.0.1",
     "@types/babel__code-frame": "^7.0.3",
     "@types/node": "18",
     "@types/ws": "^8.5.3",
     "babel-plugin-add-import-extension": "^1.6.0",
     "typescript": "^5.5.3"
   },
   "peerDependencies": {
-    "@react-native-community/cli-debugger-ui": ">=11.3.1",
     "@react-native-community/cli-server-api": ">=11.3.1"
```
thymikee, line 55: bump to 15.0.1

jbroma: do we want to bump it tho? We could bump to version that was associated with RN 0.74 🤔

jbroma: bumped to 13.6.4 which is associated with 0.74 👍

PR #787, opened 2024-11-08
fix: source-map matching in React Native DevTools

jbroma:
### Summary

Fix bad source-map matching in DevTools by aligning the following:

- [x] - add namespace to `SourceMapDevToolPlugin`
- [x] - align `sourceURL` that gets passed to JSI `evaluateJavascript`

Before:
image

After:
image

### Test plan

- [x] - proper mapping in DevTools in testers on iOS
- [x] - proper mapping in DevTools in testers on Android

### DIFF
--- a/packages/repack/android/src/main/java/com/callstack/repack/ScriptConfig.kt
+++ b/packages/repack/android/src/main/java/com/callstack/repack/ScriptConfig.kt
@@ -6,6 +6,7 @@ import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.RequestBody
 import okhttp3.RequestBody.Companion.toRequestBody
 import java.net.URL
+import java.net.URI
 
 data class ScriptConfig(
     val scriptId: String,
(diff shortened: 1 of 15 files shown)


packages/repack/android/src/main/java/com/callstack/repack/ScriptConfig.kt
```diff
@@ -33,13 +35,25 @@ data class ScriptConfig(
...
 
-            val url = URL(
-                    if (query != null) {
-                        "$urlString?$query"
-                    } else {
-                        urlString
-                    }
-            )
+            val initialUrl = URL(urlString)
+            val initialUri = initialUrl.toURI()
```
okwasniewski, line 39: nit: its almost impossible to distinguish those two variable names 🤣

packages/repack/ios/ScriptConfig.h
```diff
@@ -20,6 +20,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readonly) NSNumber *timeout;
 @property (readonly) NSString *verifyScriptSignature;
 @property (readonly) NSString *uniqueId;
+@property (readonly) NSString *sourceUrl;
```
okwasniewski, line 23: nit: a further improvement, you can probably make all of those properties nonatomic, https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocProperties.html

okwasniewski (approved):

packages/repack/android/src/main/java/com/callstack/repack/ScriptConfig.kt
jbroma: damn, good point 👍

packages/repack/ios/ScriptConfig.h
jbroma: what would be the benefit of that?

jbroma:
> Could you document the addition of sourceUrl somewhere?

It's not an addition per se - we should do a whole doc page on using DevTools and dealing with issues - the only thing that user needs to do here is to remember to set `output.uniqueName`

packages/repack/ios/ScriptConfig.h
okwasniewski: _speed_ 🚅, if you add nonatomic the access should be faster

jbroma: will do in a separate PR to keep this on topic, thx!

PR #803, opened 2024-11-25
refactor: split federation runtime plugin into separate plugins

jbroma:
### Summary

Splitting the runtime plugin in order to enable more granular control. FederationRuntimePlugin now becomes two runtime plugins:

- [x] - CorePlugin which enables loading of MF containers/chunks in RN context
- [x] - ResolverPlugin which maps the the MF loading requests to actual URLs.

Users can modify which default plugins are included via `defaultRuntimePlugins` in the MF2 Plugin config. For example, if users want to opt out of the `ResolverPlugin` they can set the prop to `['@callstack/repack/mf/core-plugin']` in order to include just the `CorePlugin`.

ResolverPlugin can be configured similar to how `RetryPlugin` from MF2 core: https://module-federation.io/plugin/plugins/retry-plugin.html#usage (i.e. creating own runtime plugin that wraps the ResolverPlugin)

### Test plan

- [x] - tests pass
- [x] - tester apps work


packages/repack/src/plugins/ModuleFederationPluginV2.ts
```diff
@@ -83,11 +85,14 @@ export interface ModuleFederationPluginV2Config
...
   private deepImports: boolean;
+  private disableDefaultRuntimePlugins: boolean;
 
   constructor(pluginConfig: ModuleFederationPluginV2Config) {
-    const { reactNativeDeepImports, ...config } = pluginConfig;
+    const { disableDefaultRuntimePlugins, reactNativeDeepImports, ...config } =
+      pluginConfig;
     this.config = config;
     this.deepImports = reactNativeDeepImports ?? true;
+    this.disableDefaultRuntimePlugins = disableDefaultRuntimePlugins ?? false;
```
thymikee, line 95: Not a fan of netagtions in publicAPI config. Consider changing this to `defaultRuntimePlugins: string[]` and allow user to override this with e.g. empty array in case they don't want default.

thymikee (approved):

packages/repack/src/plugins/ModuleFederationPluginV2.ts
jbroma: Good point, let's go with that 👍

jbroma: updated the PR with the changes we talked about in here 👍

thymikee (approved):
👌🏼

PR #817, opened 2024-12-09
feat: reanimated plugin

jbroma:
### Summary

Reanimated plugin that takes care of configuring module rules for the user. It simplifies the steps needed to get Reanimated working and limits the number of files the full transformation runs on.

### Test plan

- [x] - testers work with reaniamted


thymikee (approved):

Esemesek (approved):

jbroma:
@thymikee @Esemesek thanks for the reviews 🎉

PR #828, opened 2024-12-17
feat: wait for device before adb reverse

jbroma:
### Summary

When starting the dev server, wait for device (`adb wait-for-device`) before running `adb reverse`.

This also makes the ADB reverse more resilient by running reverse on all available devices (similar to how Expo does it)

### Test plan

- [x] - works in testers

### DIFF
--- a/packages/repack/src/commands/common/runAdbReverse.ts
+++ b/packages/repack/src/commands/common/runAdbReverse.ts
@@ -40,6 +40,8 @@ async function waitForDevice(logger: Logger) {
     await executeAdbCommand('wait-for-device', logger);
   } catch (error) {
     const message = (error as Error).message;
+    // Ignore the error if there are multiple devices/emulators
+    // we only care about about at least 1 device being online
     if (/more than one device\/emulator/.test(message)) {
       return;
     }
--- a/packages/repack/src/commands/common/runAdbReverse.ts
+++ b/packages/repack/src/commands/common/runAdbReverse.ts
@@ -11,9 +11,7 @@ interface RunAdbReverseParams {
 
 function getAdbPath() {
   const androidHome = process.env.ANDROID_HOME;
-  return androidHome
-    ? path.resolve(androidHome, 'platform-tools', 'adb')
-    : 'adb';
+  return androidHome ? path.join(androidHome, 'platform-tools', 'adb') : 'adb';
 }
 
 async function waitForDevice() {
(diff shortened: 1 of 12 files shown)


packages/repack/src/commands/common/runAdbReverse.ts
```diff
@@ -1,35 +1,54 @@
...
+}
+
+function getAdbPath() {
+  const androidHome = process.env.ANDROID_HOME;
+  return androidHome ? path.join(androidHome, 'platform-tools', 'adb') : 'adb';
+}
+
+async function waitForDevice() {
+  const adbPath = getAdbPath();
+  const command = `${adbPath} wait-for-device`;
```
thymikee, line 19: this waits for the device to be opened, but not booted yet. Test if that's enough for adb reverse to work

jbroma: it did work in testers without a hitch 👍

jbroma: my theoery is that `adb reverse` only interacts with the network layer of the Android VM and not the OS so no need for waitining for booted state

thymikee: also this will fail when more than 1 device is connected

jbroma: What's the failure scenario here? That it will fail to do adb reverse on the correct device?

jbroma: both `adb reverse` and `adb wait-for-device` can fail with more than 1 available, I've addressed this in the newest commit

packages/repack/src/commands/common/runAdbReverse.ts
```diff
@@ -1,35 +1,98 @@
...
+  }
+}
+
+async function waitForDevice(logger: Logger) {
+  try {
+    await executeAdbCommand('wait-for-device', logger);
+  } catch (error) {
+    const message = (error as Error).message;
+    if (/more than one device\/emulator/.test(message)) {
+      return;
```
thymikee, line 44: add a comment that this is expected and handled later

thymikee (approved):

packages/repack/src/commands/common/runAdbReverse.ts
jbroma: done 👍

jbroma:
thank you for the review @thymikee & @szymonrybczak

PR #829, opened 2024-12-17
fix: handle early JS errors like in Metro

jbroma:

### DIFF
--- a/packages/repack/src/plugins/RepackTargetPlugin/implementation/init.ts
+++ b/packages/repack/src/plugins/RepackTargetPlugin/implementation/init.ts
@@ -13,6 +13,18 @@ module.exports = function () {
 
   __webpack_require__.repack = $globalObject$.__repack__ = repackRuntime;
 
+  // intercept module factory calls to forward errors to global.ErrorUtils
+  __webpack_require__.i.push(function (options) {
+    var originalFactory = options.factory;
+    options.factory = function (moduleObject, moduleExports, webpackRequire) {
+      try {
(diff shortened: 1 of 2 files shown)


packages/repack/src/plugins/RepackTargetPlugin/implementation/init.ts
```diff
@@ -13,6 +13,18 @@ module.exports = function () {
...
   __webpack_require__.repack = $globalObject$.__repack__ = repackRuntime;
 
+  // intercept module factory calls to forward errors to global.ErrorUtils
+  __webpack_require__.i.push(function (options) {
+    var originalFactory = options.factory;
+    options.factory = function (moduleObject, moduleExports, webpackRequire) {
+      try {
+        originalFactory.call(this, moduleObject, moduleExports, webpackRequire);
+      } catch (e) {
+        $globalObject$.ErrorUtils.reportFatalError(e);
```
thymikee, line 23: can we link to Metro sources for reference?

thymikee (approved):

packages/repack/src/plugins/RepackTargetPlugin/implementation/init.ts
jbroma: done!

jbroma:
added some extra error handling for a rare scenario where there might be no `ErrorUtils` defined in the global scope 👍

jbroma:
thank you for the review @thymikee 🎉

PR #835, opened 2024-12-20
fix: ScriptManager import path in MF runtime plugins (CorePlugin & ResolverPlugin)

mrsasuu:
Fix for the reverted changes: https://github.com/callstack/repack/pull/834


jbroma (approved):
LGTM thx!

PR #836, opened 2024-12-20
fix: enable bridgeless mode to fix iOS test app build issues

mrsasuu:
The iOS test apps were failing to build due to the following error:
`/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.1.sdk/usr/include/c++/v1/__memory/unique_ptr.h:64:19 Invalid application of 'sizeof' to an incomplete type 'facebook::react::JSExecutorFactory'
`
Only happened with: `bridgeless_enabled => false`.

### DIFF
--- a/apps/tester-federation-v2/android/gradle.properties
+++ b/apps/tester-federation-v2/android/gradle.properties
@@ -40,8 +40,9 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
 # Note that this is incompatible with web debugging.
-#newArchEnabled=true
-#bridgelessEnabled=true
+newArchEnabled=true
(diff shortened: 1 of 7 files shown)


apps/tester-federation-v2/android/gradle.properties
```diff
@@ -40,8 +40,9 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
 # Note that this is incompatible with web debugging.
-#newArchEnabled=true
-#bridgelessEnabled=true
+newArchEnabled=true
```
mrsasuu, line 43: It was commented only in the tester-federation-v2

jbroma (approved):
LGTM 👍

PR #841, opened 2024-12-27
feat: add NativeWind plugin

borisyankov:
Implements a NativeWind plugin for Re.Pack.

Adds a demo with different NativeWind features to tester-app.

Simulator Screenshot - iPhone 16 Pro - 2025-01-07 at 03 38 48


jbroma:
Regarding `1. How to resolve 'react-native-css-interop'` - the issue lies with the nativewind itself and it's package setup - `pnpm` isolated linker only allows declared dependencies to be imported in the project which is a good thing - setting the node-linker to hoisted is a dirty workaround and should be avoided. 

For now let's just go with explicit dependency of `react-native-css-interop` in the project which also fixes the issue

jbroma:
Regarding other issues: 
2. `VerifyInstall` - this will never work since it's very closely tied to Metro, skip
3. Solved - reworked loader
4. Solved - should have worked regardless of nativewind

szymonrybczak (approved):

borisyankov:
LGTM.

jbroma (approved):

PR #855, opened 2025-01-08
perf: do 'lazy' comparison in 'checkIfCacheDataOutdated'

borisyankov:
Simplifies comparison with cached data. Also does not 'precompute' it before doing the comparison, so expression terminates on first difference.

### Summary

Not sure if it is a real problem as the code might be called infrequently. I was just bugged by the eager version.


borisyankov:
Changeset added!

jbroma (approved):
LGTM 🎉

Issue #866, opened 2025-01-13
iOS build failed in release mode


ra1nj:
### Describe the bug

After upgraded to repack5.0, the debug mode is fine,  but once i run the app with release mode, the build is failing:
```
UnableToResolveError: Unable to resolve module app1/App from /Users/xxx/Repack/RepackHost/App.tsx: app1/App could not be found within the project or in these directories:
```

### System Info

```shell
.
```

### Re.Pack Version

5.0.rc.2

### Reproduction

https://github.com/ra1nj/RepackHost/tree/repack_5.0

### Steps to reproduce

Clone reproduce repo and install deps then run in release mode

jbroma:
I've managed to confirm this and this needs to be fixed in 2 places:
- platform setup (since RN 76 changed some CLI commands used when bundling for releasing)
- `RepackTargetPlugin` needs some reworking to always include ScriptManager no matter what

jbroma:
hi @ra1nj, the [5.0.0-rc.7](https://github.com/callstack/repack/releases/tag/%40callstack%2Frepack%405.0.0-rc.7) is out which should fix the issue you were having, let me know if this fixes the problem for you so we can close this issue 👍

ra1nj:
I have upgraded the repack version to 5.0.0-rc.7 and pod install, but still facing the same issue:
```
/Users/vn564c0/.nvm/versions/node/v20.5.1/bin/node /Users/vn564c0/Repack/RepackHost/node_modules/react-native/scripts/bundle.js webpack-bundle --entry-file index.js --platform ios --dev false --reset-cache --bundle-output /Users/vn564c0/Library/Developer/Xcode/DerivedData/RepackHost-curtkupfvqvvmyaqkkdcwvttwpmj/Build/Products/Release-iphonesimulator/main.jsbundle --assets-dest /Users/vn564c0/Library/Developer/Xcode/DerivedData/RepackHost-curtkupfvqvvmyaqkkdcwvttwpmj/Build/Products/Release-iphonesimulator/RepackHost.app --minify false --config-cmd '/Users/vn564c0/.nvm/versions/node/v20.5.1/bin/node  /Users/vn564c0/Repack/RepackHost/node_modules/react-native/cli.js config'
 WARN  the transform cache was reset.
                Welcome to Metro v0.81.0
              Fast - Scalable - Integrated


/Users/vn564c0/Repack/RepackHost/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:148
[...]

PR #868, opened 2025-01-14
fix: decouple init & module error handling from runtime module

jbroma:
### Summary

partial solution to #866

In some cases `loadScript` runtime module might not get injected into the bundle. Since all of the logic for Repack runtime lives there, it might cause undefined behaviour like failing to process EarlyJS errors or failure to initialise Repack global object. 

This PR separates the `loadScript` part into it's own runtime module, and makes `init` and `moduleErrorHandler` parts separate runtime modules.

Previously in Rspack there was no way to add runtime modules - we could only edit existing ones and we were forced to concatenate init & load script modules into one. This is no longer the case so we can split them into separate entities.

The types injected into the global scope were also reworked - now most types are hidden behind a `RepackRuntimeGlobals` namespace to prevent polluting the global typescript scope

### Test plan

- [x] - testers work

### DIFF
--- a/packages/repack/src/plugins/RepackTargetPlugin/InitRuntimeModule.ts
+++ b/packages/repack/src/plugins/RepackTargetPlugin/InitRuntimeModule.ts
@@ -1,19 +1,32 @@
-import { RuntimeModule, Template } from '@rspack/core';
+import type {
+  Compiler,
+  RuntimeModule as RuntimeModuleType,
+} from '@rspack/core';
 
 interface InitRuntimeModuleConfig {
   globalObject: string;
 }
 
-export class InitRuntimeModule extends RuntimeModule {
-  constructor(private config: InitRuntimeModuleConfig) {
-    super('repack/init', RuntimeModule.STAGE_BASIC);
-  }
+export const makeInitRuntimeModule = (
+  compiler: Compiler,
+  moduleConfig: InitRuntimeModuleConfig
+): RuntimeModuleType => {
+  const Template = compiler.webpack.Template;
+  const RuntimeModule = compiler.webpack.RuntimeModule;
 
-  generate() {
-    return Template.asString([
-      Template.getFunctionContent(
-        require('./implementation/init.js')
-      ).replaceAll('$globalObject$', this.config.globalObject),
-    ]);
-  }
-}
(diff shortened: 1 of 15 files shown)


packages/repack/src/plugins/RepackTargetPlugin/InitRuntimeModule.ts
```diff
@@ -0,0 +1,19 @@
...
+export class InitRuntimeModule extends RuntimeModule {
+  constructor(private config: InitRuntimeModuleConfig) {
+    super('repack/init', RuntimeModule.STAGE_BASIC);
+  }
+
+  generate() {
+    return Template.asString([
+      Template.getFunctionContent(
+        require('./implementation/init.js')
+      ).replaceAll('$globalObject$', this.config.globalObject),
```
thymikee, line 16: any chance you could inline the `init.js` here so the implementation would live close to the plugin and there would be no need to replace `$spcialChars$`? same question applies for `LoadScriptRuntimeModule ` and `ModuleErrorHandlerRuntimeModule `

jbroma:
> Looks like a solid improvement with less hacks. Any ideas how you could test it?

I would just stick to existing integration tests that we have - this is a very internal change. The easiest way to test it in an additional way would be to compile a bundle and verify that the modules get injected properly.

packages/repack/src/plugins/RepackTargetPlugin/InitRuntimeModule.ts
jbroma: We have separate transpilation rules for the runtime code so it needs to live in a separate scope. NGL it's gonna be problematic and IMO not worth the effort.

jbroma:
I needed to make the runtime module classes dynamic to deal with the compatibility between webpack and rspack, all works good now on both platforms 👍

packages/repack/src/plugins/RepackTargetPlugin/InitRuntimeModule.ts
jbroma: to put this into perspective: similar pattern is used in Rspack and we have it a tad bit better here because we use TypeScript in the runtime code, while Rspack just uses blueprints in form of plain JS files

PR #869, opened 2025-01-15
feat: use RNC CLI explicitly when bundling the project

jbroma:
### Summary

- [x] - added noop `--config-cmd` to satisfy `bundle.js` script from RN core
- [x] - init now replaces the whole shellScript for "Bundle React Native code and images"
- [x] - init for iOS now exports `CLI_PATH` pointing to RNC CLI explicitly
- [x] - ~init for Android now configures `cliFile` and points to RNC CLI explicitly~
- [x] - dropped unofficial support for modifying `build.gradle` for projects using RN < 0.71
- [x] - use default `bundle` instead of `webpack-bundle` commands
- [x] - removed `modifyAndroid` phase of init altogether 

Partial solution to #866

### Test plan

- [x] - init works locally 
- [x] - project created with canary release builds for release


jbroma:
canary: https://www.npmjs.com/package/@callstack/repack/v/5.0.0-canary-20250115185948

needs bumping Rspack version manually, will be done inside separate PR

packages/init/src/tasks/modifyAndroid.ts
```diff
@@ -1,92 +1,67 @@
...
+  }`;
 
-  // Make sure we capture actual config and not the commented out one
-  const applyReactGradleRegex =
-    /(?<!`)apply from: "\.\.\/\.\.\/node_modules\/react-native\/react\.gradle"/;
-  const projectExtReactRegex = /(?<!\*\s)project\.ext\.react\s*=\s*\[/;
-  const reactConfigIndex = config.search(projectExtReactRegex);
+function updateBundleCommand(config: string): string {
+  const bundleCommandRegex = /(\/\/\s)?bundleCommand\s*=\s*".*?"/;
+  const bundleCommandContent = `bundleCommand = "webpack-bundle"`;
```
thymikee, line 19: I'd only set that if `webpack` is explicitly set as bundler, otherwise it's just `bundle` and it's the same as in Metro

jbroma: true, we can ditch this altogether or set it explicitly to `bundle` - might be useful when somebody has it set to something else

jbroma: we can ditch `webpack-start` and `webpack-bundle` altogether tbh, these are just for compat and maybe it's confusing for new users?

thymikee: 💯

jbroma: ok, I'll do `bundle` here and remove the webpack commands in a separate PR 👍

jbroma: for fresh project yes - for existing ones it's better to replace it since they might be using `bundleCommand = 'export:embed' for example

jbroma:
new canary: https://www.npmjs.com/package/@callstack/repack/v/5.0.0-canary-20250117101323

thymikee (approved):

PR #872, opened 2025-01-16
feat: enable inlining RN entrypoints

jbroma:
### Summary

Looking over the RN codebase I've found that `ReactNativePrivateInitializeCore` reexports `InitializeCore`. Inside commit history we can find [attempts at making this module a noop](https://github.com/facebook/react-native/commit/b14b34b232b4873137ea3f3c7a90f86b9013885a) but ultimately it was always reverted because of issues at metro. 

We're not using metro and this isn't an issue with webpack/rspack so no reason why not to try it, especially since this allows for inlining entrypoints for faster startup (before webpack bailed out of this optimization because `InitializeCore` was used in other places)

### Test plan

- [x] - testers work

### DIFF
--- a/.changeset/gorgeous-nails-watch.md
+++ b/.changeset/gorgeous-nails-watch.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+Enable inlining entry modules by making runtime initialization from React renderers a no-op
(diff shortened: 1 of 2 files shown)


thymikee (approved):
Any measurable improvements?

jbroma:
> Any measurable improvements?

measurable? maybe
noticeable? nope - but we would probably need to go to low-end android devices to see the benefit tbh

PR #874, opened 2025-01-18
docs: update React Navigation page

szymonrybczak:
### Summary

With React Navigation 7.x there's [`screenLayout`](https://reactnavigation.org/docs/group/#screen-layout) which perfectly fits into a scenario mentioned as an example in React Navigation page.


szymonrybczak:
Btw I don't see this page in new 5.x docs?

jbroma:
> Btw I don't see this page in new 5.x docs?

yeah, the plan was to rework code-splitting altogether there - let's merge this as it is and we will see how we teach this in 5.X

jbroma (approved):
LGTM, thanks for that @szymonrybczak

PR #875, opened 2025-01-18
chore(docs): get year in footer dynamically

szymonrybczak:
### Summary

to do not update it every year :D


jbroma (approved):
I'm curious to see if this break somehow along the way, but lets do it :D

PR #877, opened 2025-01-18
feat(init): add choose bundler prompt

szymonrybczak:
### Summary

Adds a prompt to choose a bundler:

CleanShot 2025-01-18 at 17 21 24@2x

### DIFF
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -1,4 +1,3 @@
-import { select } from '@inquirer/prompts';
 import addDependencies from './tasks/addDependencies.js';
 import checkPackageManager from './tasks/checkPackageManager.js';
 import checkReactNative from './tasks/checkReactNative.js';
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -1,3 +1,4 @@
+import { select } from '@inquirer/prompts';
 import addDependencies from './tasks/addDependencies.js';
 import checkPackageManager from './tasks/checkPackageManager.js';
 import checkReactNative from './tasks/checkReactNative.js';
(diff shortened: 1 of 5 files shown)


packages/init/src/index.ts
```diff
@@ -33,6 +34,31 @@ export default async function run({
 
     checkReactNative(cwd);
 
+    if (!bundler) {
```
jbroma, line 37: lets move this to a separate file inside tasks

packages/init/src/index.ts
```diff
@@ -33,6 +34,31 @@ export default async function run({
...
+      const isCI = process.env.CI === 'true';
```
jbroma, line 38: maybe instead of checking for `true` lets check if it's not 0 and not `false`, like in LoggerPlugin:

```
const isTruthyEnv = (env: string | undefined) => {
      return !!env && env !== 'false' && env !== '0';
    };
```

jbroma (commented):
2 things to address

packages/init/src/index.ts
```diff
@@ -1,10 +1,12 @@
+import { select } from '@inquirer/prompts';
```
jbroma, line 1: not needed here anymore

packages/init/src/tasks/selectBundler.ts
```diff
@@ -0,0 +1,25 @@
+import { select } from '@inquirer/prompts';
+import logger from '../utils/logger.js';
+import { isTruthyEnv } from '../utils/isTruthyEnv.js';
```
jbroma, line 3: order import needs fixing (biome should this for you)

jbroma (approved):
cleanup and then its LGTM 👍

jbroma:
thanks for working on this @szymonrybczak 🎉

PR #880, opened 2025-01-19
fix: use specific swc assumptions instead of general loose mode

jbroma:
### Summary

before too many loose mode transformations were enabled, being not spec compliant with `@react-native/babel-preset` - now only the default ones are provided with the config - making the compiled code more strict and spec compliant.

### Test plan

- [x] - all testers work


jbroma:
fixes the issue encountered after migrating from rc.5 to rc.6

cc @maciekBudzinski @borisyankov

PR #892, opened 2025-01-22
fix: force container name to be a valid js identifier

lukewalczak:
### Summary

Adding a check in `ModuleFederationPluginV2` to validate that a container's name is a valid JavaScript identifier.

It fixes the following error:

```
✖ [11:12:29.747Z][DevServer]   × Library name base (mini-app) must be a valid identifier when using a var declaring library type. Either use a valid identifier (e. g. mini_app) or use a different library type (e. g. `type: 'global'`, which assign a property on the global scope instead of declaring a variable). Common configuration options that specific library names are 'output.library[.name]', 'entry.xyz.library[.name]', 'ModuleFederationPlugin.name' and 'ModuleFederationPlugin.library[.name]'.
 {
  reqId: 'req-2',
  err: {
    type: 'Error',
    message: "  × Library name base (mini-app) must be a valid identifier when using a var declaring library type. Either use a valid identifier (e. g. mini_app) or use a different library type (e. g. `type: 'global'`, which assign a property on the global scope instead of declaring a variable). Common configuration options that specific library names are 'output.library[.name]', 'entry.xyz.library[.name]', 'ModuleFederationPlugin.name' and 'ModuleFederationPlugin.library[.name]'.\n",
    stack: "Error:   × Library name base (mini-app) must be a valid identifier when using a var declaring library type. Either use a valid identifier (e. g. mini_app) or use a different library type (e. g. `type: 'global'`, which assign a property on the global scope instead of declaring a variable). Common configuration options that specific library names are 'output.library[.name]', 'entry.xyz.library[.name]', 'ModuleFederationPlugin.name' and 'ModuleFederationPlugin.library[.name]'.\n",
    code: 'GenericFailure'
  }
} 
```

### Test plan

- [x] - added unit test
- [x] - tested manually within `tester-federation-v2` repository:

Zrzut ekranu 2025-01-22 o 16 27 30

### DIFF
--- a/packages/repack/src/plugins/ModuleFederationPluginV2.ts
+++ b/packages/repack/src/plugins/ModuleFederationPluginV2.ts
@@ -117,7 +117,7 @@ export class ModuleFederationPluginV2 implements RspackPluginInstance {
     if (!isIdentifier(name)) {
       const error = new Error(
         `[ModuleFederationPlugin] The container's name: '${name}' must be a valid JavaScript identifier. ` +
-          'Please correct it to proceed.'
+          'Please correct it to proceed. For more information, see: https://developer.mozilla.org/en-US/docs/Glossary/Identifier'
       );
       // remove the stack trace to make the error more readable
       error.stack = undefined;
--- a/packages/repack/src/plugins/ModuleFederationPluginV2.ts
+++ b/packages/repack/src/plugins/ModuleFederationPluginV2.ts
@@ -1,5 +1,6 @@
 import type { moduleFederationPlugin as MF } from '@module-federation/sdk';
 import type { Compiler, RspackPluginInstance } from '@rspack/core';
+import { name as isIdentifier } from 'estree-util-is-identifier-name';
 import { isRspackCompiler } from './utils/isRspackCompiler.js';
 
 type JsModuleDescriptor = {
@@ -53,7 +54,7 @@ export interface ModuleFederationPluginV2Config
  * import * as Repack from '@callstack/repack';
  *
  * new Repack.plugins.ModuleFederationPlugin({
- *   name: 'host,
+ *   name: 'host',
  * });
  * ```
  *
@@ -62,7 +63,7 @@ export interface ModuleFederationPluginV2Config
  * import * as Repack from '@callstack/repack';
  *
  * new Repack.plugins.ModuleFederationPlugin({
- *   name: 'host,
+ *   name: 'host',
  *   shared: {
  *     react: Repack.Federated.SHARED_REACT,
  *     'react-native': Repack.Federated.SHARED_REACT,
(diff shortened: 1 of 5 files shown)


packages/repack/src/plugins/__tests__/ModuleFederationPluginV2.test.ts
```diff
@@ -244,4 +244,37 @@ describe('ModuleFederationPlugin', () => {
...
+
+  it('should not throw an error for a valid container name', () => {
+    const invalidContainerNames = [
+      'app_name',
+      'appName',
+      'appName123',
+      '$appName',
+    ];
+
+    invalidContainerNames.forEach((name) => {
```
jbroma, line 274:
```suggestion
    validContainerNames.forEach((name) => {
```

packages/repack/src/plugins/__tests__/ModuleFederationPluginV2.test.ts
```diff
@@ -244,4 +244,37 @@ describe('ModuleFederationPlugin', () => {
...
+        new ModuleFederationPluginV2({ name }).apply(mockCompiler);
+      }).toThrow(
+        `[ModuleFederationPlugin] The container's name: '${name}' must be a valid JavaScript identifier. ` +
+          'Please correct it to proceed.'
+      );
+    });
+  });
+
+  it('should not throw an error for a valid container name', () => {
+    const invalidContainerNames = [
```
jbroma, line 267:
```suggestion
    const validContainerNames = [
```

packages/repack/src/plugins/ModuleFederationPluginV2.ts
```diff
@@ -111,12 +112,25 @@ export class ModuleFederationPluginV2 implements RspackPluginInstance {
     ];
   }
 
+  private validateModuleFederationContainerName(name: string | undefined) {
+    if (!name) return;
+    if (!isIdentifier(name)) {
+      const error = new Error(
+        `[ModuleFederationPlugin] The container's name: '${name}' must be a valid JavaScript identifier. ` +
+          'Please correct it to proceed.'
```
jbroma, line 120: would be nice to have a link to reference what constitutes a valid identifier like this: https://developer.mozilla.org/en-US/docs/Glossary/Identifier

jbroma (commented):
few nits

packages/repack/src/plugins/ModuleFederationPluginV2.ts
lukewalczak: Sure, added!

jbroma:
@lukewalczak please also add changeset by running `pnpm changeset` in root of the monorepo

jbroma (approved):

jbroma (approved):

PR #893, opened 2025-01-23
feat: rework repack-init

jbroma:
### Summary

- [x] - reworked how init looks
- [x] - leave installing dependencies to the user
- [x] - guide for next steps

https://github.com/user-attachments/assets/1ba5f6df-7546-4488-8d08-21bb5c86a495

Mostly based on `create-rsbuild` 🙌 

### Test plan

- [x] - works locally
- [x] - canary release works


packages/init/src/tasks/checkPackageManager.ts
```diff
@@ -1,17 +1,56 @@
...
+  if (rootDir) {
+    const result = await preferredPM(rootDir);
+    if (result) {
+      packageManager = result.name;
+      logger.info(`Detected ${packageManager} as package manager`);
+      return PM_COMMANDS[packageManager];
+    }
+  }
+
+  const candidate = process.argv0;
```
thymikee, line 43: can you pass this as a function argument?

packages/init/src/utils/logger.ts
```diff
@@ -1,15 +1,18 @@
-import chalk from 'chalk';
+import { log } from '@clack/prompts';
 
 let verbose = false;
 
+type LogFn = (message: string) => void;
+
+const verboseWrapper = (logFn: LogFn) => {
+  return (message: string) => (verbose ? logFn(message) : undefined);
```
thymikee, line 8: good call

packages/init/src/tasks/modifyDependencies.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const packageJson = fs.readFileSync(packageJsonPath, 'utf-8');
+  const { version } = JSON.parse(packageJson) as { version: string };
+
+  return '~' + version;
+}
+
+async function getLatestVersion(dependency: string) {
+  try {
+    const { stdout } = await execa('npm', ['view', dependency, 'version']);
```
thymikee, line 22: this is gonna be slow compared to a list of supported dependency versions, plus latest versions may be incompatible. ideally you would have a json file tied to the version of repack with specific versions – you'd get safety and speed

thymikee (commented):
Proper improvement 👍🏼

packages/init/src/tasks/checkPackageManager.ts
jbroma: dropped this logic altogether since it wasn't reliable

packages/init/src/tasks/modifyDependencies.ts
jbroma: yeah, even doing network requests for stuff like this feels very wasteful

jbroma: hardcoded in root now 👍

jbroma:
latest canary: https://www.npmjs.com/package/@callstack/repack-init/v/5.0.0-canary-20250123192942

thymikee (approved):

jbroma:
bundled canary: https://www.npmjs.com/package/@callstack/repack-init/v/5.0.0-canary-20250123203953

PR #896, opened 2025-01-24
feat: expo-modules plugin

jbroma:
### Summary

Added expo-modules plugin package that fills the gap for integrating expo-modules with Re.Pack - for now the plugin only defines the `EXPO_OS` env var in runtime which is normally set by `babel-preset-expo`. 

### Test plan

- [x] - no `EXPO_OS` warning present in runtime after adding the plugin


jbroma:
needs a separate PR with normalisation of `compiler.name`

PR #913, opened 2025-02-04
fix: correct 'postcss-loader' peer dependency version

borisyankov:
The last version of `postcss-loader` is 8.1.1


jbroma (commented):
Nice catch 🎉 Can we make the changeset more meaningful, like:

"Fixed incorrect peer dependency version of `postcss-loader`"

jbroma (approved):

jbroma:
thanks @borisyankov 👍

PR #942, opened 2025-02-04
ci: skip tests when only website changes

szymonrybczak:
### Summary

Skips tests when change only in `website/`


szymonrybczak:
We have to merge this one to actually test it, because under this PR tests will be triggered because there's a change outside of `website/` folder 😅

jbroma (approved):
lets merge and try it out 👍

PR #948, opened 2025-02-05
docs: use `5.x` as version in dev mode

szymonrybczak:
### Summary

As we intensively work on docs currently switching every time to `5.x` is annoying so in development mode I switched to `5.x` by default. 

### Test plan

Deployed version should use `4.x`


jbroma (approved):
:shipit:

PR #950, opened 2025-02-05
fix: don't error on missing `--entry-file` when entry is defined in `rspack/webpack.config.js`

szymonrybczak:
### Summary

We shouldn't exit with an error if there's no `--entry-file` option passed, because in we have the same value in `rspack.config.js` under `entry`.


jbroma (approved):
LGTM, nice one 👍

PR #955, opened 2025-02-05
fix: use already downloaded bundle

hosseinmd:
if an script doesn't cached the steps to run it is download, write, read, run
there is no need to read it again, 
This changes will improve performance.


hosseinmd:
@jbroma please review

jbroma (approved):
LGTM 👍 Very nice find, thank you @hosseinmd!

PR #968, opened 2025-02-10
refactor: normalise plugin names and error formatting

jbroma:
### Summary

- [x] - every plugin registers with Rspack/webpack API using `Repack<plugin>` convention
- [x] - every plugin reports errors with the plugi name in front in brackets 

### Test plan

n/a


jbroma:
> 👌. How about Reanimated and Natewind plugins?

oh nice find! forgot about those :D

PR #975, opened 2025-02-12
feat: deprecate `getPublicPath` and make it a no-op

jbroma:
### Summary

- [x] - make `getPublicPath` a no-op and deprecate it
- [x] - add automatic public path handling in `DevelopmentPlugin` for development
- [x] - use `noop:///` as default `publicPath`
- [x] - remove deprecated `getPublicPath` function usage from all config files
- [x] - remove `get-public-path` from documentation

### Test plan

- [x] - testers work
- [x] - tests pass


jbroma:
cc @szymonrybczak

thymikee:
> we could make this non breaking by making the getPublicPath helper function a noop (since it will be configured internally anyways in the exact same way)
can mark is as deprecated and scheduled for removal

let's do this and inform the user that they can safely remove it

thymikee (approved):

PR #977, opened 2025-02-12
[BREAKING] feat: internal `SourceMapPlugin`

jbroma:
### Summary

- [x] - consolidated uses of `SourceMapDevToolPlugin` into a separate `SourceMapPlugin`
- [x] - sourcemaps can now be configured through `devtool` option
- [x] - unsupported sourcemap formats cause a configuration error forcing user to change the configuration
- [x] - removed `devtool: false` from all configs and templates since that will essentially disable the sourcemap
- [x] - `devtool: 'source-map'` is now the default (matches existing behaviour)
  
### Test plan

- [x] - tests pass
- [x] - testers work

### DIFF
--- a/packages/repack/src/plugins/SourceMapPlugin.ts
+++ b/packages/repack/src/plugins/SourceMapPlugin.ts
@@ -1,0 +1,70 @@
+import type { Compiler, RspackPluginInstance } from '@rspack/core';
+
+export class SourceMapPlugin implements RspackPluginInstance {
+  apply(compiler: Compiler) {
+    if (!compiler.options.devtool) {
+      return;
+    }
+
+    const platform = compiler.name as string;
+    const format = compiler.options.devtool;
+    const devtoolModuleFilenameTemplate =
+      compiler.options.output.devtoolModuleFilenameTemplate;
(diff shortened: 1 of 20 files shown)


packages/repack/src/plugins/SourceMapPlugin.ts
```diff
@@ -0,0 +1,84 @@
...
+      compiler.options.output.devtoolFallbackModuleFilenameTemplate;
+
+    if (
+      format === 'eval' ||
+      format === 'eval-source-map' ||
+      format === 'eval-cheap-source-map' ||
+      format === 'eval-cheap-module-source-map' ||
+      format === 'eval-nosources-source-map' ||
+      format === 'eval-nosources-cheap-source-map' ||
+      format === 'eval-nosources-cheap-module-source-map'
```
thymikee, line 39: how about this?
```suggestion
      format.startsWith('eval')
```

jbroma: oh yeah totally - I was trying to narrow down typescript string literals that make up the format type but now it seems useless, we can def simplify this 👍

thymikee (approved):

PR #981, opened 2025-02-13
feat:  add defaults for `config.output` and `config.optimization`

jbroma:
### Summary

- [x] - added defaults for `config.output`
- [x] - added defaults for `config.optimization`
- [x] - added normalization of output path in order to resolve `[platform]` & `[context]` placeholders (useful in static configs when there is no access to `platform` variable)
- [x] - removed default configuration from configs & templates

### Test plan

- [x] - tests pass
- [x] - testers work


jbroma:
Added two changesets since it wouldn't make much sense in the release notes otherwise

szymonrybczak (approved):
Awesome!! 🔥

PR #997, opened 2025-02-17
feat: getAssetTransformRules

jbroma:
### Summary

- [x] - added `getAssetTransformRules` which creates a default configuration for asset handling in both `rspack` and `webpack` configurations
- [x] - added tests for `getAssetTransformRules`
- [x] - added docs for `getAssetTransformRules`
- [x] - updated templates
- [x] - updated configs in testers  

### Test plan

- [x] - tests pass
- [x] - testers work

### DIFF
--- a/.changeset/eighty-phones-deny.md
+++ b/.changeset/eighty-phones-deny.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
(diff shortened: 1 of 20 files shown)


thymikee (approved):
add docs and note the svg support and we're good 👍🏼

PR #998, opened 2025-02-17
fix: remove `checkReactNative` util

szymonrybczak:
### Summary

Previously when tried to run `init` in subdirectory (a directory without `package.json`) - `init` would fail to specify React Native version in the project.


.changeset/warm-clocks-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@callstack/repack-init": patch
+---
+
+Remove `checkReactNative` util function to allow creating in exsiting monorepos
```
jbroma, line 5:
```suggestion
Remove `checkReactNative` util function from init
```

jbroma (approved):
1 nit and LGTM 👍

PR #1010, opened 2025-02-18
docs: setup proper versioning

szymonrybczak:


website/rspress.config.ts
```diff
@@ -56,6 +56,14 @@ export default defineConfig({
     ],
   },
   builderConfig: {
+    source: {
+      define: {
+        'global.__REPACK_DOC_VERSION__': JSON.stringify(
```
jbroma, line 61: was global required here? might be cleaner without it but thats a nit anyways

website/rspress.config.ts
```diff
@@ -5,9 +5,13 @@ import { pluginFontOpenSans } from 'rspress-plugin-font-open-sans';
 import vercelAnalytics from 'rspress-plugin-vercel-analytics';
 import { defineConfig } from 'rspress/config';
 
+const LATEST_VERSION = 'v5';
+
 export default defineConfig({
-  root: path.join(__dirname, 'src'),
-  title: 'Re.Pack',
+  root: path.join(__dirname, 'src', process.env.REPACK_DOC_VERSION ?? 'latest'),
```
jbroma, line 11: nit: move to the top of config

jbroma (approved):
2 nits, but LGTM 👍

website/rspress.config.ts
szymonrybczak: it wasn't but I just like this convention more.

PR #1033, opened 2025-02-25
fix: OutputPlugin not using bundleOutput & sourcemapOutput args

jbroma:
### Summary

- [x] - fixed a type that prevent setting up the env variables for OutputPlugin correctly

### Test plan

- [x] - testers work


jbroma:
cc @draggie

PR #1049, opened 2025-02-26
chore: remove examples

jbroma:
### Summary

removed link to examples repo since they are not up to date and we will work on this later this quater

### Test plan

n/a


prosandmusic:
🎬

PR #1050, opened 2025-02-26
docs: use gradient in logo

jbroma:
### Summary

- [x] - use gradient in logo to be consistent with the CLI

### Test plan

n/a


thymikee (approved):
Dope

PR #1062, opened 2025-03-03
docs: fix ReanimatedPlugin import

ammarahm-ed:
I was migrating my project to Re.Pack v5 and found some issues in docs.


jbroma (approved):
Thanks for your contribution!

Issue #996, opened 2025-02-16
Add dependency helper hints


borisyankov:
### Description

Users might be unaware that some libraries require an existing plugin in order to work.

### Suggested solution

If Reanimated, NativeWind (or in future other dependencies) are detected in `package.json` but the corresponding Re.Pack plugin is not installed/configured, show a helpful hint letting the user know and link to the plugin's `readme.md`.

PR #1066, opened 2025-03-03
feat: add detecting dependency specific plugins

szymonrybczak:
### Summary

Closes https://github.com/callstack/repack/issues/996

### Test plan

1. Go to `apps/tester-app` and remove `NativeWindPlugin` from `rspack.config.ts`. 
2. Run `pnpm start` the warning regarding lack of plugin should be presented.


packages/repack/src/commands/rspack/start.ts
```diff
@@ -51,6 +51,7 @@ export async function start(
     rootDir: cliConfig.root,
     platforms: args.platform ? [args.platform] : detectedPlatforms,
     reactNativePath: cliConfig.reactNativePath,
+    noRecommendedPlugins: args.noRecommendedPlugins,
```
jbroma, line 54: can be removed

packages/repack/src/commands/consts.ts
```diff
@@ -41,3 +41,22 @@ export const DEV_SERVER_ASSET_TYPES = new RegExp(
     '^@mf-types.d.ts$',
   ].join('|')
 );
+
+export const DEPENDENCIES_WITH_SEPARATE_PLUGINS: Record<
```
jbroma, line 45: I think we can move this to where validate lives

packages/repack/src/commands/options.ts
```diff
@@ -71,6 +71,10 @@ export const startCommandOptions = [
       '[DEPRECATED] Path to a bundler config file, e.g webpack.config.js. Please use --config instead.',
     parse: (val: string) => path.resolve(val),
   },
+  {
+    name: '--no-recommended-plugins',
```
jbroma, line 75: RNC CLI does this magic with `--no` flags, take a look at `interactive` flag

packages/repack/src/commands/common/config/validateRequiredPlugins.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const validateRequiredPlugins = (
+  rootDir: string,
+  plugins: Set<string>,
+  bundler: 'rspack' | 'webpack'
+) => {
+  const packageJson = JSON.parse(
+    readFileSync(join(rootDir, 'package.json'), 'utf-8')
+  );
+  const dependencies = Object.keys(packageJson.dependencies);
```
jbroma, line 14: try catch this with a meaningful error

packages/repack/src/commands/common/config/makeCompilerConfig.ts
```diff
@@ -69,5 +70,16 @@ export async function makeCompilerConfig<C extends ConfigurationObject>(
...
   );
 
+  if (!options.args.noRecommendedPlugins) {
+    const activePlugins = new Set(
+      normalizedConfigs
+        .flatMap((c) => ('plugins' in c ? c.plugins : []))
+        .map((p) => p?.constructor.name)
+        .filter((p): p is string => p !== undefined)
+    );
+
```
jbroma, line 80: should be part of validate plugins function

packages/repack/src/commands/common/config/validateRequiredPlugins.ts
```diff
@@ -0,0 +1,38 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { bold } from 'colorette';
+import { DEPENDENCIES_WITH_SEPARATE_PLUGINS } from '../../consts.js';
+
+const validateRequiredPlugins = (
```
jbroma, line 6: these are not really required so we should name this like `validatePluginConfiguration`

jbroma (changes requested):
this is a feature so we can work this on a little longer since 5.1 needs to include OTA & rsbuild too.

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,80 @@
...
+  const activePlugins = new Set(
+    configs
+      .flatMap((c) => ('plugins' in c ? c.plugins : []))
+      .map((p) => p?.constructor.name)
+      .filter((p): p is string => p !== undefined)
+  );
+
+  try {
+    const packageJson = JSON.parse(
+      readFileSync(join(rootDir, 'package.json'), 'utf-8')
```
jbroma, line 42: I think we can just do dynamic import here, would also provide better type safety

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,80 @@
...
+      );
+    }
+    throw new CLIError(
+      'Failed to parse package.json: ' +
+        (error instanceof Error ? error.message : 'Unknown error')
+    );
+  }
+
+  dependencies
+    .filter((d) => {
```
jbroma, line 59: no single word variables please

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,80 @@
...
+      const plugin = DEPENDENCIES_WITH_SEPARATE_PLUGINS[d];
+
+      if (!plugin) {
+        return false;
+      }
+
+      return plugin.bundler ? plugin.bundler === bundler : true;
+    })
+    .forEach((d) => {
```
jbroma, line 68: no single word variables please

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,80 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { CLIError } from '@react-native-community/cli-tools';
```
jbroma, line 3: use our own CLIError

jbroma (changes requested):
few things to fix in the PR

I'm also thinking about generalising this into `recommendations` instead of just `recommendedPlugins`. We could potentially add more recommendations about loaders or other settings based on user config in the future. This way we ensure the CLI API doesn't grow in the future and there is no need to rework this.

There is also discrepancy between recommendations and validation in here that needs to be addressed somehow 🤔 - we are not really validating configs here, we're just creating recommendations.

I'm still torn whether we should enable this by default 🤔 

cc @thymikee

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,77 @@
...
+  );
+
+  try {
+    const packageJson = require(join(rootDir, 'package.json'));
+    dependencies = Object.keys(packageJson.dependencies || {});
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
+      throw new CLIError(
+        'Could not find package.json in your project root. ' +
+          'Make sure you are running this command from your React Native project root directory.'
```
thymikee, line 46: Isn't `rootDir` always the same regardless of the CWD? It should be where node_modules are or where configured, so you can run commands from nested folders

thymikee (commented):
Since these are warnings, and assuming no other tools are using your output to parse something, I would get rid of the flag and just turn it on by default. Reality is that developers will rarely read warnings, and what I would consider is whether this should be an error. I'd start with the warning though and wait for some community feedback, then once you're certain this logic is good switch to erroring

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
jbroma: it comes from RNC CLI so you guys know best :D

thymikee: CLI uses `find-up` to find the root based on package.json, so we should take this for granted that package.json exists, otherwise the CLI would fail

szymonrybczak: It looks like `process.cwd()` is also always equal to `rootDir` 🙈
```
❯ pwd
/_/repack/apps/tester-app/ios

~/c/repack/a/tester-app/ios feat/detect-…ific-plugins !1
❯ nr start

{
  cwd: '/_/repack/apps/tester-app',
  rootDir: '/_/repack/apps/tester-app'
}
```

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,76 @@
+import { join } from 'node:path';
```
jbroma, line 1:
```suggestion
import path from 'node:path';
```

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const activePlugins = new Set(
+    configs
+      .flatMap((c) => ('plugins' in c ? c.plugins : []))
+      .map((p) => p?.constructor.name)
+      .filter((p): p is string => p !== undefined)
+  );
+
+  try {
+    const packageJson = require(join(rootDir, 'package.json'));
```
jbroma, line 40:
```suggestion
    const packageJson = require(path.join(rootDir, 'package.json'));
```

packages/repack/src/commands/common/config/validatePluginConfiguration.ts
```diff
@@ -0,0 +1,76 @@
...
+  configs: C[],
+  bundler: 'rspack' | 'webpack'
+) {
+  let dependencies: string[] = [];
+
+  const activePlugins = new Set(
+    configs
+      .flatMap((c) => ('plugins' in c ? c.plugins : []))
+      .map((p) => p?.constructor.name)
+      .filter((p): p is string => p !== undefined)
```
jbroma, line 36: no single name variables please

packages/repack/src/commands/types.ts
```diff
@@ -14,6 +14,7 @@ export interface BundleArguments {
   watch?: boolean;
   config?: string;
   webpackConfig?: string;
+  recommendedPlugins?: boolean;
```
jbroma, line 17: should be removed since flag was removed

packages/repack/src/commands/types.ts
```diff
@@ -31,6 +32,7 @@ export interface StartArguments {
   verbose?: boolean;
   config?: string;
   webpackConfig?: string;
+  recommendedPlugins?: boolean;
```
jbroma, line 35: should be removed since flag was removed

packages/repack/src/commands/common/config/makeCompilerConfig.ts
```diff
@@ -69,5 +70,11 @@ export async function makeCompilerConfig<C extends ConfigurationObject>(
     normalizeConfig(config, options.platforms[index])
   );
 
+  await validatePluginConfiguration(
+    rootDir,
+    normalizedConfigs,
```
jbroma, line 75: nit but we pass configs but the function name says `validatePluginConfiguration` - lets adjust either one of them

packages/repack/src/commands/common/config/makeCompilerConfig.ts
```diff
@@ -65,9 +66,15 @@ export async function makeCompilerConfig<C extends ConfigurationObject>(
   );
 
   // normalize the configs
-  const normalizedConfigs = configs.map((config, index) =>
+  const normalizedConfig = configs.map((config, index) =>
```
jbroma, line 69:
```suggestion
  const normalizedConfigs = configs.map((config, index) =>
```

packages/repack/src/commands/common/config/makeCompilerConfig.ts
```diff
@@ -65,9 +66,15 @@ export async function makeCompilerConfig<C extends ConfigurationObject>(
...
     normalizeConfig(config, options.platforms[index])
   );
 
-  return normalizedConfigs as C[];
+  const plugins = normalizedConfig.flatMap((config) =>
+    'plugins' in config ? config.plugins : []
+  );
+
+  await validatePlugins(rootDir, plugins, options.bundler);
```
jbroma, line 77: the warning will get printed twice right? because we're combining both platforms here

szymonrybczak: Why? It does work correctly:

```
❯ p start

> tester-app@0.0.1 start /Users/szymonrybczak/callstack/repack/apps/tester-app
> react-native webpack-start

WARNING: Detected nativewind package which requires NativeWindPlugin plugin but it's not configured. Please add the following to your configuration file.
Read more https://github.com/callstack/repack/tree/main/packages/plugin-nativewind/README.md.

▄▀▀▀ ▀▀▀▀   █▀▀█ █▀▀█ ▄▀▀▀ █  █
█    ▀▀▀▀   █▀▀▀ █▀▀█ █    █▀▀▄
▀    ▀▀▀▀ ▀ ▀    ▀  ▀  ▀▀▀ ▀  ▀
5.0.0, powered by Rspack

 r: Reload app
 d: Open developer menu
 j: Open debugger
 a: Run adb reverse

Press Ctrl+c or Ctrl+z to quit the dev server

ℹ [19:22:18.940Z][DevServer] Server listening at http://127.0.0.1:8081
ℹ [19:22:18.942Z][DevServer] Server listening at http://[::1]:8081
ℹ [19:22:18.942Z][DevServer] Starting build for platforms: ios, android
```

szymonrybczak:
@jbroma is there anything that is blocking us here?

packages/repack/src/commands/common/config/makeCompilerConfig.ts
jbroma: ah yeah, nvm my bad, I looked at the logic and its all good 👍

jbroma:
> @jbroma is there anything that is blocking us here?

nah, lets just merge & release https://github.com/callstack/repack/pull/1090 first and then we can merge this and the next release will be 5.1

jbroma (approved):

PR #1067, opened 2025-03-03
fix(init): override commit for new project

szymonrybczak:
### Summary

Override commit when creating new project so git status won't be dirty after initialization.


.changeset/five-cats-hope.md
jbroma, line 1: please be more verbose :D

packages/init/src/tasks/overrideCommit.ts
```diff
@@ -0,0 +1,17 @@
+import { execa } from 'execa';
+
+const overrideCommit = async (rootDir: string) => {
+  try {
+    await execa('git', ['--version']);
```
jbroma, line 5: do we need this? the second one will fail the same way if git is not present

packages/init/src/tasks/overrideCommit.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    // Amend the existing commit without changing the message, RNC CLI creates a new commit with a message "Initial commit" already.
+    await execa('git', ['commit', '--amend', '--no-edit'], {
+      cwd: rootDir,
```
jbroma, line 9: format this to be in the line above if possible

packages/init/src/tasks/overrideCommit.ts
```diff
@@ -0,0 +1,17 @@
+import { execa } from 'execa';
+
+const overrideCommit = async (rootDir: string) => {
```
jbroma, line 3: match the format as in other tasks

jbroma (changes requested):

jbroma (approved):

packages/init/src/tasks/overrideCommit.ts
```diff
@@ -0,0 +1,17 @@
+import { execa } from 'execa';
+
+export default async function overrideCommit(rootDir: string) {
+  try {
+    await execa('git', ['add', '.'], {
+      cwd: rootDir,
+    });
+    // Amend the existing commit without changing the message, RNC CLI creates
+    // new commit with a message "Initial commit" already.
+    await execa('git', ['commit', '--amend', '--no-edit'], {
```
jbroma, line 10: what happens when there is already git history here? We dont want to override last commit in the monorepo for example

jbroma (changes requested):

packages/init/src/tasks/overrideCommit.ts
szymonrybczak: resolved in https://github.com/callstack/repack/pull/1067/commits/c7a6e83a8ef51b48c80be9aaacea85dec54284b0

packages/init/src/tasks/setupGit.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (isRepo) {
+      return;
+    }
+
+    // Amend the commit
+    await execa('git', ['init'], { cwd: rootDir });
+    await execa('git', ['add', '.'], { cwd: rootDir });
+    await execa('git', ['commit', '-m', 'Initial commit'], { cwd: rootDir });
+  } catch (e) {
+    // Silently return if any git command fails
```
jbroma, line 27: do we want to silently return here if this fails?

szymonrybczak: Yeah, if we fail to create a Git repo, that's a small thing; there are tools that don't init it at all.

jbroma: just this and we're g2g here

jbroma (approved):
LGTM 👍

PR #1074, opened 2025-03-06
fix: remove enforcement of output filenames in dev

szymonrybczak:
### Summary

After removing and testing in debug everything works the same way as before in TesterApp.

### Test plan

1. Open TesterApp, everything should work in the same way as before 👍

### DIFF
--- a/.changeset/good-apes-grab.md
+++ b/.changeset/good-apes-grab.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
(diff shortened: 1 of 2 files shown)


jbroma (commented):
looks good, change the name to "fix" instead of "chore" since it does change things + add a changeset please

szymonrybczak:
@jbroma done ✅

jbroma (approved):
LGTM 👍

PR #1075, opened 2025-03-06
feat: noramlize publicPath

szymonrybczak:
### Summary

Normalize public path and allow for overrides. From now users will be able to set in the config:

```
    output: {
      uniqueName: 'tester-app',
      publicPath: '[protocol]://[host]:[port]/[platform]/',
    },
```

and this will be properly replaced with current values. 

### Test plan

1. TesterApp works without setting publicPath
2. TesterApp uses publicPath if set.


packages/repack/src/commands/common/config/getCommandConfig.ts
```diff
@@ -19,6 +22,9 @@ function getBundleCommandDefaults() {
     optimization: {
       minimize: true,
     },
+    output: {
+      publicPath: '[protocol]://[host]:[port]/[platform]/',
+    },
```
jbroma, line 27: we def dont want this set in prod :)

packages/repack/src/commands/common/config/normalizeConfig.ts
```diff
@@ -60,11 +79,17 @@ export function normalizeConfig<C extends ConfigurationObject>(
...
+  /* normalize public path by resolving [protocol], [host], [port] & [platform] placeholders */
+  if (config.output?.publicPath) {
     normalizedConfig.output = {
       ...normalizedConfig.output,
-      publicPath: 'noop:///',
+      publicPath: normalizePublicPath(
+        config.output.publicPath ?? '[protocol]://[host]:[port]/[platform]/',
+        config.devServer?.protocol ?? 'http',
+        config.devServer?.host ?? 'localhost',
+        config.devServer?.port ?? 8081,
```
jbroma, line 90: lets not use any defaults here - solely because we will have yet another place with defaults defined

packages/repack/src/commands/common/config/getCommandConfig.ts
```diff
@@ -9,6 +9,9 @@ function getStartCommandDefaults() {
       hot: true,
       server: 'http',
     },
+    output: {
+      publicPath: '[protocol]://[host]:[port]/[platform]/',
```
jbroma, line 13: everything is known at this point except for `platform` so let's just normalize that part

jbroma (approved):
LGTM 👍

PR #1079, opened 2025-03-11
fix(init): wrong project root dir when migrating with init

jbroma:
### Summary

- [x] - fixes the issue where migrating existing project would fail because wrong root dir would be picked by `repack-init`

### Test plan

- [x] - migration with init works locally


szymonrybczak (commented):
👍

szymonrybczak (approved):

PR #1081, opened 2025-03-12
refactor: drop `@react-native-community/cli-server-api` dep

### LINKED ISSUE #1080: Dependency issue on pnpm monorepo with Expo app and `node-linker=hoisted`
### Documentation is

- [x] Missing
- [ ] Outdated
- [ ] Confusing
- [ ] Not sure?

### Explain in Detail

While trying to setup `repack` for an Expo app inside a pnpm monorepo, I had the following error:

```
error Named export 'openStackFrameInEditorMiddleware' not found. The requested module '@react-native-community/cli-server-api' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@react-native-community/cli-server-api';
const { openStackFrameInEditorMiddleware, openURLMiddleware, } = pkg;
.
file:///home/busy/dev/VANGUARD/node_modules/@callstack/repack-dev-server/dist/plugins/devtools/devtoolsPlugin.js:1
import { openStackFrameInEditorMiddleware, openURLMiddleware, } from '@react-native-community/cli-server-api';
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'openStackFrameInEditorMiddleware' not found. The requested module '@react-native-community/cli-server-api' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@react-native-community/cli-server-api';
const { openStackFrameInEditorMiddleware, openURLMiddleware, } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:180:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:263:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:547:26)
    at async Object.start [as func] (/home/busy/dev/VANGUARD/node_modules/@callstack/repack/dist/commands/rspack/start.js:56:7)
    at async Command.handleAction (/home/busy/dev/VANGUARD/node_modules/@react-native-community/cli/build/index.js:139:9)
```

This was caused by `@callstack/repack-dev-server` requiring version `15.0.1` of `@react-native-community/cli-server-api`, but my Expo app already had version `18.0.0` installed because of its dependencies. This (probably) would've worked fine with non-hoisted modules.

This version does not export `openStackFrameInEditorMiddleware` under the same name anymore (it became `openStackFrameMiddleware` on v17.0.0 with [this](https://github.com/react-native-community/cli/pull/2588) commit). 

**EDIT**: While investigating further it looks like the whole module is exported differently.
In the case of docs change this doesn't really matter - but it could be useful for debugging. 

---

I recognize this isn't the fault of repack, and it should not be treated as such - but it's a pitfall I could see many developers fall into without understanding the reason or thinking they did something wrong.

I believe it should be acknowledged somewhere in the documentation that it might not work "straight out of the box", especially with Expo or other miscellaneous wrappers that already depend on RN modules and thus will not allow complete control of repack dependencies versions.

### Your Suggestion for Changes

I'm not sure how to word this, sorry.

I'm not even sure this is a good suggestion haha (sorry if it's not!).

jbroma:
### Summary

Closes #1080 

- [x] - removed `@react-native-community/cli-server-api` from dev server
- [x] - replaced it with own implementation based on the reference implementation 

### Note

We still need to fix `/open-stack-frame` behaviour since it seems to be broken for some time (we need to rewrite `webpack://` urls into actual filepaths

dedicated bug report: https://github.com/callstack/repack/issues/1082

### Test plan

- [x] - testers work

### DIFF
--- a/.changeset/weak-candles-dream.md
+++ b/.changeset/weak-candles-dream.md
@@ -1,0 +1,6 @@
+---
+"@callstack/repack-dev-server": patch
+"@callstack/repack": patch
+---
+
+Drop dependency on `@react-native-community/cli-server-api` in the DevServer
--- a/packages/dev-server/package.json
+++ b/packages/dev-server/package.json
@@ -35,6 +35,8 @@
     "fastify": "^4.24.3",
     "fastify-favicon": "^4.3.0",
     "fastify-plugin": "^4.5.1",
+    "launch-editor": "^2.10.0",
+    "open": "^10.1.0",
     "pretty-format": "^28.1.0",
     "source-map": "^0.7.4",
     "ws": "^8.7.0"
@@ -42,19 +44,10 @@
   "devDependencies": {
     "@babel/cli": "^7.23.9",
     "@babel/core": "^7.23.9",
-    "@react-native-community/cli-server-api": "15.0.1",
     "@types/babel__code-frame": "^7.0.3",
     "@types/node": "catalog:",
     "@types/ws": "^8.5.3",
     "babel-plugin-add-import-extension": "^1.6.0",
     "typescript": "catalog:"
-  },
-  "peerDependencies": {
-    "@react-native-community/cli-server-api": ">=13.6.4"
-  },
-  "peerDependenciesMeta": {
-    "@react-native-community/cli-server-api": {
-      "optional": true
-    }
   }
 }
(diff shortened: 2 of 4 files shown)


thymikee (approved):
make sure editor integration works 👍🏼

jbroma:
> make sure editor integration works 👍🏼

opened a separate issue for that: https://github.com/callstack/repack/issues/1082

szymonrybczak (approved):

PR #1087, opened 2025-03-12
refactor: drop RNC CLI peer deps

jbroma:
### Summary

There is no need for having `@react-native-community/cli` as a peer dependency and as #1080 it might be blocking for some setups so following the same path as in #1081 it was removed in Re.Pack core itself

### Test plan

- [x] - testers work


jbroma:
leaving this for 5.1

thymikee (approved):
LFG

PR #1099, opened 2025-03-20
feat: update `getResolveOptions`

jbroma:
### Summary

Since Metro enables package exports by default now, it's about time we finish the support for package exports in Re.Pack as well. For now we will keep it disabled by default but we might enable it in the next minor/major version depending on feedback from users.

Alignment with updated tests from Metro allowed to find missing pieces in the package exports configuration which in turn enabled package exports to work in all of the testers 🎉 

- [x] - enable `packageImports` at all times since metro supports it
- [x] - use `import` / `require` condition name at all times when source type is known
- [x] - add platform & native extensions when using package exports
- [x] - update `metro-resolver` tests  

### Test plan

- [x] - tests pass
- [x] - testers work with package exports enabled

### DIFF
--- a/.changeset/eighty-yaks-kick.md
+++ b/.changeset/eighty-yaks-kick.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+Enable support for package imports by default when resolving modules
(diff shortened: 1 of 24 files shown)


jbroma:
canary for testing: https://www.npmjs.com/package/@callstack/repack/v/6.0.0-canary-20250321121727

PR #1132, opened 2025-04-27
[chore] add enums for error messages

12ya:


jbroma (changes requested):

12ya:
> Hi @12ya,
> 
> 
> 
> what's the rationale behind this change?

hey, @jbroma.
thought it's gonna be good to use enums for messages instead of strings, and easier maintenance in case it gonna get larger

jbroma (changes requested):
This doesn't compile properly and needs to be fixed - but I'm also not sure this is worth it at this point 🤔 

image

jbroma:
Closing since the PR went stale

PR #1135, opened 2025-04-28
refactor: use `@react-native/assets-registry/registry`

jbroma:
### Summary

based on the suggestion from @huntie, instead of using deep import from RN core, we can import `@react-native/assets-registry/registry` directly.

### Test plan

- [x] - testers work

### DIFF
--- a/.changeset/polite-lizards-divide.md
+++ b/.changeset/polite-lizards-divide.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
+Use `@react-native/assets-registry/registry` instead of deep import from React Native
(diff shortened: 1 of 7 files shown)


jbroma:
upon further tests turns out it's not that simple - when we use deep import of AssetRegistry from core it results in a singleton registry module (since all imports of `@react-native/assets-registry/registry` are from within `react-native`)

if we however request the module from the context of project it will result in duplicate inclusion of assets registry, causing the module to be included twice in the build and not pointing to the same registry

jbroma:
> @jbroma We probably want some abstraction then similar to [`rn-get-polyfills.js`](https://github.com/facebook/react-native/blob/main/packages/react-native/rn-get-polyfills.js). I say "similar" in concept, not in exact design (this root file is somewhat horrible 😅). But it's an API design problem we'll have to take on over the coming months.
> 
> Making a note to discuss this at the Core Contributors Summit.

also not a big fan of root level entrypoints 😅

I've managed to almost get around this by rerouting all imports in the bundle to a virtual module from Re.Pack (kinda like a `rn-get-polyfills.js`) but then in the tests this fails because `@react-native/assets-registry` is not a peer dependency of Re.Pack 🙃 - it's a rather hacky solution to the problem.

I'll close this PR for now and we will tackle this in the future 👍

PR #1172, opened 2025-06-12
fix: sourcemaps handling in dev & prod

jbroma:


jbroma:
missing tests + handling for `SourceMapPlugin` for scenarios when dev server is not available - we should align with Metro and always use absolute paths

PR #1175, opened 2025-06-14
feat: extend signature to allow svgr options

### LINKED ISSUE #1171: Extend signature of getAssetTransformRules
### Description

I'm using the `getAssetTransformRules` function in the following way:
```
Repack.getAssetTransformRules({svg: 'svgr'})
```

To reduce the amount of boilerplace, I would love to pass some options to `@svgr/webpack` like `dimensions: false`.

### Suggested solution

Would you consider a contribution in order to extend the signature of `getAssetTransformRules` to allow an additional option name `svgOptions`, or something like that?

If not, I would like to contribute using the right approach that works for you.

### Additional context

_No response_

ilteoood:
### Summary

Closes #1171

### DIFF
--- a/.changeset/three-cups-juggle.md
+++ b/.changeset/three-cups-juggle.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+feat: allow svgr options from getAssetTransformRules
--- a/packages/repack/src/utils/getAssetTransformRules.ts
+++ b/packages/repack/src/utils/getAssetTransformRules.ts
@@ -4,11 +4,17 @@ import {
   getAssetExtensionsRegExp,
 } from './assetExtensions.js';
 
-function getSvgRule(type: 'svgr' | 'xml' | 'uri') {
-  if (type === 'svgr') {
+function getSvgRule(type: SvgType) {
+  const isTypeObject = typeof type === 'object';
+
+  if (type === 'svgr' || (isTypeObject && type?.type === 'svgr')) {
+    const additionalOptions = isTypeObject && type?.options;
     return {
       test: /\.svg$/,
-      use: { loader: '@svgr/webpack', options: { native: true } },
+      use: {
+        loader: '@svgr/webpack',
+        options: { native: true, ...additionalOptions },
+      },
     };
   }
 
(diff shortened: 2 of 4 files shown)


packages/repack/src/utils/getAssetTransformRules.ts
```diff
@@ -18,6 +24,12 @@ function getSvgRule(type: 'svgr' | 'xml' | 'uri') {
   };
 }
 
+type SvgType =
+  | 'svgr'
+  | 'xml'
+  | 'uri'
+  | { type: 'svgr'; options: Record<string, any> };
```
jbroma, line 31: can we move this to the top of the file?

website/src/latest/api/utils/get-asset-transform-rules.md
```diff
@@ -67,13 +67,15 @@ See [`assetsLoader` documentation](/api/loaders/assets-loader#remoteassetpath) f
 
 ### options.svg
 
-- Type: `'svgr' | 'xml' | 'uri'`
+- Type: `'svgr' | 'xml' | 'uri' | { type: "svgr", options: Record<string, any>}`
 
 Determines how SVG files should be processed:
 
 - `'svgr'`: Uses `@svgr/webpack` to transform SVGs into React Native components
+  - `type: "svgr"`: Uses `@svgr/webpack` to transform SVGs into React Native components and allows to configure additional options to `@svgr/webpack`.
```
jbroma, line 75: lets mention this together with `svgr` first and then at the bottom, lets add another paragraph that says that you can configure `svgr` with additional options through the object config

jbroma (commented):
looks good, lets improve the docs tho 👍

website/src/latest/api/utils/get-asset-transform-rules.md
ilteoood: I hope this will be fine

.changeset/three-cups-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@callstack/repack": minor
+---
+
+feat: allow svgr options from getAssetTransformRules
```
jbroma, line 5:
```suggestion
feat: allow passing additional SVGR options through `getAssetTransformRules`
```

jbroma (commented):
1 last nit

jbroma (approved):

jbroma:
Thank you for your contribution @ilteoood 🎉

jbroma:
heads up: since this is a feature it will be released with `Re.Pack 5.2` - we still need to collect some more features and then we will release so it might be a while - at least 2 weeks I'm guessing

PR #1213, opened 2025-07-07
[FIX] fix(android): load inspector libraries on initialization

raghav-reglobe:
### Summary
**Problem**  
Opening the DevTools network-inspector on Android crashes with  
`UnsatisfiedLinkError: ...InspectorNetworkRequestListener.onHeaders(...)`  
when the app is built with React Native 0.80 or later.

**Root cause**  
RN 0.80 stopped pre-loading `libreactnativejni.so`, which contains the
`InspectorNetworkRequestListener` JNI stubs.  
Re.Pack initialises its `ScriptManagerModule` before RN has loaded the library.

**Fix**  
Call `SoLoader.loadLibrary("reactnativejni")` once during
`ScriptManagerModule` initialisation, **guarded by `BuildConfig.DEBUG`** so it
runs only in debug builds.

### Implementation details
* Added `ensureInspectorLoaded()` in `ScriptManagerModule`.
* Guarded with a `@Volatile` flag to ensure the load happens exactly once.
* No impact on release APKs (code path is skipped).

#### Preconditions
* Android device/emulator.
* App built with **RN 0.80+** and this patch applied.

#### Steps
1. Launch the app.
2. Open the dev menu (⌘ M / Ctrl M or shake).
3. Press **`j`** in the terminal to open the JS debugger.

#### Expected
* App remains running.
* A Chrome tab opens and successfully attaches to the JS runtime.

#### Before this fix
* The app crashed immediately with the `UnsatisfiedLinkError`.

### Changelog
* **Android**: Fix crash when opening DevTools on RN 0.80 + by lazily loading `libreactnativejni.so`.

### Checklist
- [x] Ran on a physical device
- [x] Verified debug build still strips inspector code in release
- [ ] Added a unit test (N/A – runtime linker issue)


jbroma:
@raghav-reglobe I've reviewed the PR and our DevTools integration and found that these changes are not needed. In principle, `ScriptManager` module should not have any dependency / reliance on the DevTools integration, so the changes in this PR are not fixing the root cause. 

I've made necessary fixes for RN 80 compatibility in #1227 and got the DevTools working for all of the testers in the monorepo.

Sorry for the delayed response, and thank you for your time & contribution nonetheless. I'll close this PR since it's not needed anymore, but please do not feel discouraged from submitting another contributions in the future!

raghav-reglobe:
Thanks

Issue #1098, opened 2025-03-20
Opening Devtools from iOS emulator crashes app with `/open-debugger not found` after 5.0.0


daveskybet:
### Describe the bug

When attempting to open Devtools via the iOS emulator (i imagine also on other platforms/devices given the issue) either with the warnings dialogue or dev menu results in the app crashing with the error:

```
TypeError: this[writeSym] is not a function
```

Which apart from seemingly being an issue with Pino, digging deeper the underlying error is:

```
[17:11:45.663Z][DevServer] File open-debugger not found {
  reqId: 'req-12',
  err: {
    type: 'CLIError',
    message: 'File open-debugger not found',
    stack: 'CLIError: File open-debugger not found\n' +
      '    at Compiler.getSource (/Users/david.hales/dev/tbd/apps/bf/native/node_modules/@callstack/repack/dist/commands/rspack/Compiler.js:207:13)\n' +
      '    at async Object.handler (file:///Users/david.hales/dev/tbd/node_modules/@callstack/repack-dev-server/dist/plugins/compiler/compilerPlugin.js:36:31)',
    name: 'CLIError'
  }
}
```

Even attempting to hit `http://localhost:8081/open-debugger?...` in browser returns a 404.

v5.0.0 is fine and opens devtools without error, v5.0.1/2 suffer a different error(`Conflict: Multiple assets emit different content to the same filename index.bundle`) so can't test, but that error is resolved by v5.0.3 where then the above issue appears. 

This issue seems consistent across multiple projects, from a large project i'm migrating to your own `super-app-showcase`

Opening Devtools via `j` in terminal does succeed however.

### System Info

```shell
System:
  OS: macOS 14.6.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 413.83 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.14.0
    path: ~/.asdf/installs/nodejs/22.14.0/bin/node
  Yarn:
    version: 3.2.1
    path: ~/.asdf/installs/nodejs/22.14.0/bin/yarn
  npm:
    version: 10.9.2
    path: ~/.asdf/installs/nodejs/22.14.0/bin/npm
  Watchman:
    version: 2025.03.03.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods: Not Found
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.2
      - iOS 17.2
[...]

PR #1228, opened 2025-07-25
feat: obtain`@react-native/dev-middleware` dynamically

jbroma:
### Summary

This change removes the pinned dependency on `@react-native/dev-middleware` in `@callstack/repack-dev-server` and instead dynamically resolves and imports it based on the React Native version used in the target project.

- [x] Remove direct dependency on `@react-native/dev-middleware` from dev-server package
- [x] Add dynamic resolution of dev middleware to match project's React Native version
- [x] Update both Rspack and Webpack start commands to use dynamic dev middleware resolution

This ensures compatibility with different React Native versions without forcing users to a specific dev middleware version.

Should resolve problems like in https://github.com/callstack/repack/issues/1098

### Test plan

- [x] - DevTools work in testers

### DIFF
--- a/.changeset/beige-humans-bake.md
+++ b/.changeset/beige-humans-bake.md
@@ -1,0 +1,6 @@
+---
+"@callstack/repack-dev-server": minor
+"@callstack/repack": minor
+---
+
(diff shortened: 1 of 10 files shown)


jbroma:
cc @thymikee @satya164 @V3RON would love your insights on this 🙏

PR #1237, opened 2025-07-29
fix: temporarily use Terser for minifying Rspack projects

jbroma:
### Summary

Ship with `terser-webpack-plugin` and use it preconfigured for Rspack configuration which should be using `SwcJsMinimizerRspackPlugin` preconfigured.

This change is temporary and the default will be removed once https://github.com/web-infra-dev/rspack/issues/11183 is resolved in Rspack

### Test plan

- [x] - testers production bundle gets minified correctly

### DIFF
--- a/.changeset/tall-towns-pay.md
+++ b/.changeset/tall-towns-pay.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
+Use preconfigured Terser as minimizer for Rspack configurations


longb1997:
look like https://github.com/web-infra-dev/rspack/issues/11183 was merged

jbroma:
> look like [web-infra-dev/rspack#11183](https://github.com/web-infra-dev/rspack/issues/11183) was merged

https://github.com/callstack/repack/pull/1258 builds on this PR to include that 👍

PR #1239, opened 2025-07-30
chore: add badge and link to Twitter social network

operagxsasha:


jbroma:
@operagxsasha thanks for the PR but we are fine with the current badges as they are, please don't feel discouraged from contributing in the future!

PR #1242, opened 2025-07-31
fix plugin-nativewind: accumulate compiled nativewind styles 

makivlach:
Solves having multiple MFE sharing rules generated by nativewind

### Summary

Solves: https://github.com/callstack/repack/discussions/1032 

### Test plan

1. Have 1. MFE - declare nativewind and tailwind there
2. Have 2. MFE - declare a separate nativewind configuration or use different utility classes not used in 1st 
3. Connect 1 and 2 to Host app. Previously, one MFE override the other due to registerCompile calling anew. Now they are reusing previous state

### DIFF
--- a/packages/plugin-nativewind/src/loader.ts
+++ b/packages/plugin-nativewind/src/loader.ts
@@ -23,6 +23,17 @@ export default function nativeWindLoader(
     const jsCss = cssToReactNativeRuntime(source, options);
     const code = dedent`
       import { StyleSheet } from "nativewind";
+      
+      const originalRegisterCompiled = StyleSheet.registerCompiled;
+      let rules = {}
+      
+      // Monkey patch StyleSheet.registerCompiled to merge rules across multiple MFE
+      StyleSheet.registerCompiled = (compiledStyles) => {
+        rules = Object.assign(compiledStyles.rules, rules);
+        compiledStyles.rules = rules;
+        return originalRegisterCompiled(compiledStyles)
+      }
+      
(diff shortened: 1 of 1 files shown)


jbroma:
@makivlach any reason why you’ve closed this PR?

makivlach:
> @makivlach any reason why you’ve closed this PR?

Yep, it was too unreliable in various situations. We've been dealing with this the whole day at work with my colleague and at the end of the day, we just decided that the most realible solution is really to just let tailwind/nativewind know about all used styling classes used throughout all MFEs

sahajarora1286:
Hey @makivlach , I am the author of the original issue post ( https://github.com/callstack/repack/discussions/1032 ) .  
Wondering if you were able to create a standard system to let nativewind know about all used styling classes used throughout all MFEs ? 

I had somehow managed to get all styling working as expected in the host and MFE apps in the dev builds. However, in release builds, the MFE app crashes when it uses a component that internally calls the `tva` tailwind utility to build classnames based on the tailwind design tokens.  The same code / component works just fine in the `host` app but it crashes in the MFE. On debugging, I realized that the generated MFE app bundles (e.g app1.container.bundle) don't contain the tailwind tokens (e.g text-primary-600). 
I believe that tree-shaking at some step causes the classname strings to be purged wherever a component tries to dynamically construct them using the `tva` utility. 

I was wondering if you had run into a similar issue which is why you had to let nativewind know of all the classnames throughout all MFEs before hand ?

PR #1244, opened 2025-08-02
fix: async config return type

V3RON:
### Summary

This pull request updates the return type of the async config function for RsPack. Currently, it is declared as RspackConfiguration, but it should be RepackRspackConfig to align with the other elements in the config type union.

### Test plan

Use an async function to configure the test app. Note that it currently doesn't accept the correct configuration. Check out this branch to see the difference.


jbroma:
nice catch, thank you 🙏

PR #1271, opened 2025-08-23
Add expo plugin

ceopaludetto:
### Summary

- Add an expo plugin that allows expo router + expo CNG (expo managed projects) to work with repack.
- Add a new guide to Expo

### Test plan

- Added a new expo managed tester application

### Todos

- [ ] Test Webpack Integration

### DIFF
--- a/.changeset/fluffy-mice-roll.md
+++ b/.changeset/fluffy-mice-roll.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack-plugin-expo": minor
+---
+
+Add Expo Router + CNG Plugin
--- a/.changeset/full-nails-grin.md
+++ b/.changeset/full-nails-grin.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack-plugin-expo": minor
+---
+
+Add Expo Managed tester application
--- a/apps/tester-expo-app/.gitignore
+++ b/apps/tester-expo-app/.gitignore
@@ -1,0 +1,7 @@
+# Build artifacts
+.expo
+build
+
+# Native directories
+/android
+/ios
--- a/apps/tester-expo-app/app.json
+++ b/apps/tester-expo-app/app.json
@@ -1,0 +1,16 @@
+{
+  "name": "tester-expo-app",
+  "slug": "tester-expo-app",
+  "scheme": "testerexpoapp",
+  "version": "0.0.1",
+  "android": {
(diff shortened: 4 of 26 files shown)


jbroma:
> ### Test plan
> Maybe we should create an expo managed example in `./apps` folder

Hey @ceopaludetto, absolutely, go for it 👍

ceopaludetto:
@jbroma added a new expo guide in the docs and also added the tester application. We can combine both `expo-modules` and `expo` guides in just one page. Maybe we can also merge both plugins in just one, its up to you!

j0nl1:
I'm tracking the progress in this PR, very interested on it. I just wondering if would be possible to launch just with expo instead of react-native cli since it wouldn't require to manage pods locally or any kind of build just fetching a development build from expo cloud and run it using the repack bundler.

ceopaludetto:
> I'm tracking the progress in this PR, very interested on it. I just wondering if would be possible to launch just with expo instead of react-native cli since it wouldn't require to manage pods locally or any kind of build just fetching a development build from expo cloud and run it using the repack bundler.

Hey, thanks for the interest! The main reason we switched from expo CLI to React Native CLI is that when you run something like `expo run ios --mode Release` expo always builds the metro bundle before the native build (you'd need a patch to avoid this). Expo CLI works fine for development though (just use `expo run ios --no-bundler` to skip Metro).
As for expo cloud, I haven't really played around with that feature much, so I can't say for sure how it would work with this setup. If you end up experimenting with it, would love to hear how it goes!

jbroma (changes requested):
hey @ceopaludetto sorry for the delay with the review! I found few things we need to address but the rest looks good 👍 

one more thing: when doing prebuild + run-ios I kept running into issues because I don't have cocoapods globally installed and use local ruby setup - perhaps we could fix that with Gemfile + Gemfile.lock like in other testers? (or maybe there is a nicer fix for that). Prebuild + android worked fine for me, good job with the tester 👍

Rnbsov:
waiting for merge sooooo much!!!!!!!

woals4815:
> waiting for merge sooooo much!!!!!!!

@Rnbsov  me either!!!!!!

dannyhw:
Since this pr has become stale, conflicts and we now have a new pr for expo support I'll be closing this one
Thanks for your work on this, we will make expo work soon even if in a slightly different way.

PR #1286, opened 2025-09-08
fix: inherit source type from babel

V3RON:
### Summary

Babel can detect whether a file being transformed is an ESM or CJS and return this information. However, SWC assumes by default that it will transform only ESM files. This can lead to issues with default export interop in certain cases.
See: https://github.com/callstackincubator/rozenite/issues/96

### Test plan

Use the reproduction repository provided in the linked issue. Link this branch with the app and try running it. If there are no errors reported, everything works as expected.


jbroma (approved):
LGTM 👍

Issue #1312, opened 2025-11-24
Whats the status of nativewind support


Kotletka89:
### Description

I would like to see #1271 merged because I think it would fix nativewind support

### Suggested solution

merge #1271

jbroma:
which PR are you referring to?

longb1997:
@Kotletka89 Show some respect, dude — @jbroma is one of the most active maintainers who has helped me a lot. Which PR are you referring to?

Kotletka89:
@jbroma sorry for disrespect but I meant that#1271

Kotletka89:
https://github.com/callstack-internal/vitalii-datsiuk-onboarding/pull/2

Kotletka89:
@jbroma @longb1997 up

Kotletka89:
there is also no react native web support ;(

ElielC:
> [callstack-internal/vitalii-datsiuk-onboarding#2](https://github.com/callstack-internal/vitalii-datsiuk-onboarding/pull/2)

This PR seems to fix a issue with nativewind on a project that uses re.pack if you are facing issues with nativewind on your project maybe applying the same fix on your project could fix it

dannyhw:
I edited your issue to represent the question you're actually asking, please remember that open source is given to you as a gift and remember to consider the people behind the screen.

Rnbsov:
+1 on this one, tailwind is soooo essential, really missing it 

or uniwind support would be nice ;( 

please collaborators, consider adding support please

dannyhw:
@Rnbsov looking into it

dannyhw:
the changes in #1315 should fix this

heres a example repo with the patch applied that shows nativewind working

https://github.com/dannyhw/repack_nativewind

heres the patch


you can see the patch file called `@callstack%2Frepack@5.2.2.patch`, with this content:
```diff
diff --git a/dist/loaders/babelSwcLoader/swc.js b/dist/loaders/babelSwcLoader/swc.js
index 03e548fb45bf0aaac0598c07c9630d5fcb76a795..ea482dbd92b2b314f18ea784e940d0846b50f05f 100644
--- a/dist/loaders/babelSwcLoader/swc.js
+++ b/dist/loaders/babelSwcLoader/swc.js
@@ -50,8 +50,8 @@ function getTransformReactRuntimeConfig(swcConfig, reactRuntimeConfig = {
         ...swcConfig.jsc?.transform,
         react: {
[...]

Issue #1295, opened 2025-09-26
Nativewind doesn't work with Rspack and SWC loader


sahajarora1286:
### Describe the bug

Hello, this is an issue found while attempting to resolve [this original issue](https://github.com/callstack/repack/issues/1289). 

I migrated to RsPack (from Webpack) and to latest Repack (v5.2.1). 

I was able to get the host app launching in dev mode without any errors.

However, Nativewind doesn't seem to be working at all now, i.e, nativewind styles are no longer applied. They worked when webpack was the bundler.

Is there any additional work that needs to be done to Nativewind configuration to get it working with RsPack ?

(NativeWind version is latest and unchanged from `v4.1.23`.)

Snippets from my rspack.config.mjs :

```
  module: {
      rules: [
        ...Repack.getJsTransformRules(),
        ...Repack.getAssetTransformRules(),
      ],
    },
    plugins: [
      /**
       * Configure other required and additional plugins to make the bundle
       * work in React Native and provide good development experience with
       * sensible defaults.
       *
       * `Repack.RepackPlugin` provides some degree of customization, but if you
       * need more control, you can replace `Repack.RepackPlugin` with plugins
       * from `Repack.plugins`.
       */
      new Repack.RepackPlugin(),
      new ReanimatedPlugin(),
      new NativeWindPlugin(),
      new rspack.IgnorePlugin({
        resourceRegExp: /^@react-native-masked-view/,
      }),
    ],
```

After adding verifyInstallation from nativewind to my app entry component, here's a more specific error that I get from nativewind:

`jsxImportSource was not set to 'nativewind'.`

Nativewind's official setup involves adding the nativewind preset and some config to babel config. However now that babel is not being used due to SWC, we're probably missing that step ?

I tried updating Repack.getJsTransformRules as such:
```
...Repack.getJsTransformRules({
          swc: {
            jsxRuntime: "automatic",
            importSource: "nativewind",
          },
        }),
```

which led to another error:

```
Error: Your 'metro.config.js' has overridden the 'config.resolver.resolveRequest' config setting in a non-composable manner. Your styles will not work until this issue is resolved. Note that 'require('metro-config').mergeConfig' is a shallow merge and does not compose existing resolveRequest functions together.
```
I also do have "jsxImportSource": "nativewind" set in my tsconfig.

This has become a blocker in a crucial project that I'm driving, and any help would be highly appreciated.

Environment info:

### System Info

```shell
System:
  OS: macOS 15.5
[...]

PR #1315, opened 2025-12-08
fix: nativewind broken by import source being overwritten in default case

dannyhw:
fixes: #1312  #1295

### Summary

Fixes NativeWind integration being broken when using SWC loader.

The issue was in `getTransformReactRuntimeConfig` which unconditionally overwrote `runtime` and `importSource` from `reactRuntimeConfig`, even when those values were `undefined`. This would clobber the custom `importSource` that NativeWind sets in the SWC config (e.g., `nativewind`).

The fix uses nullish coalescing (`??`) to preserve existing values when `reactRuntimeConfig` doesn't provide overrides.

### Test plan

1. Set up a project with NativeWind configured
2. Run the app with SWC loader
3. Verify NativeWind styles work correctly (previously they would break because `importSource` was overwritten)

### DIFF
--- a/packages/repack/src/loaders/babelSwcLoader/swc.ts
+++ b/packages/repack/src/loaders/babelSwcLoader/swc.ts
@@ -77,9 +77,7 @@ function getTransformReactDevelopmentConfig(
 
 function getTransformReactRuntimeConfig(
   swcConfig: SwcLoaderOptions,
-  reactRuntimeConfig: Record<string, any> = {
-    runtime: 'automatic',
-  }
+  reactRuntimeConfig: Record<string, any> = {}
 ): SwcLoaderOptions {
   return {
     ...swcConfig,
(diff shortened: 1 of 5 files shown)


packages/repack/src/loaders/babelSwcLoader/swc.ts
```diff
@@ -89,8 +89,12 @@ function getTransformReactRuntimeConfig(
...
         react: {
           ...swcConfig.jsc?.transform?.react,
-          runtime: reactRuntimeConfig.runtime,
-          importSource: reactRuntimeConfig.importSource,
+          runtime:
+            reactRuntimeConfig.runtime ??
+            swcConfig.jsc?.transform?.react?.runtime,
+          importSource:
+            reactRuntimeConfig.importSource ??
+            swcConfig.jsc?.transform?.react?.importSource,
```
jbroma: same as before, lets use `react` as fallback here

jbroma:
btw @dannyhw we should totally have unit tests for this and test the possible scenarios 💯

dannyhw:
tests added 👍

jbroma (approved):
LGTM 🎉

jbroma:
@dannyhw can you please add changeset entry? Run `pnpm changeset add` in the root of the repo and follow instructions. You can take a look at release notes of `@callstack/repack` to see how the descriptions should roughly look like

dannyhw:
On it 👍

PR #1317, opened 2025-12-09
docs: update docs on nativewind setup to be clearer

dannyhw:
### Summary

Removes the note about postcss config in troubleshooting since its not needed and causes confusion.
Updates some nativewind links
Removes references to manually setting import source since the nativewind preset handles that.

### Test plan

check the docs preview


jbroma (approved):
LGTM

PR #1318, opened 2025-12-09
docs: update rspress preset so that links are more visible

dannyhw:
### Summary

Updated the preset version so that links are more clear 

image

### Test plan

check docs preview


jbroma (approved):
LGTM 🎉

PR #1319, opened 2025-12-09
feat: expose cache entry on shouldUpdateScript

ElielC:
### Summary

Exposes a new property on `locator.shouldUpdateScript` in order to enable more custom caching behaviors on `afterResolve` hooks, following this [discussion](https://github.com/callstack/repack/discussions/1283)

### Test plan

Ensure all tests pass and there are no regression issues

### DIFF
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -505,7 +505,8 @@ export class ScriptManager extends EventEmitter {
       const fetch = await locator.shouldUpdateScript(
         scriptId,
         caller,
-        script.shouldUpdateCache(this.cache[cacheKey])
+        script.shouldUpdateCache(this.cache[cacheKey]),
+        this.cache[cacheKey]
       );
 
       // If it returns true, we need to fetch the script
--- a/packages/repack/src/modules/ScriptManager/ScriptManager.ts
+++ b/packages/repack/src/modules/ScriptManager/ScriptManager.ts
@@ -506,7 +506,7 @@ export class ScriptManager extends EventEmitter {
         scriptId,
         caller,
         script.shouldUpdateCache(this.cache[cacheKey]),
-        this.cache[cacheKey]
+        this.cache[cacheKey] ? { ...this.cache[cacheKey] } : undefined
       );
 
       // If it returns true, we need to fetch the script
--- a/website/src/latest/api/runtime/script-manager.md
+++ b/website/src/latest/api/runtime/script-manager.md
@@ -256,6 +256,43 @@ ScriptManager.shared.hooks.afterResolve(async (args) => {
 });
 ```
 
+### Override shouldUpdateScript logic
+
+```js
+ScriptManager.shared.hooks.afterResolve(async (args) => {
+  const { locator } = args;
+  locator.shouldUpdateScript = async (scriptId, caller, isScriptCacheOutdated, cachedData) => {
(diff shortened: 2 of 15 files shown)


dannyhw:
Nice one @ElielC 🙌

ElielC:
The changeset bot references the release as a Major, but i created the changeset as a Minor release, is that right?

dannyhw:
Thats strange, @jbroma is this a bug  with the changesets bot?

jbroma:
> The changeset bot references the release as a Major, but i created the changeset as a Minor release, is that right?

hey, sorry for lack of activity lately, changesets can always be adjusted manually before the release so no need to worry about that

website/src/latest/api/runtime/script-manager.md
```diff
@@ -256,6 +256,43 @@ ScriptManager.shared.hooks.afterResolve(async (args) => {
...
+  locator.shouldUpdateScript = async (scriptId, caller, isScriptCacheOutdated, cachedData) => {
+    // Custom logic to determine if the script should be updated
+    return (
+        cachedData.method !== locator.method ||
+        cachedData.url !== locator.url ||
+        cachedData.query !== locator.query ||
+        !shallowEqual(cachedData.headers, locator.headers) ||
+        cachedData.body !== locator.body
+      );
+  };
```
MikitasK, line 292: This code assumes `cachedData` always exists, but it might be `undefined` on first load. so let's guard no-cache case before dereferencing `cachedData`

```suggestion
    // Custom logic to determine if the script should be updated
    if (!cachedData) return true;
    return (
        cachedData.method !== locator.method ||
        cachedData.url !== locator.url ||
        cachedData.query !== locator.query ||
        !shallowEqual(cachedData.headers, locator.headers) ||
        cachedData.body !== locator.body
      );
  };
```

MikitasK: I’d also suggest adding a unit test covering for these cases:

- `cachedData === undefined` on first load
- `cachedData` populated on subsequent load
- changed locator making `isScriptCacheOutdated === true`

packages/repack/src/modules/ScriptManager/ScriptManager.ts
```diff
@@ -505,7 +505,8 @@ export class ScriptManager extends EventEmitter {
       const fetch = await locator.shouldUpdateScript(
         scriptId,
         caller,
-        script.shouldUpdateCache(this.cache[cacheKey])
+        script.shouldUpdateCache(this.cache[cacheKey]),
+        this.cache[cacheKey]
```
MikitasK: passing `this.cache[cacheKey]` directly into `shouldUpdateScript` implies that users receive the actual cached object, not a defensive copy. which means that Re.Pack’s internal cache entry might be affected If user's code mutates `cachedData`. so, I'd prefer to pass it in more safer way:


```suggestion
        script.shouldUpdateCache(this.cache[cacheKey]),
        this.cache[cacheKey] ? { ...this.cache[cacheKey] } : undefined
```

MikitasK (commented):
overall, these changes looks good to me 👍 also core behavior works as expected on my side ☑️  
just a few minor comments:

bartekkrok:
This looks like a solid addition!

Could you add a test case that specifically covers this scenario ?

ElielC:
Hello, sorry for taking too long and thanks for the feedback.

I will be implementing the suggested comments!

dannyhw:
@ElielC is it ready to review again?

PR #1332, opened 2026-01-26
fix: hermes-compiler path for RN 0.82+ in the hermes bytecode plugin

ismarbesic:
### Summary

The path of the `hermesc` binary was recently changed to support bundling the app using hermes v1, see https://github.com/facebook/react-native/pull/53581. This makes the HermesBytecodePlugin fail since the old path isn't present anymore in React Native 0.82+. This PR adds support for the new path while preserving a fallback to the old path.

### Test plan

- Make sure the app compiles with older react native versions. The test app can be used for this.
- Make sure the app compiles with newer react native versions. This has been tested with a RN 0.83 example app.


ismarbesic:
Thanks @jbroma @thymikee. I missed there were test cases for this. Those should be updated now and also added a test case to cover the new path and old path.

jbroma:
Thanks again for your contribution @ismarbesic 🎉

Issue #1296, opened 2025-10-08
[Bug] Rspack does not emit index.bundle file in production mode, only assets are generated


artiko00:
### Describe the bug

Complete Bug Report for Re.Pack GitHub Issue

  

  ## Description

  When building a React Native application for Android in production mode using Re.Pack 5.2.1 with Rspack 1.5.8, the bundler compiles
  successfully without errors but **fails to write the main `index.bundle` file to disk**. Only PNG assets are written to the output
  directory. This causes the build to fail when the OutputPlugin attempts to copy the non-existent bundle file, resulting in an `ENOENT`
  error.

  ## Environment

  - **Re.Pack version:** 5.2.1 (latest)
  - **Rspack version:** 1.5.8 (latest)
  - **React Native version:** 0.78.2
  - **Node version:** v23.6.1
  - **Package manager:** pnpm
  - **OS:** Linux 6.16.3-76061603-generic (Ubuntu)
  - **Module Federation:** @module-federation/enhanced 0.19.1, @module-federation/runtime 0.19.1
  - **Build command:** `npx react-native webpack-bundle --platform android --dev false --entry-file index.js --bundle-output
  build/generated/android/index.bundle`
  - **Architecture:** Monorepo with Module Federation micro-frontends

  ## Steps to Reproduce

  1. Set up a React Native 0.78.2 project with Re.Pack 5.2.1
  2. Configure `rspack.config.mjs` with Module Federation (see configuration below)
  3. Run production bundle command: `ENVFILE=.env.dev npx react-native webpack-bundle --platform android --dev false --entry-file index.js
  --bundle-output build/generated/android/index.bundle`
  4. Check output directory: `ls -la build/generated/android/`

  ## Expected Behavior

  The bundler should generate:
  - ✅ `index.bundle` - Main application bundle
  - ✅ `index.bundle.map` - Source map file
  - ✅ `*.chunk.bundle` - Code-split chunks (if applicable)
  - ✅ `drawable-*` directories with PNG assets

  ## Actual Behavior

  The bundler only generates:
  - ❌ `index.bundle` - **NOT CREATED**
  - ❌ `index.bundle.map` - **NOT CREATED**
  - ❌ `*.chunk.bundle` - **NOT FOUND in output directory**
  - ✅ `drawable-hdpi/`, `drawable-mdpi/`, `drawable-xhdpi/`, `drawable-xxhdpi/`, `drawable-xxxhdpi/` - EMPTY directories created

  **Console output:**
[...]

PR #1333, opened 2026-01-26
fix: assert compilation errors in OutputPlugin 

OrfeasZ:
Previously, the `RepackOutputPlugin` would always assume compilation had succeeded. The `done` hook gets fired even after compilation finishes with errors, which would cause the plugin to try to copy around chunks that haven't been written to disk (due to said errors), resulting in cryptic errors about `index.bundle` not being found. This change adds a check for compilation errors, and aborts the bundling process if it finds any.

This likely "fixes" #1296 (or at least makes what's happening clearer).

### DIFF
--- a/packages/repack/src/plugins/__tests__/OutputPlugin.test.ts
+++ b/packages/repack/src/plugins/__tests__/OutputPlugin.test.ts
@@ -28,7 +28,7 @@ async function testCompile(code: string, expectThrow?: string) {
       entry: 'index.js',
       output: {
         filename: 'index.bundle',
-        path: '/out',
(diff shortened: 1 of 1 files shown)


jbroma:
@OrfeasZ please also fix the existing lint / test issues as well 👍

OrfeasZ:
@jbroma Done!

OrfeasZ:
Oops, looks like I missed something and it was passing locally because of some stale files. Should be fixed now.

mikeduminy:
This works super well for us as it uncovers the actual cause of bundle failures instead of cryptic logs. Would love it to get merged 🙏

PR #1376, opened 2026-03-20
feat(android): allow custom OkHttpClient via factory

whydidoo:
Replace direct OkHttpClient instantiation with a configurablefactory in companion object, enabling custom client configuration(SSL, interceptors, timeouts) from the host application.

### Summary

Currently, `RemoteScriptLoader` creates a hardcoded `OkHttpClient()` instance with no way to customize it. This is a problem for apps that need custom HTTP configuration — for example, certificate pinning, custom SSL trust managers, logging interceptors, or proxy settings.

This PR introduces a static `okHttpClientFactory` lambda in a `companion object`, allowing host applications to provide their own `OkHttpClient` instance before any network requests are made:

```kotlin
RemoteScriptLoader.okHttpClientFactory = {
    OkHttpClient.Builder()
        .addInterceptor(loggingInterceptor)
        .sslSocketFactory(sslSocketFactory, trustManager)
        .connectTimeout(30, TimeUnit.SECONDS)
        .build()
}
```

The client is lazily initialized via by `lazy`, so the factory is only called once on the first use.

Test plan

1. **Default behavior** — verify that without setting `okHttpClientFactory`, a default `OkHttpClient()` is used (no regression).
2. **Custom client** — set `okHttpClientFactory` in `Application.onCreate()` or a React Native module initializer with a custom interceptor, confirm the interceptor is invoked during remote script loading.
3. **Lazy initialization** — confirm the factory is called only once, regardless of how many scripts are loaded.


bartekkrok:
Thanks for the contribution! A couple of thoughts:
- Could you add some documentation for this feature? 
- Looking at the iOS side, ScriptManager.mm has the same limitation. Would you be open to adding an equivalent mechanism for iOS?

whydidoo:
> Thanks for the contribution! A couple of thoughts:
> 
> * Could you add some documentation for this feature?
> * Looking at the iOS side, ScriptManager.mm has the same limitation. Would you be open to adding an equivalent mechanism for iOS?

@bartekkrok Thanks for the replay.

I will add documentation for this feature.

As for iOS, can I do it in this PR?

bartekkrok:
> > Thanks for the contribution! A couple of thoughts:
> > 
> > * Could you add some documentation for this feature?
> > * Looking at the iOS side, ScriptManager.mm has the same limitation. Would you be open to adding an equivalent mechanism for iOS?
> 
> @bartekkrok Thanks for the replay.
> 
> I will add documentation for this feature.
> 
> As for iOS, can I do it in this PR?

You can create separate for ios or separate including both platforms together.

dannyhw:
@whydidoo hey, just bumping here. What are your thoughts on providing the ios solution? are you able to add this functionality? Whats your thoughts on how we should move forward?

whydidoo:
@dannyhw Hey, sorry for the delay — I’ve been pretty swamped with work. I’ll take a closer look at the iOS solution and start working on it this weekend.

whydidoo:
@dannyhw and @bartekkrok, sorry for the delay — I finally found some time to finish this PR.

dannyhw:
thanks @whydidoo I will take some time to try this myself

dannyhw:
tested it, seems good to me

dannyhw (approved):
thanks 🙏

PR #1378, opened 2026-04-07
feat: add maxInlineSize option for size-based asset inlining

bartekkrok:
### Summary

Currently, the assets loader only supports two modes for asset delivery: always
extract as separate files or always inline as base64 URIs. There is no way to       
automatically inline small assets (where the base64 overhead is negligible) while
still extracting large ones — developers had to choose one strategy for all assets  
or manually split rules by path.

This PR adds a maxInlineSize option (in bytes) to both assets-loader and            
getAssetTransformRules. Assets whose largest scale variant is within the threshold
are inlined as base64 URIs; larger assets are extracted as separate files. The      
comparison uses the largest scale variant (e.g. @3x) intentionally — when an asset
is inlined, all variants are embedded in the bundle, so the largest one determines
the worst-case size impact.

### Test plan

1. Add an asset smaller than the threshold and one larger than the threshold to a   
  React Native screen.
2. Configure the assets loader with maxInlineSize: <threshold in bytes>.            
3. Build the app — verify the small asset is inlined as a base64 URI and the large
asset is extracted as a separate file.                                              
4. Alternatively, use the tester app (apps/tester-app) which already has assets set
up for manual validation.                                                           
5. Verify the option works both when passed directly to assets-loader and via
getAssetTransformRules({ maxInlineSize: ... }).


packages/repack/src/loaders/assetsLoader/assetsLoader.ts
```diff
@@ -222,8 +222,16 @@ export default async function repackAssetsLoader(
       })
     );
 
+    const largestVariantSize = Math.max(
+      ...assets.map((asset) => asset.data.length)
+    );
+    const shouldInlineAsset =
+      options.inline === true ||
+      (options.maxInlineSize !== undefined &&
+        largestVariantSize <= options.maxInlineSize);
```
MikitasK: nit: 
maybe boolean type conversion might make it a bit more readable
```suggestion
    const shouldInlineAsset =
      !!options.inline ||
      (!!options.maxInlineSize &&
        largestVariantSize <= options.maxInlineSize);
```

packages/repack/src/loaders/assetsLoader/assetsLoader.ts
MikitasK, line 1: there's no automated coverage for new `maxInlineSize`prop - nothing exercises new threshold behavior in either `assets-loader.test.ts` or `getAssetTransformRules.test.ts`. it'd be great cover this new logic with tests

MikitasK (commented):
overall, these changes looks good to me 👍 
new `maxInlineSize` prop works like a charm 🚀 here's the recording:

https://github.com/user-attachments/assets/cb3535b8-0c92-45fb-9ea4-026e9e98d6ff

I'd just prefer to take into account a few things before merge:

packages/repack/src/loaders/assetsLoader/assetsLoader.ts
bartekkrok: Sure thing ! I refered to Your comments in new commit.

MikitasK (approved):
LGTM! :+1:

dannyhw:
@bartekkrok could you please resolve the linting/test ci failure and then I think we can look to merge this

packages/repack/src/loaders/assetsLoader/assetsLoader.ts
```diff
@@ -222,8 +222,15 @@ export default async function repackAssetsLoader(
       })
     );
 
+    const largestVariantSize = Math.max(
+      ...assets.map((asset) => asset.data.length)
+    );
+    const shouldInlineAsset =
```
dannyhw, line 228: I think it would be more intuitive if it was something like

`!!inline && (!options.maxInlineSize || largestVariantSize <= options.maxInlineSize)`

So that if inline=false then its disabled and then maxinline setting requires inline=true

bartekkrok: Yeah, that makes sense, I'll add changes

dannyhw (approved):

PR #1379, opened 2026-04-11
fix(CodeSigningPlugin): sign assets at processAssets ANALYSE stage before REPORT

### LINKED ISSUE #1377: CodeSigningPlugin signs too late (assetEmitted) – incompatible with in-memory asset consumers (e.g. Zephyr)
### Describe the bug

# 🐛 CodeSigningPlugin signs too late (`assetEmitted`) – incompatible with in-memory asset consumers (e.g. Zephyr)

## Problem

`CodeSigningPlugin` currently signs bundles using `compiler.hooks.assetEmitted`.

When using `withZephyr()`, Zephyr captures and uploads assets during the `processAssets` phase at `PROCESS_ASSETS_STAGE_REPORT` (5000), which occurs **before** `assetEmitted`.

As a result, bundles are uploaded to the CDN **unsigned**, making `verifyScriptSignature: 'strict'` ineffective when using Re.Pack together with Zephyr.

---

## Root Cause

There is a timing mismatch in the compilation lifecycle:

```
processAssets (ANALYSE 2000)   ← no signing happens here
processAssets (REPORT 5000)    ← Zephyr uploads assets (unsigned ✗)
emit                           ← assets written to disk
assetEmitted                   ← CodeSigningPlugin signs (too late)
```

Zephyr registers its upload step at `PROCESS_ASSETS_STAGE_REPORT`:

```ts
compilation.hooks.processAssets.tapPromise(
  {
    name: pluginName,
    stage: compiler.webpack.Compilation.PROCESS_ASSETS_STAGE_REPORT, // 5000
  },
  async (assets) => {
    // uploads to CDN
  }
);
```

---

## Expected Behavior

Assets should be signed **before** they are consumed by other plugins/tools during the compilation pipeline.

---

## Proposed Solution

Move the signing step earlier in the lifecycle by using `processAssets`, specifically at `PROCESS_ASSETS_STAGE_ANALYSE` (2000), while assets are still in memory.

```
processAssets (ANALYSE 2000)   ← sign here in memory ✓
processAssets (REPORT 5000)    ← Zephyr uploads signed assets ✓
emit                           ← assets written to disk (already signed) ✓
```

---

## Suggested Implementation

Instead of signing in `assetEmitted`, perform signing during `processAssets` and update assets in memory using `compilation.updateAsset()`:

```ts
compiler.hooks.thisCompilation.tap('RepackCodeSigningPlugin', (compilation) => {
  const { Compilation, sources } = compiler.webpack;

  compilation.hooks.processAssets.tap(
    {
      name: 'RepackCodeSigningPlugin',
      stage: Compilation.PROCESS_ASSETS_STAGE_ANALYSE,
    },
    () => {
      for (const chunk of compilation.chunks) {
        for (const file of chunk.files) {
          // apply existing signing logic here
          compilation.updateAsset(file, new sources.RawSource(signed));
        }
      }
    }
  );
});
```

---

## Compatibility Consideration

To preserve backward compatibility, this could be introduced as a configurable option:

```ts
type CodeSigningPluginConfig = {
  signingStage?: 'assetEmitted' | 'processAssets';
};
```

- `assetEmitted` → current default behavior
- `processAssets` → enables in-memory signing (recommended for Zephyr)

---

## Verified

This approach has been tested using a custom plugin:

- Bundles are signed in memory before `PROCESS_ASSETS_STAGE_REPORT`
- Zephyr uploads already signed assets
- Output files on disk remain correctly signed
- `verifyScriptSignature: 'strict'` works as expected end-to-end

---

## Environment

- `@callstack/repack`: [version]
- Zephyr integration: [version]
- Platform: Android / iOS

### System Info

```shell
macOS (Apple Silicon)
Node: 22.19.0
Pnpm: 9.15.3
React Native: 0.78.2
Re.Pack: 5.2.5
Rspack: 1.3.4

Build type: Release
Platform: Android / iOS
```

### Re.Pack Version

5.2.5

### Reproduction

Not a minimal reproduction repository, but the issue can be consistently reproduced when using CodeSigningPlugin together with Zephyr integration (withZephyr()).  Happy to provide a minimal reproduction if needed.

### Steps to reproduce

1. Configure Re.Pack with CodeSigningPlugin enabled
2. Integrate Zephyr using withZephyr()
3. Build the project (e.g. release bundle)
4. Observe that Zephyr uploads assets during processAssets
5. Inspect uploaded bundle → it is not signed
6. Inspect output on disk → bundle is signed

Result:
Assets uploaded to CDN are unsigned, while local output is signed.

JhohellsDL:
## Summary

Fixes #1377

`CodeSigningPlugin` was signing bundles in `compiler.hooks.assetEmitted`, 
which fires after `processAssets` completes. When using `withZephyr()`, 
Zephyr captures and uploads assets at `PROCESS_ASSETS_STAGE_REPORT` (5000) 
— before `assetEmitted` fires — resulting in unsigned bundles being uploaded 
to the CDN, making `verifyScriptSignature: 'strict'` ineffective.

## Changes

- Moved signing logic from `assetEmitted` to `processAssets` at 
  `PROCESS_ASSETS_STAGE_ANALYSE` (2000), before Zephyr's `REPORT` stage (5000)
- Assets are now signed in memory via `compilation.updateAsset()` 
  instead of reading/writing from disk
- Removed `chunkFilenames` Set and `emit` hook — no longer needed since 
  signing iterates `compilation.chunks` directly inside `processAssets`
- Added test verifying assets are signed before `REPORT` stage
- Updated documentation with `## Behavior` section explaining the signing stage

## Testing

- All existing tests pass
- Added new test simulating a plugin at `REPORT` stage confirming 
  assets are already signed when captured
- Verified end-to-end in production with `withZephyr()` — bundles 
  uploaded to CDN now contain the signature and `verifyScriptSignature: 'strict'` 
  works correctly


dannyhw:
Is there some simple way to test this locally? What would you recommend? @JhohellsDL

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -76,40 +67,65 @@ export class CodeSigningPlugin {
...
-      async (file, { outputPath, compilation }) => {
-        const outputFilepath = path.join(outputPath, file);
-        const readFileAsync = util.promisify(
-          compiler.outputFileSystem!.readFile
-        );
-        const content = (await readFileAsync(outputFilepath)) as Buffer;
+    compiler.hooks.thisCompilation.tap(
+      'RepackCodeSigningPlugin',
+      (compilation) => {
+        // @ts-ignore — sources is available on both rspack and webpack compilers
```
MikitasK: is it possible to get rid of `@ts-ignore` here?

packages/repack/src/plugins/__tests__/CodeSigningPlugin.test.ts
```diff
@@ -81,6 +83,59 @@ describe('CodeSigningPlugin', () => {
     expect(chunkBundle.length).toBeGreaterThan(1280);
   });
 
+  it('exposes signed chunk assets to processAssets REPORT (after ANALYSE signing)', async () => {
```
MikitasK, line 86: this test proves assets are signed by REPORT, which is good 👍 but what about one more test for asserting the actual delta from old behavior?  it might be done by checking that content at REPORT is larger than unsigned chunk or by detecting `/* RCSSB */` substring in content
wdyt?

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -76,40 +67,65 @@ export class CodeSigningPlugin {
...
+                  .digest('hex');
+                /** generate token */
+                const token = jwt.sign({ hash }, privateKey, {
+                  algorithm: 'RS256',
+                });
+                /** combine the bundle and the token */
+                const signedBundle = Buffer.concat(
+                  [content, Buffer.from(BEGIN_CS_MARK), Buffer.from(token)],
+                  content.length + TOKEN_BUFFER_SIZE
+                );
```
MikitasK: [nit]
what about making this logic a bit more readable by moving this code into dedicated `signAsset()` helper?
```suggestion
const signedBundle = this.signAsset(file, asset, privateKey, BEGIN_CS_MARK, TOKEN_BUFFER_SIZE);
```

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -76,40 +67,65 @@ export class CodeSigningPlugin {
...
 
-        const writeFileAsync = util.promisify(
-          compiler.outputFileSystem!.writeFile
+        compilation.hooks.processAssets.tap(
+          {
+            name: 'RepackCodeSigningPlugin',
+            // Sign at ANALYSE (2000) so assets are signed before any plugin
+            // running at REPORT (5000) — e.g. withZephyr() — captures them.
+            // The original assetEmitted hook fires after processAssets completes,
+            // which is too late when Zephyr uploads assets at REPORT stage.
```
MikitasK: [nit] this comment might be a bit shorter
```suggestion
// Sign at ANALYSE (2000) so later processAssets consumers, 
// such as Zephyr at REPORT (5000), receive already-signed assets
```

MikitasK (commented):
the PR is pretty solid 👍 

I tested it locally with `apps/tester-app` using temporary `processAssets` REPORT-stage capture plugin & the results satisfied my expectations: 

1. capture report showed all remote chunks signed at REPORT stage (they're showed unsigned on `main` branch tho)
2.  remote chunk loads successfully when `verifyScriptSignature: 'strict'` & public key embedded in the app

https://github.com/user-attachments/assets/12569489-dbeb-4b6b-bf45-391006aa5454

can you just consider a few suggestions before merge:

packages/repack/src/plugins/__tests__/CodeSigningPlugin.test.ts
JhohellsDL: Good call. I extended the test so it doesn’t only assert the JWT-shaped tail at REPORT.

There is now a processAssets tap at one stage before ANALYSE (same chunk iteration as REPORT) to snapshot the asset before CodeSigningPlugin runs. Then at REPORT we assert:

the pre-sign snapshot has no /* RCSSB */ marker
the REPORT snapshot includes /* RCSSB */
REPORT content is strictly longer than the pre-sign snapshot
The existing regex checks on the chunk vs main bundle are unchanged. That should document the regression we care about: anything observing assets at REPORT must see the already signed bytes, not only the final emitted file.

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
JhohellsDL: Addressed the nit: extracted the hash/JWT/buffer concat path into a private signAsset() helper so the processAssets callback stays easier to scan. Behavior unchanged; CodeSigningPlugin tests still pass.

MikitasK (approved):
LGTM 🙌

dannyhw:
@JhohellsDL could you please take a look and resolve the linting errors

JhohellsDL:
@dannyhw, thanks for the feedback! I’ve addressed the linting issues and pushed an update. Please take another look.

website/src/v4/docs/plugins/code-signing.md
jbroma, line 1: v4 docs are read-only, should be removed

website/src/latest/api/plugins/code-signing.md
jbroma, line 1: nit: either add this to details section or skip altogether

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -1,24 +1,15 @@
...
 import { type CodeSigningPluginConfig, validateConfig } from './config.js';
 
 export class CodeSigningPlugin {
-  private chunkFilenames: Set<string>;
-
-  /**
-   * Constructs new `RepackPlugin`.
-   *
-   * @param config Plugin configuration options.
-   */
```
jbroma, line 16: nit: keep it

jbroma (approved):
few nits, LGTM overall 👍

dannyhw:
@JhohellsDL if you can address those last comments I think we can move forward and merge this 👍

JhohellsDL:
@dannyhw, Ready, I’ve addressed the comments. I’ll keep an eye out for any further feedback 👍

dannyhw:
Thanks @JhohellsDL, appreciate your responsiveness and all your hard work 🙇‍♂️

dannyhw (approved):
thanks for your contribution 🙇‍♂️

Issue #1247, opened 2025-08-07
HMR not working in Android Physical Device


AbrahamBaby:
### Describe the bug

Hi @jbroma 
 I have observed that HMR is not working as expected in android phyiscal device whereas it works fine on android emulator and IOS simulator and IOS physical device, is this a know issue or some configuration issue, please let me know if can be resolved by changing some configuration

### System Info

```shell
Macos
Ract native - 0.79.3
Repack - 5.1.2
```

### Re.Pack Version

5.1.2

### Reproduction

-

### Steps to reproduce

Run yarn install and open android physical device change something in the screen and save it, you will not be able to see any change since HMR is not working, once you kill and rerun the app, we will be able to see the change got reflected, these steps will be able to reproduce the issue

jbroma:
hi @AbrahamBaby 

can you try running the `start` command with your actual network IP address? By default we use `localhost`, but for physical device, you should use your address on the network. Please check this discussion and see whether it fixes your issue: https://github.com/callstack/repack/discussions/1166#discussioncomment-13424428

AbrahamBaby:
Hi @jbroma , 
Thank you for your quick response, actually i have already tried starting the rspack server and added —host for running the android device with my n/w ip, also added it in the rspack config where we are initializing the server with local host, like
shopping: ‘shopping@http://localhost:8081{mfe-path}
Here instead of local host i have updated with my ip address, also i have updated the debug port settings in android dev menu, still not able to make the HMR working, is there anything else i can do to make it work, if anything is there, please let me know, Thanks in advance

MikitasK:
hi @AbrahamBaby 👋 

seems like the root cause is that re.pack rewrites Android `localhost` to `10.0.2.2` [here](https://github.com/callstack/repack/blob/13e7ed58020ae7ad9ed62b2f8998a2035a80b5a4/packages/repack/src/modules/getDevServerLocation.ts#L4). that's correct for emulators, but wrong for physical devices

that's why HMR websocket resolves to `ws://10.0.2.2:8081/__hmr` & never connects on physical device with the default `localhost`. I resolved that in [this](https://github.com/callstack/repack/issues/1247) PR. but just in case this is the workaround that you can use right away:

1. start the dev server with an explicit LAN IP:
2. `pnpm react-native webpack-start --host <mac-ip>`
3. set the Android dev server host to `<mac-ip>:8081`

[...]

PR #1380, opened 2026-04-16
fix(android): stop rewriting localhost to 10.0.2.2

MikitasK:
### Summary

Issue #1247

Fix Android physical-device dev server resolution by removing the unconditional runtime rewrite from `localhost` to `10.0.2.2`

Previously, re.pack treated every Android runtime like an emulator. That caused:
- HMR websocket URLs to resolve to `ws://10.0.2.2:8081/__hmr`
- async chunk URLs to resolve to `http://10.0.2.2:8081/...`

It works on emulators, but breaks on physical devices, where `10.0.2.2` is unreachable

This change makes re.pack use the configured dev-server host literally, which allows Android physical devices to work correctly with the standard `localhost` + `adb reverse` flow

### Test Plan

1. connect an Android physical device with USB debugging enabled
2. start the dev server from `apps/tester-app` with the default host by `pnpm react-native webpack-start`
3. run `adb reverse tcp:8081 tcp:8081`
4. launch "Tester App" on the Android physical device
5. confirm the app loads successfully
6. edit visible UI file such as `apps/tester-app/src/App.tsx` & save
7. confirm HMR works - the UI should updates without full app relaunch

### Recordiing

https://github.com/user-attachments/assets/773552e8-92c4-45bc-81a7-0db4d015b6f5

### DIFF
--- a/.changeset/twenty-bushes-act.md
+++ b/.changeset/twenty-bushes-act.md
@@ -1,5 +1,5 @@
 ---
-"@callstack/repack": major
+"@callstack/repack": patch
 ---
 
 Fix Android dev-server URLs on physical devices by preserving `localhost` instead of rewriting it to `10.0.2.2`.
--- a/.changeset/twenty-bushes-act.md
+++ b/.changeset/twenty-bushes-act.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": major
+---
+
+Fix Android dev-server URLs on physical devices by preserving `localhost` instead of rewriting it to `10.0.2.2`.
--- a/packages/repack/src/modules/getDevServerLocation.ts
+++ b/packages/repack/src/modules/getDevServerLocation.ts
@@ -1,14 +1,8 @@
-let hostname = __PUBLIC_HOST__;
-
-if (__PLATFORM__ === 'android' && __PUBLIC_HOST__ === 'localhost') {
-  hostname = '10.0.2.2';
-}
-
 const location = {
-  host: `${hostname}:${__PUBLIC_PORT__}`,
-  hostname,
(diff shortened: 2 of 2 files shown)


dannyhw:
What a find! Nice job!

dannyhw (approved):

dannyhw:
@MikitasK can you please add a changeset 🙏

.changeset/twenty-bushes-act.md
```diff
@@ -0,0 +1,5 @@
+---
+"@callstack/repack": major
+---
+
+Fix Android dev-server URLs on physical devices by preserving `localhost` instead of rewriting it to `10.0.2.2`.
```
dannyhw, line 5: I don't think we want a major for this, patch is probably fine for this type of fix

MikitasK: sure thing 👍  thanks for catching this up 🙏 
fixed ✅

dannyhw:
@MikitasK could you double check on emulator that it doesn't break. Like if you fully create a fresh build etc just to make sure

MikitasK:
> @MikitasK could you double check on emulator that it doesn't break. Like if you fully create a fresh build etc just to make sure

sure thing @dannyhw 👌 double-checking these changes before merge makes a lot of sense 
here is the recording of how it look like on my side from clean repo state on wiped emulator:

https://github.com/user-attachments/assets/a99b47c8-4a46-4db0-9782-c579b89b8961

the behavior is pretty much the same - it works as expected ☑️

bartekkrok (approved):

dannyhw:
Thanks for double checking, looks like this has been confirmed as working 👍

Issue #1323, opened 2025-12-30
Parse and automate setting to files like plist and xml


dougg0k:
### Description

Hey,

Why not parse files like plist and xml, check and add the data needed in them.

### Suggested solution

Here you suggest adding public keys in them. https://re-pack.dev/api/plugins/code-signing#add-the-public-key

Alternatively, the dev could specify the public key in the setting https://re-pack.dev/api/plugins/code-signing#add-the-plugin

Automatically it would be added to the files as needed.

### Additional context

- https://github.com/mat-sz/plist
- https://github.com/expo/expo/tree/main/packages/expo/plist
- https://github.com/NaturalIntelligence/fast-xml-parser

There is even gradle api/plugin and ios native api to programmatically set / modify.

dannyhw:
Available in latest canary

PR #1381, opened 2026-04-17
feat(CodeSigningPlugin): auto-embed public key into native project files

bartekkrok:
Linked to: #1323                                    
  ### Summary                                                                                                            
                                                                                                                         
  Setting up code signing with `CodeSigningPlugin` required users to manually add the RSA public key to native project   
  files (`Info.plist` on iOS and `strings.xml` on Android). This is error-prone, easy to forget, and creates a           
  maintenance burden every time the key is rotated.                                                                      
                  
  This PR adds two new optional config options — `publicKeyPath` and `nativeProjectPaths` — that automate public key     
  embedding during the build:
                                                                                                                         
  ```ts           
  new CodeSigningPlugin({
    privateKeyPath: './keys/private.pem',
    publicKeyPath: './keys/public.pem',      // new                                                                      
    nativeProjectPaths: {                    // new, optional overrides
      ios: './ios/MyApp/Info.plist',                                                                                     
      android: './android/app/src/main/res/values/strings.xml',
    },                                                                                                                   
  });
```
                                                                                                                         
  When publicKeyPath is set the plugin:
  - Auto-detects ios/<AppDir>/Info.plist and android/app/src/main/res/values/strings.xml (standard RN project layout)
  - Inserts or updates the RepackPublicKey entry in each file on every build                                             
  - Creates strings.xml if it doesn't exist yet                             
  - Accepts nativeProjectPaths to override either path for non-standard layouts                                          
  - Logs a warning and skips gracefully if the key file or target files are not found                                    
                                                                                                                         
  The embedding logic lives in the new embedPublicKey.ts module, also exported publicly for standalone use.              
                                                                                                                         
  Both platforms are idempotent — re-running the build updates the existing entry rather than duplicating it.            
                                                                                                                         
  ### Test plan                                                                                                              
                  
  - Add publicKeyPath pointing to a valid public key in CodeSigningPlugin config                                         
  - Run a build and verify RepackPublicKey entry is inserted into ios/<AppDir>/Info.plist
  - Run a build and verify RepackPublicKey entry is inserted into android/app/src/main/res/values/strings.xml            
  - Run the build again and confirm the entry is updated, not duplicated                                                 
  - Remove strings.xml, run a build, and confirm the file is created with the key                                        
  - Set publicKeyPath to a non-existent file and confirm a warning is logged, build continues                            
  - Use nativeProjectPaths to override paths and confirm the plugin writes to the specified files    
  
  ### Live demo iOS
  
  Changes were tested on: https://github.com/callstack/super-app-showcase

https://github.com/user-attachments/assets/e0df3460-0be5-4ca5-add9-ce9cbc62107a

### CODE REVIEW

File: packages/repack/src/plugins/CodeSigningPlugin/embedPublicKey.ts [hunk 1]
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  const stringsXmlPath =
+    config.androidStringsXmlPath ??
+    findAndroidStringsXmlPath(config.projectRoot);
+
+  if (stringsXmlPath) {
+    try {
+      embedPublicKeyInStringsXml(publicKey, stringsXmlPath);
```
MikitasK: based on this logic the the `embedPublicKeyInStringsXml()` invoked only when `stringsXmlPath` is truthy, but `findAndroidStringsXmlPath()` might return `null`. 
so, as far as I understand, the app with no preexisting `strings.xml` silently skips Android embedding instead of creating the file, and strict verification later fails at runtime. 
overall, I'm ok with such logic, but it'd be much better to notify user that auto-embedding skipped

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -51,6 +110,8 @@ export class CodeSigningPlugin {
       return;
     }
 
+    this.embedPublicKeyInNativeProjects(compiler);
```
MikitasK: as I see, the plugin mutates native source files before it finishes validating the build configuration. `apply()` calls `embedPublicKeyInNativeProjects()` before the unsupported dynamic filename check and before reading the private key. If the build then throws because `output.filename` is dynamic or `privateKeyPath` is bad, the compile fails but `Info.plist` / `strings.xml` may already be modified, leaving the worktree dirty after an unsuccessful build 

that's why, I suppose, those side effects should happen only after all throw-on-startup validation has passed
wdyt?

MikitasK (commented):
very solid work done 👍👍 
I verified both Android & iOS platforms & auto-embeding works perfectly fine 🚀 

**iOS**

https://github.com/user-attachments/assets/2b82983d-4192-4c84-ad3c-978fd0c91cba

**Android**

https://github.com/user-attachments/assets/4ecaf3c4-9a72-42d9-8ff1-d59762879a54

here are just a few things to consider before merge:

MikitasK (approved):
LGTM ✅

dannyhw:
@bartekkrok  please check the linting 🙏

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -39,6 +39,68 @@ export class CodeSigningPlugin {
...
+          'Skipping automatic embedding into native project files.'
+      );
+      return;
+    }
+
+    const result = embedPublicKey({
+      publicKeyPath,
+      projectRoot,
+      iosInfoPlistPath: this.config.nativeProjectPaths?.ios
+        ? path.isAbsolute(this.config.nativeProjectPaths.ios)
```
dannyhw: nitpick: maybe we could adjust some of this to avoid many nested ternaries

maybe a getInfoPlistPath function?

packages/repack/src/plugins/CodeSigningPlugin/embedPublicKey.ts
```diff
@@ -0,0 +1,200 @@
...
+  android: { modified: boolean; path?: string; error?: string };
+}
+
+/**
+ * Embeds the Re.Pack code-signing public key into native project files.
+ * Modifies `Info.plist` (iOS) and `strings.xml` (Android) so the runtime
+ * can verify signed bundles without manual file editing.
+ */
+export function embedPublicKey(config: EmbedPublicKeyConfig): EmbedPublicKeyResult {
+  const publicKey = fs.readFileSync(config.publicKeyPath, 'utf-8').trim();
```
dannyhw: any possibility this readFile could error? and would that cause any issue if this throws?

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
bartekkrok: good catch 👍

packages/repack/src/plugins/CodeSigningPlugin/config.ts
```diff
@@ -10,6 +10,24 @@ export interface CodeSigningPluginConfig {
...
+   * strings.xml for Android) so that the runtime can verify signed bundles.
+   *
+   * Relative paths are resolved from the project root (compiler context).
+   */
+  publicKeyPath?: string;
+  /**
+   * Override auto-detected paths to native project files where the public key
+   * should be embedded. Only used when `publicKeyPath` is set.
+   */
+  nativeProjectPaths?: {
```
dannyhw, line 25: I wonder if we should have a way for users to disable this behaviour? Like if the option was something like

`automaticallyEmbedPublicKey: { ios, android } | false | true  // default true`

or is it not worth having a way to disable it?

dannyhw:
@bartekkrok can you check the conflicts here?

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -39,6 +49,72 @@ export class CodeSigningPlugin {
...
+      return;
+    }
+
+    const logger = compiler.getInfrastructureLogger('RepackCodeSigningPlugin');
+    const projectRoot = compiler.context;
+
+    const publicKeyPath = resolveProjectPath(
+      projectRoot,
+      this.config.publicKeyPath
+    )!;
```
dannyhw: this `!` seems like it would be better to avoid, perhaps we can just handle the undefined case

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -62,15 +138,27 @@ export class CodeSigningPlugin {
...
     const BEGIN_CS_MARK = '/* RCSSB */';
 
-    const privateKeyPath = path.isAbsolute(this.config.privateKeyPath)
-      ? this.config.privateKeyPath
-      : path.resolve(compiler.context, this.config.privateKeyPath);
-    const privateKey = fs.readFileSync(privateKeyPath);
+    const privateKeyPath = resolveProjectPath(
+      compiler.context,
+      this.config.privateKeyPath
+    )!;
```
dannyhw: same thing about the `!` here

packages/repack/src/plugins/CodeSigningPlugin/embedPublicKey.ts
```diff
@@ -0,0 +1,218 @@
...
+ * Embeds or updates `RepackPublicKey` in an iOS `Info.plist` file.
+ */
+export function embedPublicKeyInPlist(
+  publicKey: string,
+  plistPath: string
+): void {
+  let content = fs.readFileSync(plistPath, 'utf-8');
+
+  const existingKeyPattern =
+    /[ \t]*<key>RepackPublicKey<\/key>\s*<string>[\s\S]*?<\/string>/;
```
dannyhw, line 145: I wonder if regex is the right move here

`fast-xml-parser` and `plist` or `@plist/parse` / `@plist/plist`

could probably be used to be more accurate/robust.

what do you think?

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
```diff
@@ -39,6 +49,72 @@ export class CodeSigningPlugin {
...
+
+    if (!fs.existsSync(publicKeyPath)) {
+      logger.warn(
+        `Public key not found at ${publicKeyPath}. ` +
+          'Skipping automatic embedding into native project files.'
+      );
+      return;
+    }
+
+    const result = embedPublicKey({
```
dannyhw, line 92: I wonder if we would want to think of a way to not always try to embed, like some kind of cache or something if we know it was already embedded? Maybe not worth the effort though, what do you think?

packages/repack/src/plugins/CodeSigningPlugin/embedPublicKey.ts
dannyhw: I guess the tradeoff would be plist parsing could be slower

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
bartekkrok: ok, will be better to read through

packages/repack/src/plugins/CodeSigningPlugin/embedPublicKey.ts
bartekkrok: Worth keeping as-is. The patterns are narrow and fully controlled (we own the key name RepackPublicKey).
It  would introduce a build-time dependency and more code without meaningful pros

packages/repack/src/plugins/CodeSigningPlugin/CodeSigningPlugin.ts
bartekkrok: Not big effort, I added changes

dannyhw (approved):

PR #1382, opened 2026-04-21
feat(CodeSigningPlugin): multiple public keys support

### LINKED ISSUE #826: Support bundle verification with multiple public keys
### Description

The situation we're facing is as follows:
We have multiple independent MiniApp development teams, and we want each team to use their own unique private key.
<p align="center" width="100%">

</p>

This means that when verifying a bundle, we need to pass in the public key of the respective team (which is fetched from the server).

<p align="center" width="100%">

</p>

However, when examining the repack source code, we found that the 'verify' function is currently **ONLY**  allowing for a single public key that is hardcoded into the source code, as shown below."
```
fun verifyBundle(context: Context, token: String?, fileContent: ByteArray?) {
            if (token == null) {
                throw Exception("The bundle verification failed because no token for the bundle was found.")
            }

            val stringPublicKey = getPublicKeyFromStringsIfExist(context)
                    ?: throw Exception("The bundle verification failed because PublicKey was not found in the bundle. Make sure you've added the PublicKey to the res/values/strings.xml under RepackPublicKey key.")

            val publicKey = parsePublicKey(stringPublicKey)
                    ?: throw Exception("The bundle verification failed because the PublicKey is invalid.")

            val claims: Map<String, Any?> = verifyAndDecodeToken(token, publicKey)

            val contentHash = claims["hash"] as String?
                    ?: throw Exception("The bundle verification failed because the token is invalid.")

            val fileHash = computeHash(fileContent)

            if (contentHash != fileHash) {
                throw Exception("The bundle verification failed because the bundle hash is invalid.")
            }
        }

```
```
 private fun getPublicKeyFromStringsIfExist(
                context: Context
        ): String? {
            val packageName: String = context.packageName
            val resId: Int =
                    context.resources.getIdentifier("RepackPublicKey", "string", packageName)
            if (resId != 0) {
                return context.getString(resId).ifEmpty {
                    null
                }
            }
            return null
        }
```

### Suggested solution

It would be much better if the verify bundle function could accept an additional public key as a parameter, instead of hardcoding it as it is now, such as:
```
fun verifyBundle(context: Context, 
                         token: String?, 
                         fileContent: ByteArray?,
                         publicKeyAsByteArray: ByteArray? // add this parameter
) {
  val publicKey = publicKeyFromByteArray(publicKeyAsByteArray)
// 1. verify bundle by public key
// 2. check sum bundle
```
With this approach, we can have multiple mini-app development teams, each with their own private key for signing their mini-apps. We can authenticate development teams using their keys, allowing them to join the ecosystem or rejecting teams without relying on the main app's private key.
Furthermore, if the main app's private key is lost or compromised, mini-apps created by various development teams can still operate. We have the flexibility to re-sign them whenever needed,(similar Google's Play App Signing approach)

Would you please consider this proposal

Thanks & BestRegard,

### Additional context

_No response_

MikitasK:
### Summary

Closes [#826](https://github.com/callstack/repack/issues/826)

The PR adds support for passing a per-script `publicKey` through `ScriptManager` resolver output, so signed bundles can be verified with a runtime-provided public key instead of only the app-embedded `RepackPublicKey`.

This solves the multi-team / multi-miniapp use case described in the issue:

- [x] - different teams can sign bundles with different private keys
- [x] - the host app can fetch the matching public key for a given script from its backend
- [x] - Re.Pack verifies that script with the provided key
- [x] - when publicKey is omitted, existing behavior is preserved and Re.Pack falls back to `RepackPublicKey`

#### Implementation notes:

- [x] - adds optional `publicKey` string to `ScriptLocator`
- [x] - threads it through JS normalization & native `ScriptManager` on Android and iOS
- [x] - keeps existing `verifyScriptSignature: 'strict' | 'lax' | 'off'` behavior intact
- [x] - includes cache metadata updates so changing `publicKey` invalidates the cached script config
- [x] - adds docs/examples for the new resolver pattern

## Test Plan

1. Run automated checks:
```
   pnpm --filter @callstack/repack typecheck
   pnpm --filter @callstack/repack test -- --watchman=false ScriptManager.test.ts
```
2. Manually verify the change with a local test app by returning `publicKey` from a `ScriptManager` resolver together with `verifyScriptSignature: 'strict'`
3. Confirm that a valid `publicKey` allows the signed bundle to load on iOS and Android
4. Confirm that an invalid `publicKey` fails verification on iOS and Android
5. Confirm that omitting `publicKey` still uses the existing `RepackPublicKey` fallback behavior

### Android

https://github.com/user-attachments/assets/60d7966d-e2eb-40ae-9afd-f3552da0be85

### iOS

https://github.com/user-attachments/assets/b26bf9a8-53dd-4cc0-acb4-f5d238ef85c5

### JS side `publicKey` validation (both platforms)

https://github.com/user-attachments/assets/8824bd9d-32d4-4f8c-870d-0567c414dd65

### DIFF
--- a/.changeset/gold-jokes-itch.md
+++ b/.changeset/gold-jokes-itch.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": major
+---
+
+Add support for passing a per-script `publicKey` from `ScriptManager` resolvers so signed bundles can be verified with a runtime-provided public key instead of only the app-embedded `RepackPublicKey`
--- a/packages/repack/android/src/main/java/com/callstack/repack/CodeSigningUtils.kt
+++ b/packages/repack/android/src/main/java/com/callstack/repack/CodeSigningUtils.kt
@@ -83,15 +83,17 @@ class CodeSigningUtils {
             return null
         }
 
-        fun verifyBundle(context: Context, token: String?, fileContent: ByteArray?) {
+        fun verifyBundle(
+                context: Context, token: String?, fileContent: ByteArray?, stringPublicKey: String?
+        ) {
             if (token == null) {
                 throw Exception("The bundle verification failed because no token for the bundle was found.")
             }
 
-            val stringPublicKey = getPublicKeyFromStringsIfExist(context)
+            val resolvedPublicKey = stringPublicKey ?: getPublicKeyFromStringsIfExist(context)
                     ?: throw Exception("The bundle verification failed because PublicKey was not found in the bundle. Make sure you've added the PublicKey to the res/values/strings.xml under RepackPublicKey key.")
 
-            val publicKey = parsePublicKey(stringPublicKey)
+            val publicKey = parsePublicKey(resolvedPublicKey)
                     ?: throw Exception("The bundle verification failed because the PublicKey is invalid.")
 
             val claims: Map<String, Any?> = verifyAndDecodeToken(token, publicKey)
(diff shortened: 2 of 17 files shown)


bartekkrok (approved):

dannyhw (commented):
I originally left some review comments here but I'm actually doubting the validity of those so I will update when I double check 😄

dannyhw (approved):

PR #1389, opened 2026-06-09
feat: add automated changesets release workflow

### LINKED ISSUE #1388: Automated release workflow
The changesets tool provides a setup for automated releases but we are currently still doing everything manually

https://github.com/changesets/changesets

There is an official github action here:

https://github.com/changesets/action

we should be able to use this and get a decent out of the box solution for this.

It's important that we are able to validate the release version and details before the release is actually made so effort should be made to confirm that we won't accidentally publish a major when a patch is required for example.

The solution should also generate release notes and apply them to the github releases tab

draggie:
Closes #1388

Summary:
- Adds a Changesets-based release workflow with stable release PR/publish automation and manual canary releases.
- Uses npm trusted publishing/OIDC without npm publish tokens and disables package-manager caching in release jobs.
- Adds a temporary PR release dry-run job to preview status/snapshot version generation before publishing.

Validation:
- pnpm lint:ci
- pnpm typecheck
- pnpm test
- git diff --check


package.json
```diff
@@ -13,7 +13,8 @@
     "lint:ci": "biome check",
     "typecheck": "turbo run typecheck",
     "test": "turbo run test",
-    "release": "pnpm build && pnpm lint && pnpm test && pnpm changeset publish",
+    "release": "pnpm build && pnpm lint:ci && pnpm test && pnpm changeset publish",
```
dannyhw: I think I would prefer to see these validation steps separately
like in the github action have a build, lint and test step before the publish step
so we can see more easily in the github action that for example the lint step failed rather than the release command.

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # tag=v4.1.0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # tag=v4.4.0
+        with:
+          node-version: '24'
+          registry-url: https://registry.npmjs.org/
+          package-manager-cache: false
```
dannyhw: can you double check this? I think `package-manager-cache` isn't an option? Should be `cache`? Also I think cache is not enabled unless you add the 'cache' option so its enough to just omit it

package.json
```diff
@@ -13,7 +13,10 @@
     "lint:ci": "biome check",
     "typecheck": "turbo run typecheck",
     "test": "turbo run test",
-    "release": "pnpm build && pnpm lint && pnpm test && pnpm changeset publish",
+    "version-packages": "pnpm changeset version && pnpm lint",
+    "version-packages:canary": "pnpm changeset version --snapshot canary && pnpm lint",
+    "release": "pnpm changeset publish",
+    "release:canary": "pnpm changeset publish --tag canary --no-git-tag --snapshot",
```
dannyhw: is `--snapshot` and `--no-git-tag` correct here? I don't see it on the docs:

```
Usage:
  $ changeset publish

Options:
  --otp <code>           One time password for npm publish 
  --tag <name>           Publish with the given npm dist-tag 
  --from-pack-dir <dir>  Publish from a packed output directory 
  --git-tag              Create a git tag for the release (default: true)

Examples:
  $ changeset publish --otp 123456
  $ changeset publish --tag beta
```

draggie: no-git-tag is fine, but snapshot is from different command not publish

dannyhw (approved):

dannyhw:
thanks!

PR #1392, opened 2026-07-02
Improve source map symbolication

dannyhw:
WIP based on patches received for resolving errors to editor paths etc, still needs extensive testing

## Summary
- improve `/open-stack-frame` handling by resolving macOS editor commands earlier and mapping feature-app source paths when possible
- make symbolication tolerant of bundle and hot-update frames, skip known React runtime frames, and use source map contents for code frames when available
- allow the Rspack dev server to fetch remote source maps from bundle URLs or Module Federation remote config when the local compiler cannot provide one

## Validation
- `node_modules/.bin/biome check --write packages/dev-server/src/plugins/devtools/devtoolsPlugin.ts packages/dev-server/src/plugins/symbolicate/Symbolicator.ts packages/dev-server/src/plugins/symbolicate/sybmolicatePlugin.ts packages/repack/src/commands/rspack/start.ts`
- `packages/dev-server/node_modules/.bin/tsc --noEmit -p packages/dev-server/tsconfig.json`
- `packages/repack/node_modules/.bin/tsc --noEmit -p packages/repack/tsconfig.json`
- `git diff --check`

Note: the local pre-commit hook attempted to run pnpm through the bundled runtime and stopped before checks with a non-interactive modules purge prompt, so the commit was created after running the relevant checks directly.

### DIFF
--- a/packages/dev-server/src/plugins/symbolicate/Symbolicator.ts
+++ b/packages/dev-server/src/plugins/symbolicate/Symbolicator.ts
@@ -11,6 +11,24 @@ import type {
   SymbolicatorResults,
 } from './types.js';
 
+const REACT_RUNTIME_METHOD_NAMES = new Set([
+  'react-stack-bottom-frame',
+  'renderWithHooks',
+  'updateFunctionComponent',
+  'beginWork',
+  'runWithFiberInDEV',
+  'performUnitOfWork',
+  'workLoopSync',
+  'renderRootSync',
+  'performWorkOnRoot',
+  'performSyncWorkOnRoot',
+  'flushSyncWorkAcrossRoots_impl',
+  'scheduleUpdateOnFiber',
+  'dispatchSetState',
+  'reportException',
+  'handleException',
+]);
+
 /**
  * Class for transforming stack traces from React Native application with using Source Map.
  * Raw stack frames produced by React Native, points to some location from the bundle
@@ -78,7 +96,12 @@ export class Symbolicator {
     const frames: InputStackFrame[] = [];
     for (const frame of stack) {
       const { file } = frame;
-      if (file?.startsWith('http')) {
+      if (
+        file &&
+        (file.startsWith('http') ||
+          file.includes('.bundle') ||
+          file.includes('.hot-update.js'))
+      ) {
         frames.push(frame as InputStackFrame);
       }
     }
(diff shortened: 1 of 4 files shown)


packages/dev-server/src/plugins/symbolicate/Symbolicator.ts
```diff
@@ -194,6 +241,19 @@ export class Symbolicator {
     };
   }
 
+  private getSourceContent(frame: StackFrame): string | undefined {
+    for (const consumer of Object.values(this.sourceMapConsumerCache)) {
+      try {
+        const source = consumer.sourceContentFor(frame.file, true);
```
dannyhw, line 247: codex review: `getSourceContent` searches every cached source map and returns the first map with `frame.file`, but `processFrame` has already discarded which generated bundle/source map produced that frame. In multi-bundle or federated stacks, duplicate logical names like `[projectRoot]/src/App.tsx` are plausible, so this can render a code frame from the wrong host/remote. Please carry the source-map key/consumer through `processFrame` and query that exact consumer for `sourcesContent`.

packages/dev-server/src/plugins/symbolicate/Symbolicator.ts
```diff
@@ -88,41 +111,61 @@ export class Symbolicator {
...
+            frame,
+          });
+          processedFrames.push(processedFrame);
+        } catch (error) {
+          logger.debug({
+            msg: 'Failed to symbolicate frame',
             fileUrl: frame.file,
+            error: (error as Error).message,
+          });
+          processedFrames.push({
```
dannyhw, line 164: codex review: This failed-symbolication fallback keeps the raw bundle frame visible with `collapse: false`. `getCodeFrame` later tries this raw URL first and returns after the first `getSource` failure, so one missing map can still suppress a later frame that did symbolicate correctly. Either collapse or skip failed raw frames for code-frame selection, or make `getCodeFrame` continue to the next frame after a source read miss.

packages/repack/src/commands/rspack/start.ts
```diff
@@ -26,6 +26,146 @@ import logo from '../common/logo.js';
...
+  }
+
+  return parsedUrl;
+}
+
+function getRemoteNameFromBundleUrl(url: string): string | undefined {
+  const filename = url.split(/[?#]/)[0].split('/').pop() ?? '';
+
+  return (
+    filename.match(/\.([^.]+)\.chunk\.bundle(?:\.map)?$/)?.[1] ??
```
dannyhw, line 89: codex review: This regex does not match Re.Pack's standard `[name].chunk.bundle` filenames such as `remote.chunk.bundle` because it requires a dot before the captured name. The container regex also misses the MF v1 default `<name>.container.bundle`. When direct URL fetch is unavailable, `remoteName` becomes undefined and the fallback probes all remotes, which can pick a source map from the wrong remote. Please handle the default filename shapes before falling back to all remotes.

packages/dev-server/src/plugins/symbolicate/Symbolicator.ts
```diff
@@ -146,6 +189,10 @@ export class Symbolicator {
     }
   }
 
+  private shouldSkipSourceMap(frame: InputStackFrame): boolean {
```
dannyhw, line 192: codex review: This skips purely by generated `methodName` before resolving the source. A user/app frame named `beginWork`, `dispatchSetState`, etc. would be collapsed and never symbolicated. The skip should be tied to source/package context, or applied only after the frame is known to come from React internals.

dannyhw (commented):
codex review: Posting the main review findings inline. I also did not find targeted tests for the changed symbolicator, open-stack-frame, or Rspack remote source-map fallback behavior, so the current passing checks do not exercise the highest-risk paths.

dannyhw:
I want to revisit this, I'm not confident we're solving the right thing

PR #1393, opened 2026-07-02
feat: support Rspack 2 alongside Rspack 1 - reference implementation (to be broken down)

dannyhw:
## Summary

Adds Rspack 2 support to Re.Pack while keeping Rspack 1 fully working from the same release. Re.Pack detects the installed `@rspack/core` major at runtime and adapts — no separate builds, no breaking changes for existing Rspack 1 / webpack users.

Full research, decisions, and verification behind this PR live in [`agent_context/rspackv2-jul2026/`](https://github.com/callstack/repack/tree/feat/rspack-2-support/agent_context/rspackv2-jul2026) (added on this branch): breaking-changes inventory mapped to our code, dual-version plan, decided questions, an executed V1–V11 verification checklist against `@rspack/core@2.1.2`, and a React Refresh deep dive.

### Version detection (foundation)
- New `getRspackVersion`/`getRspackMajorVersion`/`isRspack2` helpers — resolve `@rspack/core/package.json` instead of importing the package, so they're safe on any Node version and when `@rspack/core` isn't installed.

### Config generation
- `experiments.parallelLoader` is only set under Rspack 1 (removed in v2 — verified silently ignored, i.e. dead config).
- `module.parser.javascript.exportsPresence: 'auto'` under Rspack 2 — v2 changed the default from `'warn'` to `'error'`, which breaks builds on invalid imports inside node_modules (common in the RN ecosystem, tolerated by Metro). Overridable in project config.

### Persistent cache
- v2 moved `experiments.cache` → top-level `cache` and **silently ignores** the legacy key (verified: no error, cache just off). Re.Pack now reads both locations, and under v2 auto-applies a legacy `experiments.cache` value to `cache` with a warning so users don't silently lose caching. `--reset-cache` works with both shapes.

### React Refresh (biggest piece — see [agent_context/rspackv2-jul2026/06](https://github.com/callstack/repack/blob/feat/rspack-2-support/agent_context/rspackv2-jul2026/06-react-refresh-deep-dive.md))
- The `@rspack/plugin-react-refresh@1.0.0` dependency (and its removed `deprecated_runtimePaths` API) is gone.
- **Rspack 2**: applies the official `@rspack/plugin-react-refresh` v2 plugin through its new integrator options (`injectEntry: false` + `reactRefreshLoader: '@callstack/repack/react-refresh-loader'`), keeping Re.Pack's per-entrypoint entry ordering and RN-aware loader. The plugin is a new **optional peer dependency** — Rspack 2 users install it alongside `@rspack/core@2` (a friendly error explains this if missing). It's ESM-only, so it's required lazily inside the v2 branch only.
- **Rspack 1 + webpack**: the three client runtime files are vendored into `src/modules/reactRefresh/` (MIT, adapted from the v2 plugin) and wired manually exactly as before — defines updated from the removed overlay flags to `__reload_on_runtime_errors__: false`.

### Commands & environment
- Rspack commands are loaded lazily; a guard raises `Rspack X requires Node.js ^20.19.0 || >=22.12.0` instead of an obscure `ERR_REQUIRE_ESM` crash (v2 is pure ESM). Bonus: loading `react-native.config.js` no longer touches `@rspack/core` at all.
- `--trace-*` defaults to the `logger` layer under v2 (published v2 binaries are built without perfetto — verified).
- `ModuleFederationPluginV1` pre-checks `@module-federation/runtime-tools` under v2 (no longer auto-installed by `@rspack/core`) with an actionable error.

### Types & tests
- `packages/repack` now compiles against `@rspack/core@2` types (other packages stay on the v1 catalog). Fallout handled without casts: non-union `SwcConfig` alias (v2 made `SwcLoaderOptions` a union on `detectSyntax`), loader-only SWC options split off before the raw transform call, cache types derived from `Configuration['cache']` with the legacy `experiments.cache` location modeled explicitly, and `devServer` split off via destructuring in `bundle`. One irreducible cast remains in `start` (Re.Pack's `devServer` augmentation vs v2's bundled `DevServer` type differ only by duplicated http-proxy-middleware type trees) - documented in a comment at the site.
- Jest can't load the ESM-only core in its sandboxed runtime — a custom test environment preloads it natively and a `moduleNameMapper` bridge exposes it, so all suites now run against the real Rspack 2 (**280 tests, up from 234** — 4 suites previously silently type-only).

## Verification

- `pnpm build` / `typecheck` / `test` (280/280) / biome — green across all 12 package tasks.
- Smoke-tested the **built dist** against both majors in isolated projects (Node 26):
  - **Rspack 1.7.12**: config keeps `parallelLoader`, no parser override, cache accessor + migration, Node guard, and a full dev build with HMR + React Refresh using the **vendored** client files ✅
  - **Rspack 2.1.2**: no `parallelLoader`, `exportsPresence: 'auto'`, cache migration, Node guard, and a full dev build with HMR + React Refresh through the **official v2 plugin** (with Re.Pack's loader + entry ordering) ✅
- Earlier lab verification (agent_context/rspackv2-jul2026/07): the full RN chunk-loading chain (`load_script` interception, `jsonp` + `array-push` + `globalObject: 'self'`) executes correctly under v2; MF v1 + v2 (`enhanced@2.6.0`) build clean.

## Not in this PR (follow-ups per plan phases 3–4)

- CI matrix + tester-app / `tests/metro-compat` / `tests/resolver-cases` runs under v2 (needs device validation for HMR e2e)
- Flipping `repack-init`, templates, and website docs to default to Rspack 2 (decided, gated on the validation pass)
- Website migration guide for users


dannyhw:
next step is going to be to split this pr into multiple prs, this branch will serve as a reference for work done until now

dannyhw:
Closing in favor of the PR stack — this branch (`feat/rspack-2-support`) stays as the reference implementation and is not meant to merge.

The work is split as planned (design doc: [`agent_context/rspackv2-jul2026/design.md`](https://github.com/callstack/repack/blob/main/agent_context/rspackv2-jul2026/design.md), docs PR #1395 already merged):

**Sequential prefix** (each stacked on the previous):
- #1394 — version detection helpers (2a)
- #1397 — Node compatibility guard + lazy command loading (2b)
- #1398 — Rspack 2 types + dual-major test infrastructure (2c)

**Parallel set** (mutually conflict-free, based on 2c, mergeable in any order once the prefix is in):
- #1401 — config routing (3)
- #1399 — legacy cache config warning (4)
- #1400 — MFv1 runtime-tools pre-check (5)
- #1402 — React Refresh restructure (6)

**Closers** (based on a temporary integration branch = 2c + PRs 3–6; will be retargeted to main as the stack merges):
- #1404 — dual-major smoke suite + the headline "Rspack 2 support" changeset (7)
- #1403 — per-major tester apps (8)

All maintainer feedback from this PR is incorporated in the respective stack PRs (warn-only cache, vendor directory, minimized DevelopmentPlugin footprint, minimal start.ts diffs, Windows verification lane).

PR #1394, opened 2026-07-03
feat: add rspack version detection helpers

dannyhw:
## Summary

Second PR of the Rspack 2 support stack split out of #1393 (design doc in #1395). Deliberately minimal — **pure utilities, nothing wired up yet**:

- **`helpers/rspackVersion.ts`**: `getRspackVersion` / `getRspackMajorVersion` / `isRspack2` resolve the installed `@rspack/core` version via its `package.json` without importing the package — safe on any Node version (`@rspack/core@2` is ESM-only and cannot be `require()`d on Node < 20.19) and safe when rspack isn't installed (webpack-only projects). `getRspackMajorVersionFromCompiler` covers plugin contexts via `compiler.webpack.rspackVersion`.
- **`ConfigKeys` additions** (`cache`, `experiments`) in `commands/types.ts` — type-only; describes config keys the commands already read.

No behavior change, no dependency changes, no changeset. Consumers arrive with the follow-up PRs:

1. #1395 — design doc (independent)
2. **This PR** — version utilities
3. Node ≥20.19 guard for Rspack 2 + lazy command loading (clear error instead of `ERR_REQUIRE_ESM`)
4. `@rspack/core@^2` devDep + type fallout + dual-major jest lane + CI lanes (Rspack 1 + Windows)
5. Then in parallel: config routing, legacy-cache warning, MFv1 runtime-tools pre-check, React Refresh restructure; closed by a dual-major smoke suite (headline changeset) and per-major tester apps

Everything here was previously reviewed as part of #1393; the branch there stays as the reference implementation until the stack lands.

### Verification

- `turbo run typecheck test --force` — 17/17 tasks green (repack 280/280)
- `biome check` — clean

### DIFF
--- a/packages/repack/src/helpers/__tests__/rspackVersion.test.ts
+++ b/packages/repack/src/helpers/__tests__/rspackVersion.test.ts
@@ -1,0 +1,134 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import type { Compiler as RspackCompiler } from '@rspack/core';
+import type { Compiler as WebpackCompiler } from 'webpack';
+import {
+  getRspackMajorVersion,
+  getRspackMajorVersionFromCompiler,
+  getRspackVersion,
+  isRspack2,
+} from '../rspackVersion.js';
+
+/**
+ * Creates a fake project directory containing `node_modules/@rspack/core`
+ * with the given version. The fake package's entrypoint throws, so any
+ * codepath that imports `@rspack/core` (instead of only resolving its
+ * `package.json`) fails loudly.
+ */
+function createFixtureProject(root: string, version: string): string {
+  const projectDir = fs.mkdtempSync(path.join(root, 'project-'));
+  const packageDir = path.join(projectDir, 'node_modules', '@rspack', 'core');
+  fs.mkdirSync(packageDir, { recursive: true });
+  fs.writeFileSync(
+    path.join(packageDir, 'package.json'),
+    JSON.stringify({ name: '@rspack/core', version, main: 'index.js' })
+  );
+  fs.writeFileSync(
+    path.join(packageDir, 'index.js'),
+    'throw new Error("@rspack/core must not be imported by version detection");'
+  );
+  return projectDir;
+}
+
+describe('rspackVersion helpers', () => {
+  let tmpRoot: string;
+  let rspack1Project: string;
+  let rspack2Project: string;
+  let brokenProject: string;
+
+  beforeAll(() => {
+    tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'repack-rspack-version-'));
+    rspack1Project = createFixtureProject(tmpRoot, '1.7.0');
+    rspack2Project = createFixtureProject(tmpRoot, '2.3.4');
+    // simulates an unusable @rspack/core install: package.json resolves but
+    // cannot be loaded, which exercises the same `catch -> null` branch as a
+    // missing install (jest's `require.resolve` treats `paths` as additional
+    // lookup paths and falls back to the workspace install, so a truly absent
+    // package cannot be simulated in-process here)
+    brokenProject = createFixtureProject(tmpRoot, '0.0.0');
+    fs.writeFileSync(
+      path.join(
+        brokenProject,
+        'node_modules',
+        '@rspack',
+        'core',
+        'package.json'
+      ),
+      'not-json'
+    );
+  });
+
+  afterAll(() => {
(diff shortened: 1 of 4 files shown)


packages/repack/src/helpers/rspackVersion.ts
```diff
@@ -0,0 +1,61 @@
...
+ * - `@rspack/core` v2 is ESM-only, so a CJS `require('@rspack/core')` throws
+ *   `ERR_REQUIRE_ESM` on Node < 20.19,
+ * - `@rspack/core` is an optional peer dependency and may not be installed
+ *   at all (webpack-only projects).
+ *
+ * @param context Optional directory to resolve `@rspack/core` from. Defaults
+ * to Re.Pack's own resolution paths.
+ * @returns Version string or `null` when `@rspack/core` is not resolvable.
+ */
+export function getRspackVersion(context?: string): string | null {
```
dannyhw, line 18: codex review: This new helper is the contract that the Rspack 2 work depends on, but it has no direct tests. Please add focused coverage for resolving @rspack/core/package.json without importing @rspack/core, returning null when the optional peer is absent, honoring a project context, parsing major versions/isRspack2, and compiler-backed detection via compiler.webpack.rspackVersion. Without that, a later refactor could accidentally reintroduce an ESM require or lose optional-peer safety without CI catching it.

dannyhw: Added packages/repack/src/helpers/__tests__/rspackVersion.test.ts in a11dd350 (10 tests): default resolution matches the installed @rspack/core version, project-context resolution via temp fixture installs whose entrypoint throws on import (so any regression back to importing the package fails loudly), major-version parsing and isRspack2 for both majors, and compiler-backed detection via compiler.webpack.rspackVersion vs a webpack compiler. One nuance: jest's require.resolve treats the paths option as additional lookup paths and falls back to the workspace install, so a truly absent package can't be simulated in-process; the null contract is covered through the same catch branch with a fixture whose package.json resolves but can't be loaded.

.github/workflows/test.yml
```diff
@@ -53,3 +53,39 @@
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # tag=v4.2.2
+      - name: Setup workspace
+        uses: ./.github/actions/setup-pnpm-node
+        with:
+          node-version: '24'
+
+      - name: Run tests against Rspack 1
+        run: pnpm turbo run test:rspack1
+
+  test_pr_windows:
```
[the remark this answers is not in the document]
dannyhw: This PR no longer touches .github/workflows/test.yml (the diff is limited to packages/repack source). Explicit permissions blocks (contents: read) were added to both test workflows in #1398, which is part of the same stack.

.github/workflows/test.yml
```diff
@@ -53,3 +53,39 @@
...
+          node-version: '24'
+
+      # the require(esm) loading path of @rspack/core@2 and the version
+      # helpers' filesystem resolution are Windows-sensitive - run the
+      # repack unit suite under both majors
+      - name: Run repack tests
+        run: pnpm turbo run test --filter @callstack/repack
+
+      - name: Run repack tests against Rspack 1
+        run: pnpm turbo run test:rspack1
```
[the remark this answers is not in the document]
dannyhw: Same as above: this PR does not modify .github/workflows/test.yml, and #1398 adds permissions: contents: read to both test workflows.

packages/repack/src/helpers/__tests__/rspackVersion.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      expect(getRspackVersion(rspack1Project)).toBe('1.7.0');
+    });
+
+    it('does not import @rspack/core while detecting the version', () => {
+      // the fixture package's entrypoint throws on import, so a non-null
+      // result proves only package.json was read
+      expect(getRspackVersion(rspack2Project)).toBe('2.3.4');
+    });
+
+    it('returns null when the version cannot be read', () => {
```
dannyhw, line 85: codex review: This still only covers the `catch -> null` branch through invalid package metadata, not the real optional-peer-absent path where `require.resolve('@rspack/core/package.json', { paths: [project] })` throws `MODULE_NOT_FOUND`. Since optional-peer safety is part of this helper's contract, consider adding a child-process/raw Node test from a temp project with no `@rspack/core` installed; that avoids Jest's resolver fallback while locking down the actual absence case.

dannyhw (commented):
codex review: Leaving one inline test-validity concern about the optional-peer-absent path.

PR #1397, opened 2026-07-06
feat: add Rspack 2 runtime foundations and dual-major test infrastructure

dannyhw:
## Summary

Combines the remaining Rspack 2 runtime foundations into one review unit:

- the Node compatibility guard and lazy Rspack command loading from #1397;
- the Rspack 2 type migration, dual-major Jest infrastructure, CI lanes, and Windows path fix from #1398.

This PR targets `dannyhw/rspack-2-support`, where #1394 has already landed. It is the next prerequisite for the config, cache, and React Refresh work.

## What changed

- Fail clearly on unsupported Node versions before importing the ESM-only Rspack 2 core.
- Load `start` and `bundle` commands lazily and route profiling through the shared version helper.
- Compile Re.Pack against Rspack 2 types while preserving Rspack 1 runtime compatibility.
- Add honest Rspack 1 and Rspack 2 Jest lanes with a major-version guard.
- Add Rspack 1 and Windows CI coverage.
- Fix `resolveProjectPath` behavior at the Windows filesystem root.

Detailed review history for the absorbed test-infrastructure slice remains in #1398.

## References

- Feature PR: #1412
- Design: [`agent_context/rspackv2-jul2026/design.md`](https://github.com/callstack/repack/blob/main/agent_context/rspackv2-jul2026/design.md)

## Validation

The component branches previously passed:

- `pnpm turbo run typecheck test --force` — 17/17 tasks
- Re.Pack unit tests under Rspack 1 and Rspack 2
- `pnpm lint:ci`

The combined branch will run the same GitHub checks before merging into the feature branch.


packages/repack/src/commands/rspack/ensureNodeCompat.ts
```diff
@@ -0,0 +1,31 @@
...
+const RSPACK_2_NODE_RANGE = '^20.19.0 || >=22.12.0';
+
+/**
+ * Fails fast with an actionable error when the installed Rspack major
+ * requires a newer Node.js version than the current one.
+ *
+ * Without this guard, loading `@rspack/core` v2 on an unsupported Node
+ * version crashes with an obscure `ERR_REQUIRE_ESM` error.
+ */
+export function ensureNodeSupportsRspack() {
```
dannyhw, line 16: codex review: This guard is the PR’s only user-visible behavior, but there is no automated test that simulates `@rspack/core` v2 on an unsupported Node version and asserts this throws `CLIError` before `bundle`/`start` load `@rspack/core`. The current validation runs in the workspace Rspack 1 lane on Node 24, so it will not catch a regression back to the original `ERR_REQUIRE_ESM` failure. Please add a focused test that mocks `getRspackVersion()`/`process.versions.node`, plus a command-module load assertion that fails if `@rspack/core` is touched at require time.

dannyhw: Added focused coverage in 609ce329:

- `ensureNodeCompat.test.ts` unit-tests the guard with mocked `getRspackVersion()` and `process.versions.node`, covering unresolvable @rspack/core, Rspack 1.x passthrough, the `^20.19.0 || >=22.12.0` boundaries (20.18.3/20.19.0, 22.11.0/22.12.0, 21.x), prerelease 2.x versions, and the error message content.
- `commandsModuleLoad.test.ts` mocks `@rspack/core` with a throwing factory and asserts the commands module registers without touching it — this fails if the module (or anything it imports eagerly, including `bundle.js`/`start.js`) ever loads `@rspack/core` at require time again. Verified by mutation: adding an eager `import '@rspack/core'` to `index.ts` fails the suite.
- `lazyCommands.test.ts` asserts all four commands run the guard before importing the command module, and that a guard failure propagates without `bundle`/`start` ever being called.

One supporting change: the jest babel env now lowers `import(...)` to `require(...)` via `@babel/plugin-transform-dynamic-import` (test env only — verified the published dist still keeps the dynamic imports untransformed).

A true E2E lane (real unsupported Node + Rspack 2 install) is out of scope here since the workspace is pinned to Rspack 1 until the dual-version work lands; the standalone Rspack 2 test app in the later stack PRs is the place for that.

dannyhw:
to be revisited for v6

PR #1398, opened 2026-07-06
feat: add rspack 2 types and dual-major test infrastructure

dannyhw:
## Summary

Re.Pack builds and tests green against **Rspack 2**, with **zero runtime behavior change** for Rspack 1 / webpack users. This PR bumps `@callstack/repack`'s `@rspack/core` devDependency to v2 for types and tests (the workspace catalog stays on v1), absorbs the type-level fallout, and adds the dual-major test infrastructure that every later PR in the stack relies on.

**Stack position:** PR 2c, stacked on PR 2b #1397 (`feat/rspack-2-node-guard`) ← #1394 ← `main`. Reference implementation: #1393. Design doc: [`agent_context/rspackv2-jul2026/design.md`](https://github.com/callstack/repack/blob/main/agent_context/rspackv2-jul2026/design.md).

### Rspack 2 devDependency + type fallout (no runtime change)

- `@rspack/core@^2.1.2` and `@swc/helpers@^0.5.23` as `packages/repack` devDependencies — source now compiles against v2 types while remaining runtime-compatible with both majors.
- `SwcLoaderOptions` became a union discriminated on `detectSyntax` in v2 — a local non-union `SwcConfig` alias in `loaders/babelSwcLoader/options.ts` keeps the spread-and-override option helpers working. `builtin:swc-loader`-only keys (`rspackExperiments`, `transformImport`, `collectTypeScriptInfo`, `detectSyntax`) are destructured off before calling the raw SWC `transformSync` (also more correct at runtime).
- `bundle.ts` rest-destructures `devServer` off the config instead of `delete` — it isn't needed for bundling and is not assignable to v2's bundled `DevServer` type.
- **One documented cast** (calling it out per our no-casts agreement): in `start.ts`, Re.Pack's `devServer` type augmentation and Rspack 2's bundled `DevServer` type are incompatible solely because each pulls `proxy` types from a different copy of http-proxy-middleware. `devServer` can't be stripped there — the dev-server flow reads it back from `compiler.options`. The cast carries its full reasoning in a comment; the genuine fix (aligning `@callstack/repack-dev-server`'s proxy types with rspack's bundled ones) is possible future work.
- `RepackTargetPlugin` narrows `devServer` before reading `.hot` (the key can be `false` in v2 types).
- `getRspackCacheConfig` accessor in `resetPersistentCache.ts` reads the persistent cache config from both the Rspack 1 (`experiments.cache`) and Rspack 2 (top-level `cache`) locations, with derived cache types extended for the legacy location. Behavior-neutral — the user-facing warning for the legacy location lands in a later stack PR.

### Dual-major Jest lane

`@rspack/core` v2 is ESM-only, which Jest's sandboxed CJS runtime cannot load (`createRequire` inside the sandbox is wrapped by Jest and loops back into the module registry). Tests now run through a custom test environment (`jest.environment.js`) that loads the real core **outside the sandbox** with Node's real module system and bridges it back via `moduleNameMapper` (`jest.rspack-core-bridge.js`, `__esModule`-marked so Babel import interop keeps named imports working).

The environment is parameterized on `RSPACK_MAJOR` (default `2`): v2 via `await import('@rspack/core')`, v1 via plain `require` of the aliased `"@rspack/core-v1": "npm:@rspack/core@^1.7.12"` devDependency (CJS, so no reliance on cjs-module-lexer named-export synthesis). It exposes `__RSPACK_MAJOR__` for major-gated assertions, and a permanent lane-guard test (`src/__tests__/rspackTestLane.test.ts`) asserts the loaded core's major matches the requested lane, so a silently-unwired lane can't pass. New script: `test:rspack1` (`cross-env RSPACK_MAJOR=1 jest`).

### CI lanes

- **Tests [Rspack 1]**: unit suite under Rspack 1 on every PR (`pnpm turbo run test:rspack1`).
- **Tests [Windows]**: `windows-latest` lane running the repack unit suite under **both** majors — the `require(esm)` loading path of `@rspack/core@2` is Windows-sensitive.
- **Main-branch Node matrix**: the Node 18/20 lanes run with `RSPACK_MAJOR=1` — repack's Jest environment defaults to the v2 devDep, and Rspack 2 requires Node `^20.19.0 || >=22.12.0`.
- Added `permissions: contents: read` to both touched workflows (flagged by code scanning on #1394).

### Pre-existing bug fix: `resolveProjectPath` on Windows (found by the Windows lane)

Standing up the Windows CI lane surfaced a real pre-existing product bug: `resolveProjectPath` composed its result via string replacement + `path.join`, and when the `[projectRoot^N]` up-level navigation reached the filesystem root, the bare-root-plus-concatenation produced a double-separator prefix that Windows resolves as a **UNC path** (e.g. `[projectRoot^2]/shared/utils.js` with rootDir `/project/root` resolved to `\\shared\utils.js\`). Fixed at the root cause by resolving segment-by-segment with `path.resolve`, with a regression test for the root-collapse case; existing test expectations are made platform-agnostic (resolved paths carry a drive letter and win32 separators on Windows). Split into its own commit. `HermesBytecodePlugin` test expectations got the same platform-agnostic treatment (test-only).

## Design decision: plain import, no project-context resolution

`@rspack/core` is loaded with a plain `import` — shipped code carries **no project-context resolution** or workspace-aware workarounds. The consequence is monorepo-only: in-workspace apps resolve repack's own devDep (v2) instead of their manifest's pin. That is accepted and handled at the fixture level (standalone, tarball-installed test apps in a later stack PR) rather than in published code — see the design doc for the reasoning.

Same consequence for the v1 unit lane's honest scope: version-*detection* paths still resolve repack's own v2 devDep in-workspace, so the v1 lane exercises the real v1 core object and compilation surface, not version-routing logic.

## Validation

All run at the repo root unless noted:

| Command | Result |
| --- | --- |
| `pnpm turbo run typecheck test --force` | 17/17 tasks green |
| `pnpm test` (in `packages/repack`) | 282 passed, 30 suites — Rspack 2.1.2, incl. lane guard |
| `pnpm test:rspack1` (in `packages/repack`) | 282 passed, 30 suites — Rspack 1.7.12, incl. lane guard |
| `pnpm lint:ci` | clean |

## Changeset

Intentionally **no changeset**: this PR is behavior-neutral for users (devDeps, types, test/CI infrastructure). The user-visible Node-version error message ships with PR 2b (#1397) and carries the changeset there.


packages/repack/src/commands/common/resetPersistentCache.ts
```diff
@@ -4,11 +4,33 @@ import type { Configuration as RspackConfiguration } from '@rspack/core';
...
 
+/**
+ * Reads the persistent cache configuration from a Rspack config,
+ * supporting both the Rspack 1 (`experiments.cache`) and
+ * Rspack 2 (top-level `cache`) locations.
+ */
+export function getRspackCacheConfig(
+  config: RspackConfigurationWithLegacyCache
+): RspackCacheOptions {
+  return config.experiments?.cache ?? config.cache;
```
dannyhw: codex review: `getRspackCacheConfig` prefers `experiments.cache` over top-level `cache`, but Rspack 2 ignores `experiments.cache`. If a migrated/dual-major config carries both keys, or defaults are merged with a user top-level cache, `--reset-cache` will delete the legacy/default cache path and leave the active Rspack 2 cache intact. Please either collect both locations or select the cache location by detected Rspack major.

dannyhw: Fixed in de304607. `getRspackCacheConfig` is now `getRspackCacheConfigs` and collects both locations instead of preferring `experiments.cache`: it returns every defined cache config from `experiments.cache` and top-level `cache` (falling back to a single entry so the default directory is still cleaned when neither is set). Since which location is active depends on the installed Rspack major, `--reset-cache` clears both candidates rather than selecting by version - the paths are deduped, and deleting a stale legacy cache directory alongside the active one is the desired behavior for a reset. Both call sites (`rspack/start.ts`, `rspack/bundle.ts`) updated and unit tests added covering the both-keys case, including the merged-defaults scenario (`getCommandConfig` sets `experiments.cache` while a user config may set top-level `cache`).

packages/repack/src/commands/common/resolveProjectPath.ts
```diff
@@ -13,7 +13,11 @@ export function resolveProjectPath(filepath: string, rootDir: string) {
   if (!match) return filepath;
 
   const [prefix, upLevels] = match;
-  const upPath = '../'.repeat(Number(upLevels ?? 0));
-  const rootPath = path.join(rootDir, upPath);
-  return path.resolve(filepath.replace(prefix, rootPath));
+  const upSegments = Array.from({ length: Number(upLevels ?? 0) }, () => '..');
+  const restPath = filepath.slice(prefix.length + 1);
```
dannyhw: codex review: This can still escape `rootDir` when the marker is followed by duplicate separators. For example, `[projectRoot]//src/index.js` makes `restPath` start with `/`, so `path.resolve(rootDir, ...upSegments, restPath)` treats it as absolute and resolves to `/src/index.js` instead of staying under `/project/root/src/index.js` as the previous implementation did. Please strip leading path separators from the remainder or split it into relative segments before resolving, and add a regression test for the duplicate-separator case.

packages/repack/src/loaders/babelSwcLoader/babelSwcLoader.ts
```diff
@@ -163,8 +163,18 @@ export default async function babelSwcLoader(
       sourceType: babelResult.sourceType,
     });
 
+    // `SwcConfig` can carry `builtin:swc-loader`-only options which the raw
```
dannyhw, line 166: codex review: The cleanup only applies to `finalSwcConfig`; a JS config can still reintroduce raw-SWC-rejected loader-only keys through `options.swcOverrides`, which is spread into the `transformSync` options after this block. Raw SWC rejects `detectSyntax`, `transformImport`, `collectTypeScriptInfo`, and `rspackExperiments`, so if the goal is to guarantee these never reach `transformSync`, please sanitize after merging overrides or validate/reject those keys explicitly.

dannyhw (commented):
codex review: posting two actionable findings from the re-review.

packages/repack/src/commands/common/resolveProjectPath.ts
dannyhw: Fixed in 18e54f4 - the remainder after the [projectRoot] token is now split into segments (empty segments dropped) and resolved relative to the root, so a leading separator can no longer escape the project root or drop the [projectRoot^N] up-segments; added duplicate-separator regression cases for both [projectRoot] and [projectRoot^N].

packages/repack/src/loaders/babelSwcLoader/babelSwcLoader.ts
dannyhw: Keeping the current behavior deliberately: SwcOverrides omits these keys at the type level, and a JS config that reintroduces them fails loudly with SWC's unknown-field error rather than being silently stripped - silent stripping would mask the misconfiguration. Happy to revisit if this shows up in practice.

dannyhw:
Combined into #1397 as the runtime-foundations and dual-major test-infrastructure review unit. GitHub marked this PR merged when its full history became part of #1397's branch; #1397 now targets the Rspack 2 feature branch.

PR #1399, opened 2026-07-06
feat: warn about legacy experiments.cache config under rspack 2

dannyhw:
Part of the Rspack 2 support PR stack (reference implementation: #1393, design doc: [`agent_context/rspackv2-jul2026/design.md`](https://github.com/callstack/repack/blob/main/agent_context/rspackv2-jul2026/design.md)).

## What

Emits a **one-time warning** from the rspack `start`/`bundle` commands when running under Rspack 2 with a legacy `experiments.cache` value in the config, pointing the user at the top-level `cache` option. The config is **left untouched** — no value is copied or deleted.

## Why

Rspack 2 moved persistent cache configuration from `experiments.cache` to the top-level `cache` option and **silently ignores the legacy key** (validation is loose). A user migrating a Rspack 1 config would lose persistent caching with no signal at all.

Warn-only (rather than auto-migrating the value) follows the explicit maintainer guidance on #1393: users migrate their own config, and mutating it at runtime would make Re.Pack behave differently from bare Rspack given the same config. This behavior was verified against the built dist on the reference branch: warns exactly once, config untouched.

## Changes

- `commands/common/warnLegacyRspackCacheConfig.ts` (new) — the one-time warning, keyed off the `RspackConfigurationWithLegacyCache` type introduced with the cache accessor
- `commands/common/index.ts` — export
- `commands/rspack/start.ts` / `bundle.ts` — call sites only, gated on `isRspack2`; the diff is kept to the minimal wiring (the helper lives in `commands/common/` so it could serve a webpack counterpart if one ever existed, per the command-path-convergence feedback on #1393)
- `.changeset/rspack-2-cache-warning.md` — patch changeset

## Stack position

Parallel set — based on PR 2c (#1398) ← 2b (#1397) ← 2a (#1394). No shared files with the other parallel-set PRs (this is the only one touching `start.ts`/`bundle.ts`).

## Validation

- `pnpm turbo run typecheck test --force` — 17/17 tasks green
- `packages/repack`: `pnpm test` and `pnpm test:rspack1` — 282/282 under both Rspack majors
- `pnpm lint:ci` — clean

### DIFF
--- a/packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
+++ b/packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
@@ -1,0 +1,32 @@
+import * as colorette from 'colorette';
+import type { RspackConfigurationWithLegacyCache } from './resetPersistentCache.js';
+
+let warningDisplayed = false;
+
+/**
+ * Rspack 2 moved the persistent cache configuration from `experiments.cache`
+ * to the top-level `cache` option and silently ignores the legacy key
+ * (validation is loose) - users migrating a Rspack 1 config would lose
+ * persistent caching without any signal.
+ *
+ * Warn (once) when running Rspack 2 with a legacy `experiments.cache` value,
+ * but leave the config untouched - migrating it is the user's move, and
+ * mutating it here would make Re.Pack behave differently from bare Rspack
+ * given the same config.
+ */
+export function warnLegacyRspackCacheConfig(
+  configs: RspackConfigurationWithLegacyCache[]
+) {
+  if (warningDisplayed) return;
+  if (configs.every((config) => config.experiments?.cache === undefined)) {
+    return;
+  }
+  warningDisplayed = true;
+  console.warn(
+    colorette.yellow(
+      "Rspack 2 ignores the legacy 'experiments.cache' option, so persistent " +
+        "caching is NOT enabled. Move the value to the top-level 'cache' " +
+        'option in your Rspack config.\n'
+    )
+  );
+}
--- a/packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
+++ b/packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
@@ -7,35 +7,95 @@ function hasPersistentCacheEnabled(config: RspackConfigurationWithLegacyCache) {
   return typeof config.cache === 'object' && config.cache.type === 'persistent';
 }
 
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+function isDeepEqual(a: unknown, b: unknown): boolean {
+  if (a === b) return true;
+  if (Array.isArray(a) && Array.isArray(b)) {
+    return (
+      a.length === b.length && a.every((item, i) => isDeepEqual(item, b[i]))
+    );
+  }
+  if (isPlainObject(a) && isPlainObject(b)) {
+    const aKeys = Object.keys(a);
+    const bKeys = Object.keys(b);
+    return (
+      aKeys.length === bKeys.length &&
+      aKeys.every((key) => key in b && isDeepEqual(a[key], b[key]))
+    );
+  }
+  return false;
+}
+
+/**
+ * A legacy `experiments.cache` value carries extra information when it is an
+ * object with options beyond `type` that are not mirrored by the top-level
+ * `cache` value - Rspack 2 drops those options and runs with defaults.
+ * Booleans, a bare `{ type: 'persistent' }` and a value deep-equal to the
+ * top-level `cache` are inert leftovers of a completed migration.
+ */
+function legacyCacheCarriesExtraOptions(
+  config: RspackConfigurationWithLegacyCache
+) {
+  const legacyCache = config.experiments?.cache;
+  if (!isPlainObject(legacyCache)) return false;
+  const extraKeys = Object.keys(legacyCache).filter((key) => key !== 'type');
+  if (extraKeys.length === 0) return false;
+  return !isDeepEqual(legacyCache, config.cache);
+}
+
 /**
  * Rspack 2 moved the persistent cache configuration from `experiments.cache`
  * to the top-level `cache` option and silently ignores the legacy key
  * (validation is loose) - users migrating a Rspack 1 config would lose
  * persistent caching without any signal.
  *
- * Warn (once) when running Rspack 2 with a legacy `experiments.cache` value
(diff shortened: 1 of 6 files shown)


packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Warn (once) when running Rspack 2 with a legacy `experiments.cache` value,
+ * but leave the config untouched - migrating it is the user's move, and
+ * mutating it here would make Re.Pack behave differently from bare Rspack
+ * given the same config.
+ */
+export function warnLegacyRspackCacheConfig(
+  configs: RspackConfigurationWithLegacyCache[]
+) {
+  if (warningDisplayed) return;
+  if (configs.every((config) => config.experiments?.cache === undefined)) {
```
dannyhw: codex review: This only checks whether `experiments.cache` exists, but the warning says persistent caching is “NOT enabled”. That is false for a migrated config that keeps the legacy key around while also setting top-level `cache: { type: 'persistent' }`, and webpack-merge preserves that shape. It is also reachable through the existing `REPACK_EXPERIMENTAL_CACHE` defaults if the user then adds the top-level persistent cache. Please either only make the “NOT enabled” claim when top-level persistent cache is absent, or reword this to warn that the legacy key is ignored without asserting the effective cache state.

packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+ * to the top-level `cache` option and silently ignores the legacy key
+ * (validation is loose) - users migrating a Rspack 1 config would lose
+ * persistent caching without any signal.
+ *
+ * Warn (once) when running Rspack 2 with a legacy `experiments.cache` value,
+ * but leave the config untouched - migrating it is the user's move, and
+ * mutating it here would make Re.Pack behave differently from bare Rspack
+ * given the same config.
+ */
+export function warnLegacyRspackCacheConfig(
```
dannyhw, line 68: codex review: This new behavior needs direct test coverage. The helper has module-level once-only state, command call sites are gated on Rspack 2, `start` passes multiple configs, and the important migrated shape (`experiments.cache` plus top-level persistent `cache`) is currently untested. A small helper test matrix would catch the false-warning case and protect the intended once-only behavior.

dannyhw (commented):
codex review: I reviewed this against the stacked base branch `feat/rspack-2-types-test-infra`. I’m leaving the concrete issues inline.

packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
dannyhw: Fixed in c90a6773. The warning is now gated on the top-level persistent cache being absent in the same config: it only fires for configs where 'experiments.cache' is set and 'cache' is not an object with type 'persistent'. This keeps the strong "NOT enabled" message accurate for the cases where it fires — including the typical Rspack 1 shape ('cache: true' + 'experiments.cache', where 'cache: true' under Rspack 2 only enables the memory cache) and the REPACK_EXPERIMENTAL_CACHE defaults — while staying silent for a migrated config that only left the inert legacy key behind.

packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
dannyhw: Added in c90a6773: a 7-case matrix in packages/repack/src/commands/common/__tests__/warnLegacyRspackCacheConfig.test.ts covering legacy-key-only, the typical Rspack 1 shape ('cache: true' + 'experiments.cache'), memory-only top-level cache, the migrated shape ('experiments.cache' plus top-level persistent 'cache' — the false-warning case, now asserted silent), multi-config arrays with a mix of migrated and legacy configs, and the once-only latch across repeated calls (jest.resetModules() re-evaluates the module-level state per test). The Rspack 2 gating at the call sites is version detection covered by the isRspack2 helper tests in the foundations PR (#1394).

dannyhw:
Both inline review points addressed in c90a6773:

- **False "NOT enabled" claim**: the warning is now gated on top-level persistent cache being absent in the same config that carries the legacy key, so a migrated config with a leftover 'experiments.cache' no longer gets a false warning. It still fires for the shapes where the claim is true (legacy-only, 'cache: true' + legacy key, memory-only top-level cache).
- **Test coverage**: added a helper test matrix in packages/repack/src/commands/common/__tests__/warnLegacyRspackCacheConfig.test.ts (7 cases, including the migrated shape and the once-only latch).

Validated with turbo typecheck/test across the monorepo, the rspack 1 jest lane in packages/repack, and biome.

packages/repack/src/commands/common/warnLegacyRspackCacheConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+ * differently from bare Rspack given the same config.
+ */
+export function warnLegacyRspackCacheConfig(
+  configs: RspackConfigurationWithLegacyCache[]
+) {
+  if (warningDisplayed) return;
+  const misconfigured = configs.some(
+    (config) =>
+      config.experiments?.cache !== undefined &&
+      !hasPersistentCacheEnabled(config)
```
dannyhw, line 75: codex review: This now suppresses the warning as soon as top-level `cache.type` is `persistent`, but that also hides partially migrated configs where the ignored legacy value still carries options that are not present at the top level, e.g. `experiments.cache.storage` or `buildDependencies`. Rspack 2 would use the top-level persistent cache and silently drop those legacy options, so the user gets no signal even though there is still legacy cache config to migrate. The new test asserts silence for the broad `cache: { type: "persistent" }` + `experiments.cache` shape, so it would miss this partial-migration regression. Please either keep warning when `experiments.cache` contains ignored options with wording that does not claim persistent caching is disabled, or only suppress when the legacy cache value is known to be fully represented by the top-level cache.

dannyhw: Fixed in 614bdfe1 - the suppression is now two-tier: the strong warning still fires when no top-level persistent cache is set; when it is set but experiments.cache carries options beyond type that are not reflected there, a softer warning says those options are ignored and should be moved to the top-level cache option or removed. Silence only remains for a truly inert legacy key: a boolean, a bare { type: "persistent" }, or a value deep-equal to the top-level cache. Tests updated to cover both partial-migration shapes plus the deep-equal and strong-wins-over-soft cases.

dannyhw:
Combined into #1401 as the configuration and cache-compatibility review unit. This PR remains available for its detailed review history.

PR #1400, opened 2026-07-06
feat: check module federation runtime-tools availability under rspack 2

dannyhw:
## Summary

Adds a pre-check to `ModuleFederationPluginV1.apply`: when running under Rspack >= 2, verify that `@module-federation/runtime-tools` is resolvable from the compiler context and fail with an actionable error if it is not.

## Why

In Rspack 2, `@module-federation/runtime-tools` became an optional peer dependency of `@rspack/core` and is no longer installed automatically. The built-in `ModuleFederationPluginV1` still needs it, so MFv1 users on Rspack 2 must install it explicitly. Without this check, the missing package surfaces as a cryptic module resolution failure at build time; with it, the user gets a clear message telling them exactly what to install.

Under Rspack 1 (and webpack) nothing changes — the check only runs when the compiler reports major version >= 2.

This change was maintainer-approved on the reference PR #1393 (feedback item 8: "MFv1 changes look good").

## Stack position

Part of the Rspack 2 support PR stack (reference implementation: #1393, design doc: `agent_context/rspackv2-jul2026/design.md`).

- Sequential prefix: #1394 (version utilities) <- #1397 (Node guard + lazy command loading) <- #1398 (Rspack 2 types + dual-major test infrastructure)
- **This PR** is part of the parallel set based on the prefix (targets #1398's branch; merges after the prefix lands). It shares no files with the other parallel-set PRs — it touches only `packages/repack/src/plugins/ModuleFederationPluginV1.ts` plus its own uniquely-named changeset.

## Files changed

- `packages/repack/src/plugins/ModuleFederationPluginV1.ts` — the pre-check (uses `getRspackMajorVersionFromCompiler` from #1394)
- `.changeset/rspack-2-mfv1-precheck.md` — patch changeset

## Validation

- `pnpm turbo run typecheck test --force` — 17/17 tasks green
- `packages/repack`: `pnpm test` (Rspack 2 lane) — 282/282 green
- `packages/repack`: `pnpm test:rspack1` (Rspack 1 lane) — 282/282 green
- `pnpm lint:ci` — clean

### DIFF
--- a/packages/repack/src/plugins/ModuleFederationPluginV1.ts
+++ b/packages/repack/src/plugins/ModuleFederationPluginV1.ts
@@ -1,6 +1,9 @@
 import type { Compiler as RspackCompiler, container } from '@rspack/core';
 import type { Compiler as WebpackCompiler } from 'webpack';
-import { isRspackCompiler } from '../helpers/index.js';
+import {
+  getRspackMajorVersionFromCompiler,
+  isRspackCompiler,
+} from '../helpers/index.js';
 import { Federated } from '../utils/federated.js';
 
 type MFPluginV1 = typeof container.ModuleFederationPluginV1;
@@ -123,6 +126,31 @@ export class ModuleFederationPluginV1 {
     }
     // @ts-expect-error webpack has MF1 under ModuleFederationPlugin
     return compiler.webpack.container.ModuleFederationPlugin;
+  }
+
+  /**
+   * In Rspack 2, `@module-federation/runtime-tools` became an optional
+   * peer dependency of `@rspack/core` and is no longer installed
+   * automatically. The built-in ModuleFederationPluginV1 still needs it,
+   * so verify it's installed to fail with an actionable error instead of
+   * a raw module resolution error at build time.
+   */
+  private ensureModuleFederationRuntimeToolsInstalled(context: string) {
+    try {
+      require.resolve('@module-federation/runtime-tools', {
+        paths: [context],
+      });
+    } catch {
+      const error = new Error(
+        '[RepackModuleFederationPluginV1] ' +
+          "Dependency named '@module-federation/runtime-tools' is required when using Module Federation with Rspack 2 " +
+          '(it became an optional peer dependency of @rspack/core) but is not found in your project. ' +
+          'Did you forget to install it?'
+      );
+      // remove the stack trace to make the error more readable
+      error.stack = undefined;
+      throw error;
+    }
   }
 
   private replaceRemotes<T extends string | string[] | RemotesObject>(
@@ -279,6 +307,11 @@ export class ModuleFederationPluginV1 {
 
   apply(__compiler: unknown) {
     const compiler = __compiler as RspackCompiler;
+
+    const rspackMajor = getRspackMajorVersionFromCompiler(compiler);
+    if (rspackMajor !== null && rspackMajor >= 2) {
+      this.ensureModuleFederationRuntimeToolsInstalled(compiler.context);
+    }
 
     const ModuleFederationPlugin = this.getModuleFederationPlugin(compiler);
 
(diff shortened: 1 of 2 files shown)


packages/repack/src/plugins/ModuleFederationPluginV1.ts
```diff
@@ -280,6 +308,11 @@ export class ModuleFederationPluginV1 {
   apply(__compiler: unknown) {
     const compiler = __compiler as RspackCompiler;
 
+    const rspackMajor = getRspackMajorVersionFromCompiler(compiler);
+    if (rspackMajor !== null && rspackMajor >= 2) {
+      this.ensureModuleFederationRuntimeToolsInstalled(compiler.context);
```
dannyhw, line 313: codex review: This guard appears to reject valid Rspack 2 MFv1 builds. Re.Pack delegates to `compiler.webpack.container.ModuleFederationPluginV1`, but in a throwaway project with only `@rspack/core@2.1.2` installed, `@module-federation/runtime-tools` was `MODULE_NOT_FOUND` and Rspack `ModuleFederationPluginV1` still built shared-only, host/remote, and Re.Pack-style promise-remote configs with 0 errors/warnings. The missing runtime-tools failure belongs to Rspack’s newer `container.ModuleFederationPlugin`, not `ModuleFederationPluginV1`, so this turns an unused optional peer into a hard requirement for MFv1 users.

dannyhw: Confirmed, and reproduced independently. With @rspack/core@2.1.2 in a scratch project where `require.resolve('@module-federation/runtime-tools', { paths: [context] })` throws MODULE_NOT_FOUND, an MFv1 config with `exposes` + promise-based remotes + a consumed singleton `shared` builds with 0 errors/warnings, the emitted container executes correctly (`init`/`get` return the exposed module), and the output contains no `@module-federation` references.

Source-level confirmation: in the @rspack/core 2.1.2 dist, every `runtime-tools` reference lives inside `container.ModuleFederationPlugin` (MF 1.5), which resolves it from `compiler.context` and already throws its own actionable install error when missing. `ModuleFederationPluginV1.apply` only applies ShareRuntimePlugin/ContainerPlugin/ContainerReferencePlugin/SharePlugin and never touches runtime-tools — same split in 1.7.12, so V1 has never needed the package in either supported major. The optional-peer change (rspack#12663) only affects the enhanced plugin, which Re.Pack does not delegate to (our MFv2 plugin uses @module-federation/enhanced with its own dependency check).

So the guard is checking for a dependency the delegated plugin does not use and would hard-fail valid MFv1 builds. Since the pre-check is this PR's only change, the right move is to drop it — likely closing this PR — and correct the corresponding paragraph in agent_context/rspackv2-jul2026/design.md (§ Module Federation v1, merged in #1395), which encodes the same incorrect assumption. Leaving that call to the stack owner rather than emptying the branch here.

dannyhw:
Closing unmerged: the premise of this PR doesn't hold. Rspack's delegated `ModuleFederationPluginV1` never resolves `@module-federation/runtime-tools` under either supported major — only the enhanced MF 1.5 plugin (`container.ModuleFederationPlugin`) does, and that one already raises its own actionable install error. Verified empirically (clean MFv1 build + end-to-end container execution under @rspack/core@2.1.2 with the package absent) and in the 1.7.12/2.1.2 sources — full evidence in the review thread above.

Since the check would have rejected working Rspack 2 MFv1 builds, dropping it entirely is the correct outcome. The design doc's MFv1 section will be corrected in a follow-up docs PR. Stack impact: `feat/rspack-2-integration` is regenerated without this branch and #1403/#1404 are rebased accordingly; the rest of the stack (#1394 → #1397 → #1398, then #1399/#1401/#1402, then #1404/#1403) is unaffected.

PR #1401, opened 2026-07-06
feat: add Rspack 2 config and cache compatibility

dannyhw:
## Summary

Combines Rspack-major configuration compatibility into one review unit:

- version-aware config and profiling routing from #1401;
- the warn-only legacy `experiments.cache` handling from #1399.

This PR targets the combined runtime-foundations PR #1397. After #1397 lands, it will retarget to `dannyhw/rspack-2-support`.

## What changed

- Keep `experiments.parallelLoader` only for Rspack 1.
- Default Rspack 2 `exportsPresence` to the React Native-friendly `auto` behavior.
- Skip the obsolete global parallel-loader warning probe under Rspack 2.
- Route profiling to the logger trace layer for published Rspack 2 builds.
- Warn once when Rspack 2 sees legacy `experiments.cache`, without mutating user configuration.
- Preserve reset-cache support for both legacy and top-level cache locations.

Detailed review history for the absorbed cache slice remains in #1399.

## References

- Feature PR: #1412
- Design: [`agent_context/rspackv2-jul2026/design.md`](https://github.com/callstack/repack/blob/main/agent_context/rspackv2-jul2026/design.md)

## Validation

The component branches previously passed:

- `pnpm turbo run typecheck test --force` — 17/17 tasks
- Re.Pack unit tests under Rspack 1 and Rspack 2
- `pnpm lint:ci`

The combined branch will run the same GitHub checks before merging.

PR #1402, opened 2026-07-06
feat: restructure react refresh support for rspack 2

dannyhw:
Part of the Rspack 2 support stack (reference implementation: #1393, design doc: [`agent_context/rspackv2-jul2026/design.md`](https://github.com/callstack/repack/blob/main/agent_context/rspackv2-jul2026/design.md)).

Restructures React Refresh so `DevelopmentPlugin` works under both Rspack majors and webpack. This is the only PR in the stack that touches `packages/repack/package.json`'s dependency surface.

## Design: split on the rspack major

`@rspack/plugin-react-refresh@1` cannot stay: its module-load-time `deprecated_runtimePaths` access hard-crashes when v2 of the plugin is installed, and depending on v1 while needing `^2` as a peer is unresolvable (same package name). Per the design doc's React Refresh decision (maintainer-validated in the #1393 review — vendoring the client files is fine, `deprecated_runtimePaths` was only ever a convenience):

- **Rspack >= 2**: apply the official `@rspack/plugin-react-refresh` v2 with its integrator options — `injectEntry: false` (Re.Pack controls entry placement per entrypoint), `forceEnable: true` (DevelopmentPlugin already gates on `devServer.hot`), and `reactRefreshLoader` pointed at Re.Pack's React Native-aware `react-refresh-loader`. The package is ESM-only (named export only), so it is `require()`d lazily inside the rspack-2 branch — it is never evaluated for webpack/Rspack 1 users — and resolved from the project context with an actionable error when missing.
- **webpack + Rspack 1**: keep the existing manual wiring, now pointed at client runtime files **vendored** from `@rspack/plugin-react-refresh@2.0.2` (MIT) instead of the removed v1 dependency.

The behavioral fork is a single ternary on `getRspackMajorVersionFromCompiler(compiler)`; everything else is extraction. A per-bundler plugin split (floated in the #1393 review) is not warranted for one ~10-line fork.

## Vendor directory placement (review feedback)

The vendored files live at package-root **`packages/repack/vendor/react-refresh/`** — outside `src/`, per the review ask. They are plain JS client-runtime modules bundled by the *user's* compiler, so they ship as-is via `package.json#files` (no babel processing) and are excluded from biome. `LICENSE` in the directory carries the upstream MIT license plus a provenance note: the files are verbatim upstream v2.0.2 apart from provenance headers and repo formatting (no functional edits; re-derivation instructions for future bumps included).

## Footprint check (review feedback: minimize the DevelopmentPlugin diff)

Verified with `git diff --color-moved=dimmed-zebra --color-moved-ws=allow-indentation-change origin/main...HEAD -- packages/repack/src/plugins/DevelopmentPlugin.ts`: the manual wiring (`setupManualReactRefresh`) registers as a **move** of the old inline block (193 moved lines). The only non-moved lines inside it are the three intentional deltas:

1. **Runtime-file source**: `ReactRefreshPlugin.deprecated_runtimePaths` destructure → `require.resolve('../../vendor/react-refresh/*.js')` (resolves identically from `src/plugins` and `dist/plugins`).
2. **Define-flag swap**: `__react_refresh_error_overlay__: false` + `__react_refresh_socket__: false` → `__reload_on_runtime_errors__: false` — the v2.0.2 client files dropped the overlay/socket knobs and read the new flag instead. Value semantics unchanged: all web-oriented behaviors stay off; RN surfaces errors via LogBox.
3. **Loader rule matcher**: `include:` → `test:` with the same regex (both match the resource path here; `test` is the semantically correct key).

Two non-substantive accompaniments, listed for completeness rather than hidden: the loader-rule regexes are hoisted into module constants (`REACT_REFRESH_LOADER_TEST` / `REACT_REFRESH_LOADER_EXCLUDE`, identical literal values) because both branches now share them, and comments were updated around the moved block. Everything else in the file diff is genuinely new code for the rspack-2 branch (`resolveReactRefreshPluginRequest`, `setupOfficialReactRefresh`).

## Dependency changes

| Package | Before | After | Notes |
| --- | --- | --- | --- |
| `@rspack/plugin-react-refresh` | dependency `1.0.0` | **optional peerDependency `^2.0.0`** | Rspack 2 users install it in their project; webpack/Rspack 1 users need nothing |
| `@rspack/plugin-react-refresh` (devDep) | — | `^2.0.2` | so the workspace/tests exercise the official-plugin path |
| `package.json#files` | — | + `"vendor"` | ships the vendored client runtime |

What users must do:

- **Rspack 2**: install `@rspack/plugin-react-refresh@^2` and `react-refresh` (pnpm-strict layouts don't hoist Re.Pack's `react-refresh` copy into the official plugin's resolution scope).
- **Rspack 1 / webpack**: nothing. Runtime behavior is preserved via the vendored upstream v2.0.2 client files (MIT, LICENSE + provenance included) — same wiring shape (`ProvidePlugin` ×2 + `DefinePlugin` + loader rule + entry), same values modulo the enumerated define-flag rename that the v2.0.2 client files require.

## Delivery position

This focused PR targets #1397, which now combines the runtime foundations and dual-major test infrastructure. After #1397 merges into `dannyhw/rspack-2-support`, this PR will retarget to that feature branch and merge there for combined verification in #1412.

It remains independent of the config/cache unit in #1401. The reduced implementation set is #1397, #1401, this PR, and #1404; #1398 was absorbed into #1397, #1399 into #1401, and #1403 into #1404.

## Validation

- `pnpm turbo run typecheck test --force`: 17/17 tasks green (includes tester-app bundle/start tests, which exercise the dev server under both the rspack and **webpack** lanes).
- `packages/repack`: `pnpm test` (v2 lane) **and** `pnpm test:rspack1` (v1 lane) — 282/282 each.
- `pnpm lint:ci`: clean (vendored files excluded as intended).
- **Webpack-path evidence** (review feedback: test both bundler paths): started `tester-app` with `USE_WEBPACK=1` and fetched the served dev bundle — it contains all three vendored runtime modules (`vendor/react-refresh/reactRefresh.js` ×90 refs, `reactRefreshEntry.js` ×3, `refreshUtils.js` ×3) and **zero** references to `@rspack/plugin-react-refresh`.
- Vendored files diffed against the actual `@rspack/plugin-react-refresh@2.0.2` `client/` files from the registry: identical apart from provenance headers and formatting (no functional deltas).
- On the reference branch, this exact code was device-validated (Pixel 7 + iPhone 17 simulator): live React Refresh edit+revert under Rspack 2 (official plugin) and Rspack 1 (vendored runtime).

> **Pending manual verification:** a device HMR pass on this branch under Rspack 1, Rspack 2, and webpack. The reference branch (#1393) passed the Rspack 1 + Rspack 2 device HMR checks with identical code, but this branch has not yet had its own on-device pass — flagging so it can happen before/at merge.

### DIFF
--- a/packages/repack/package.json
+++ b/packages/repack/package.json
@@ -26,6 +26,7 @@
     "client",
     "commands",
     "mf",
+    "vendor",
     "callstack-repack.podspec",
     "src/modules/ScriptManager/NativeScriptManager.ts"
   ],
@@ -66,6 +67,7 @@
   "peerDependencies": {
     "@module-federation/enhanced": ">=0.6.10",
     "@rspack/core": ">=1",
+    "@rspack/plugin-react-refresh": "^2.0.0",
     "react-native": ">=0.74",
     "webpack": ">=5.90"
   },
@@ -76,6 +78,9 @@
     "@rspack/core": {
       "optional": true
     },
+    "@rspack/plugin-react-refresh": {
+      "optional": true
+    },
     "webpack": {
       "optional": true
     }
@@ -83,7 +88,6 @@
   "dependencies": {
     "@callstack/repack-dev-server": "workspace:*",
     "@discoveryjs/json-ext": "^0.5.7",
-    "@rspack/plugin-react-refresh": "1.0.0",
     "babel-loader": "^9.2.1",
     "colorette": "^2.0.20",
     "dedent": "^0.7.0",
@@ -117,6 +121,7 @@
     "@module-federation/sdk": "0.6.10",
     "@rspack/core": "^2.1.2",
     "@rspack/core-v1": "npm:@rspack/core@^1.7.12",
+    "@rspack/plugin-react-refresh": "^2.0.2",
     "@swc/helpers": "^0.5.23",
     "@types/babel__core": "^7.20.5",
     "@types/dedent": "^0.7.0",
--- a/packages/repack/src/plugins/DevelopmentPlugin.ts
+++ b/packages/repack/src/plugins/DevelopmentPlugin.ts
@@ -5,14 +5,20 @@ import type {
   Plugins,
   Compiler as RspackCompiler,
 } from '@rspack/core';
-import ReactRefreshPlugin from '@rspack/plugin-react-refresh';
 import type { Compiler as WebpackCompiler } from 'webpack';
-import { isRspackCompiler, moveElementBefore } from '../helpers/index.js';
-
-const [reactRefreshEntryPath, reactRefreshPath, refreshUtilsPath] =
-  ReactRefreshPlugin.deprecated_runtimePaths;
+import {
+  getRspackMajorVersionFromCompiler,
+  isRspackCompiler,
+  moveElementBefore,
+} from '../helpers/index.js';
 
 type PackageJSON = { version: string };
+
+// matches the file conditions of Re.Pack's JS transform rules
+// (includes .flow files, unlike the react-refresh plugin defaults)
+const REACT_REFRESH_LOADER_TEST = /\.([cm]js|[jt]sx?|flow)$/i;
+const REACT_REFRESH_LOADER_EXCLUDE = /node_modules/i;
+
 /**
  * {@link DevelopmentPlugin} configuration options.
  */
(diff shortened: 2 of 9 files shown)


packages/repack/src/plugins/DevelopmentPlugin.ts
```diff
@@ -87,6 +93,113 @@ export class DevelopmentPlugin {
...
+      // full reload on unrecoverable runtime errors is a web-oriented
+      // behavior - in React Native errors are surfaced through LogBox
+      __reload_on_runtime_errors__: false,
+    }).apply(compiler);
+
+    new compiler.webpack.ProvidePlugin({
+      __react_refresh_utils__: refreshUtilsPath,
+    }).apply(compiler);
+
+    compiler.options.module.rules.unshift({
```
dannyhw, line 194: codex review: The manual React Refresh path now resolves the vendored runtime files from `packages/repack/vendor/...`, but the injected refresh-loader rule still only excludes `/node_modules/i`. In workspace or symlinked installs, those vendored files are outside `node_modules`, so they match this rule and get processed by `@callstack/repack/react-refresh-loader` themselves. I confirmed this in the tester webpack dev bundle logs: `packages/repack/vendor/react-refresh/reactRefresh*.js` appears in a loader chain containing `packages/repack/dist/loaders/reactRefreshLoader/index.js`. The previous dependency-based runtime avoided this because it lived under `node_modules`, and the official Rspack v2 plugin explicitly excludes its runtime paths. Please exclude `reactRefreshPath`, `refreshUtilsPath`, and `reactRefreshEntryPath` from this manual rule so the runtime is not instrumented.

dannyhw: Good catch — confirmed against upstream: v2's injected rule excludes `{ or: [options.exclude, [reactRefreshEntryPath, reactRefreshPath, refreshUtilsPath, refreshRuntimeDir]] }`, so the official-plugin path (Rspack 2) already self-excludes; only the manual wiring was missing it. Fixed in 05b9027f by adding the three vendored file paths to the manual rule's exclude (array condition, OR semantics — matches upstream behavior). Note this slightly extends the enumerated deltas from the byte-equivalence baseline for `setupManualReactRefresh`: the rule's exclude is now the fourth intentional difference. Validated with typecheck + tests on both rspack majors.

dannyhw (commented):
codex review: Posting the actionable findings from the latest re-review.

packages/repack/src/plugins/DevelopmentPlugin.ts
```diff
@@ -87,6 +93,120 @@ export class DevelopmentPlugin {
...
+  private setupOfficialReactRefresh(compiler: RspackCompiler): string {
+    const pluginPath = this.resolveReactRefreshPluginRequest(
+      compiler.context,
+      '@rspack/plugin-react-refresh'
+    );
+    // Rspack 2 requires Node >= 20.19 (enforced by Re.Pack commands),
+    // where require() of an ESM module is supported
+    const { ReactRefreshRspackPlugin } = require(pluginPath);
+
+    new ReactRefreshRspackPlugin({
```
dannyhw, line 138: codex review: Correction to the previous comment: the stripped identifiers were ReactRefreshRspackPlugin and @rspack/plugin-react-refresh. The requested coverage should prove that the Rspack 2 branch applies ReactRefreshRspackPlugin with Re.Pack's loader and injected entry path; webpack/Rspack 1 use the vendored runtime without evaluating @rspack/plugin-react-refresh; and Rspack 2 reports the intended actionable error when that optional peer is absent.

packages/repack/package.json
```diff
@@ -66,6 +67,7 @@
   "peerDependencies": {
     "@module-federation/enhanced": ">=0.6.10",
     "@rspack/core": ">=1",
+    "@rspack/plugin-react-refresh": "^2.0.0",
```
dannyhw, line 70: codex review: Correction to the previous comment: the stripped package name was @rspack/plugin-react-refresh. This PR makes @rspack/plugin-react-refresh an optional peer, but the current latest docs still say the plugin is shipped with Re.Pack and configured automatically. Please update the user-facing docs/install guidance in this PR or in the same stack so the package contract and docs agree.

dannyhw:
to be revisited for v6

PR #1403, opened 2026-07-06
chore: per-major tester apps for the rspack 2 dual-support window

dannyhw:
## Summary

Gives each Rspack major its own app-level validation surface for the dual-support window (part of the Rspack 2 stack; see the design doc in `agent_context/rspackv2-jul2026/design.md`).

### `apps/tester-app` is now the Rspack 2 example

In-workspace apps always run repack's `@rspack/core` devDependency major regardless of their own pin: `packages/repack` does a bare `import '@rspack/core'`, and since apps link repack as a workspace symlink, resolution walks up from `packages/repack` and finds its devDep (v2) first. Published installs are unaffected (`@rspack/core` is a peer there). This is covered in the design doc; the shipped code deliberately carries no monorepo-aware resolution workaround.

So tester-app's manifest now matches what actually runs: `@rspack/core` and `@swc/helpers` pin the workspace default catalog, plus `@rspack/plugin-react-refresh: catalog:` and a direct `react-refresh@^0.18.0` (pnpm-strict layouts don't hoist repack's copy into the official plugin's resolution scope). Nothing else changes — the full feature surface (assets matrix incl. remote, async/remote chunks, mini-apps, nativewind, reanimated, webpack lane) stays where it always was.

### `apps/tester-app-rspack1` is the standalone Rspack 1 lane

A standalone app *outside* the pnpm workspace (negation glob in the root `pnpm-workspace.yaml` + its own workspace root), with its own `node_modules` and repack installed from a packed tarball (`pnpm run pack-repack`). Because of the devDep shadowing above, this is the only way to genuinely run Rspack 1 against the repo — resolution behaves exactly like a real user project. Minimal own src (async local chunk + HMR target), intentionally not shared with tester-app (shared src across the workspace boundary would resolve two copies of react).

Retires when Rspack 1 support ends at the next major.

### Default catalog flip

The default catalog moves to `@rspack/core: ^2.1.2`, `@swc/helpers: ^0.5.23`, `@rspack/plugin-react-refresh: ^2.0.2`. The old v1 default was cosmetic — every workspace consumer already ran v2 via repack's devDep — so the workspace now says what it runs. Both federation apps declare the refresh plugin like any real v2 app.

Known cosmetic wart: `postcss-loader@8` declares peer `@rspack/core@"0.x || 1.x"`, so there's one unmet-peer install warning on `@rspack/core@2`.

### Integration-test fixes travel with the flip

The flip puts `tests/integration`'s rspack lane on true v2 for the first time, which surfaced three **output-format** changes (no behavioral regressions):

1. Custom runtime-module banners lost the `webpack/runtime/` prefix (`// repack/polyfills` instead of `// webpack/runtime/repack/polyfills`).
2. Unminified module factories use shorthand method syntax (`721() { ... }`).
3. Runtime-module order vs MFv2's `embed_federation_runtime` flipped — the real invariant (polyfills execute before `__webpack_require__.x()`) holds under both majors.

Fixes in `NativeEntryPlugin.test.ts`: version-agnostic markers, an added Rspack 2 module-factory regex, the MFv2 order assertion reduced to the real invariant, and rspack-lane snapshots regenerated under 2.1.2. `normalizeBundle` also strips ANSI codes (Rspack 2 colorizes inlined diagnostics when the env enables color) and URL-encoded machine paths (MF's `embed_federation_runtime` inlines a `data:` URI embedding the absolute repo root), so snapshots stay portable across local/CI environments.

No changeset — private apps and test-only changes.

> **Stack note:** the base `feat/rspack-2-integration` is a temporary integration branch (PR 2c + #1399/#1401/#1402 merged). This PR will be retargeted to `main` once those land.

## Validation

All run on this branch:

- `pnpm install` — regenerates the root lockfile; `apps/tester-app-rspack1` is **not** picked up as a workspace project (verified via `pnpm ls -r`); only the known `postcss-loader` unmet-peer warning
- `pnpm turbo run typecheck test --force` — 17/17 tasks green (integration, metro-compat, resolver-cases, tester-app bundle tests on Rspack 2.1.2 + webpack 5.105.3)
- `FORCE_COLOR=3 pnpm test` in `tests/integration` — 58/58, snapshots unchanged (env-dependent ANSI content confirmed normalized)
- `packages/repack`: `pnpm test` and `pnpm test:rspack1` — 282/282 under both majors
- `pnpm lint:ci` — clean
- `apps/tester-app`: `pnpm bundle:ios` / `pnpm bundle:android` → "(Rspack 2.1.2) compiled successfully"; served dev bundle contains the **official** `@rspack/plugin-react-refresh@2.0.2` client runtime (92 refs) and **zero** vendored refs
- `apps/tester-app-rspack1` (own workspace): `pnpm run pack-repack && pnpm install && pnpm bundle:ios && pnpm bundle:android` → "(Rspack 1.7.12) compiled successfully"; served dev bundle contains the **vendored** refresh runtime (`vendor/react-refresh`, 9 refs) and **zero** official-plugin refs
- `apps/tester-federation`: `pnpm bundle` → all four bundles (host+mini × ios+android) "(Rspack 2.1.2) compiled successfully"; `USE_WEBPACK=1 pnpm bundle` → all four "(webpack 5.105.3) compiled successfully"

On-device validation (Pixel 7 physical device + iPhone 17 simulator, incl. HMR and the interactive remote-chunk / mini-app / assets flows) was already performed on the reference branch — see #1393.


apps/tester-federation/package.json
```diff
@@ -36,6 +36,7 @@
     "@react-native/typescript-config": "catalog:testers",
     "@rsdoctor/rspack-plugin": "^1.5.2",
     "@rspack/core": "catalog:",
+    "@rspack/plugin-react-refresh": "catalog:",
```
dannyhw, line 39: codex review: This app now declares @rspack/plugin-react-refresh, but it does not declare the plugin's non-optional react-refresh peer. tester-app adds react-refresh for the same pnpm-strict reason, so this package is relying on another workspace importer to satisfy the peer. Please add react-refresh here, matching tester-app, so this app's manifest is self-contained.

apps/tester-federation-v2/package.json
```diff
@@ -31,6 +31,7 @@
     "@react-native-community/cli-platform-android": "catalog:testers",
     "@rsdoctor/rspack-plugin": "^1.5.2",
     "@rspack/core": "catalog:",
+    "@rspack/plugin-react-refresh": "catalog:",
```
dannyhw, line 34: codex review: This app now declares @rspack/plugin-react-refresh, but it does not declare the plugin's non-optional react-refresh peer. tester-app adds react-refresh for the same pnpm-strict reason, so this package is relying on another workspace importer to satisfy the peer. Please add react-refresh here, matching tester-app, so this app's manifest is self-contained.

apps/tester-app-rspack1/README.md
```diff
@@ -0,0 +1,49 @@
...
+surface: dev server, HMR/React Refresh (the vendored-runtime path under v1),
+async local chunks via ScriptManager, and production bundling. Full feature
+coverage (NativeWind, Reanimated, remote chunks/mini-apps, assets matrix)
+lives in [`tester-app`](../tester-app/) - the **Rspack 2** example.
+
+## Setup
+
+```sh
+pnpm run pack-repack   # builds packages/repack and packs it to ./callstack-repack.tgz
+pnpm install           # standalone install (own lockfile, gitignored)
```
dannyhw, line 44: codex review: This setup step is not CI-reproducible as written because this app's pnpm-lock.yaml is gitignored, and pnpm defaults to frozen-lockfile in CI. If this app is meant to be a validation lane, please either commit its standalone lockfile or document/use pnpm install --no-frozen-lockfile for CI/manual validation.

tests/integration/src/plugins/NativeEntryPlugin.test.ts
```diff
@@ -394,25 +420,24 @@ describe('NativeEntryPlugin', () => {
...
-          //   2. repack/polyfills IIFE executes (polyfills loaded immediately)
-          //   3. __webpack_require__.x() called → MF init → original startup (polyfills are cache hits)
+          // Rspack MF v2 wraps startup via embed_federation_runtime: it saves
+          // the original __webpack_require__.x and wraps it, while the
+          // repack/polyfills IIFE executes during runtime-section evaluation.
+          // The relative order of the two runtime modules differs between
+          // Rspack majors (polyfills first on 2.x, embed first on 1.x) - the
+          // invariant is that polyfills execute before __webpack_require__.x()
+          // is invoked (MF init → original startup, polyfills are cache hits).
           expect(code).toContain('embed_federation_runtime');
```
dannyhw: codex review: This only proves the embed runtime string exists, not that it wraps startup before __webpack_require__.x() is invoked. A regression that moves embed_federation_runtime after startup would still pass. Please also assert embed_federation_runtime appears before the startup call, while keeping the polyfill-vs-embed order flexible across Rspack majors.

apps/tester-app-rspack1/android/gradle.properties
```diff
@@ -0,0 +1,55 @@
...
+
+# Uncomment the line below to build React Native from source.
+#react.buildFromSource=true
+
+# Version of Android NDK to build against.
+#ANDROID_NDK_VERSION=26.1.10909125
+
+# Version of Kotlin to build against.
+#KOTLIN_VERSION=1.8.22
+
```
dannyhw: codex review: This added blank line contains trailing whitespace. git diff --check also reports trailing whitespace in regenerated snapshots; even if snapshot whitespace is intentional generated output, this Gradle file line should be empty.

apps/tester-federation/package.json
dannyhw: Added react-refresh ^0.18.0 to devDependencies, matching tester-app - e422002e. With autoInstallPeers: false the plugin's non-optional peer does need to be declared here for the manifest to be self-contained.

apps/tester-federation-v2/package.json
dannyhw: Added react-refresh ^0.18.0 to devDependencies, matching tester-app - e422002e (same fix as tester-federation).

apps/tester-app-rspack1/README.md
dannyhw: Committing the lockfile is not an option here: it pins file:./callstack-repack.tgz, whose contents (and integrity) change on every pack-repack, so a tracked lockfile would be permanently stale - that is why install artifacts are gitignored (see the design doc, agent_context/rspackv2-jul2026/design.md, and #1393). Two clarifications on the CI concern: pnpm 10 proceeds when the lockfile is *absent* even with CI=true (verified locally), so the fresh-clone path already worked; the real hazard is a leftover lockfile that predates a re-pack. Fixed by adding a local .npmrc with frozen-lockfile=false plus a README note, so plain pnpm install behaves the same everywhere - 9c5f2199. Verified with CI=true pnpm install followed by pnpm bundle:android (compiles with Rspack 1.7.12).

tests/integration/src/plugins/NativeEntryPlugin.test.ts
dannyhw: Good catch - containment alone would not catch the embed runtime moving after startup. Now asserting each chain independently against the startup call: expectBundleOrder(['embed_federation_runtime', '__webpack_require__.x()']) and expectBundleOrder(['repack/polyfills', '__webpack_require__.x()']), which keeps the polyfill-vs-embed relative order flexible across majors (the embed wrapper body contains __webpack_require__.x assignments but never the literal call, so the marker only matches the actual startup invocation) - 5e16a4d6. Integration suite passes for both bundler projects.

apps/tester-app-rspack1/android/gradle.properties
dannyhw: Fixed the gradle.properties line - ca2152b8. The remaining git diff --check hits are inside snapshot strings in tests/integration/__snapshots__: they are verbatim compiler output (webpack/rspack emit trailing spaces in some runtime templates), so trimming them would make the snapshots diverge from the actual bundles.

apps/tester-app-rspack1/README.md
dannyhw: codex review: Correction on my original CI claim: re-checking plain CI=true pnpm install shows pnpm proceeds when the lockfile is absent; only an explicit --frozen-lockfile fails. The added local .npmrc/README note still addresses the stale-lockfile hazard for this tarball workflow, so I consider this resolved.

apps/tester-app-rspack1/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "tester-app-rspack1",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "pack-repack": "cd ../../packages/repack && pnpm build && pnpm pack --pack-destination ../../apps/tester-app-rspack1/.repack-pack && cd ../../apps/tester-app-rspack1 && mv .repack-pack/*.tgz callstack-repack.tgz && rmdir .repack-pack",
```
dannyhw, line 6: codex review: This standalone lane only packs the local @callstack/repack tarball. The packed manifest rewrites Re.Pack's workspace @callstack/repack-dev-server dependency to the plain registry version (5.2.5 in this branch), so `pnpm start`/HMR validation here exercises the published dev-server package rather than the repo's local `packages/dev-server` code. If this app is meant to validate the repo's dev-server/HMR path during the dual-support window, please also pack/install the local dev-server tarball; otherwise narrow the README/PR claim so this lane is explicitly "packed Re.Pack against published runtime deps."

dannyhw (commented):
codex review: Added one inline finding about the standalone Rspack 1 app resolving @callstack/repack-dev-server from the registry during dev-server/HMR validation.

apps/tester-app-rspack1/package.json
dannyhw: Fixed in ce62090d - README now states this lane runs the locally packed Re.Pack against published runtime deps (pnpm pack rewrites the workspace dev-server dependency to the published version), and that dev-server changes need re-validation via tester-app (or an additional dev-server pack step). Taking the doc-narrowing option for now since the published dev-server is currently code-identical to the workspace one; we can add a dev-server pack step if that changes during the support window.

dannyhw:
Combined into #1404 as the end-to-end Rspack 1/Rspack 2 validation review unit. This PR remains available for its detailed tester-app and device-validation history.

PR #1406, opened 2026-07-07
chore: remove image-size

ilteoood:
### Summary

The image-size package is now read-only on GitHub, and on codeberg has been a while since its last commit.
This PR is to migrate to sharp.


dannyhw (commented):
Firstly thanks for taking the time for opening this pr, appreciate the interest in repack.

I'm concerned that sharp might be a large dependency to add for just image size check? 

1. what do we gain by moving to sharp 
2. is there some issue with image-size in its current form?

ilteoood:
Hey @dannyhw, to both I would answer with the following [Snyk vulnerability](https://security.snyk.io/vuln/SNYK-JS-IMAGESIZE-9634164)

dannyhw:
@ilteoood is that not addressed?
As in if we updated to the latest version on the package it also solves the vuln right?

ilteoood:
> @ilteoood is that not addressed?
> As in if we updated to the latest version on the package it also solves the vuln right?

Yes, exactly!

dannyhw:
Im not opposed to a change like this I just think there needs to be a measurable improvement. Of course we need to address the vuln but simpler is usually better i.e whats the downside of just updating the image size version to a non vulnerable one.

I also see that sharp might require us to bump the minimum node version (which may be due but not in a patch)

dannyhw:
@ilteoood i will make a canary soon could you check it on your project?

dannyhw:
After thinking more about this I have a few problems with replacing image size with sharp:

- Doesn't cover the full list of image types that image size does
- node >=20.9.0 is required (technically a breaking change)
- is a much larger dependency

I would prefer to find an alternative or just vendor a fixed version of image-size if necessary
For now I will make sure to update the package and address any security reports.

PR #1417, opened 2026-07-27
fix: prevent public key validation ReDoS

dannyhw:
## Summary

- simplify public-key delimiter validation to remove ambiguous overlapping regex quantifiers
- add a regression test covering a large malformed near-match
- exercise resolver validation with an existing real 4096-bit RSA fixture, CRLF PEM formatting, a truncated key, and a genuine PKCS#1 variant
- add a patch changeset for `@callstack/repack`

## Root cause

The existing `\s*[\s\S]+?\s*` expression allowed the JavaScript regex engine to partition an internal whitespace run in many ways before failing to find the PEM footer. Because `ScriptLocator.publicKey` can come from resolver-provided runtime metadata and had no length bound, malformed input could block the JavaScript thread with polynomial backtracking.

This addresses [code-scanning alert #10](https://github.com/callstack/repack/security/code-scanning/10) (`js/polynomial-redos`).

## Compatibility and impact

The replacement is language-equivalent to the previous validator: both accept any non-empty content between the existing `BEGIN PUBLIC KEY` and `END PUBLIC KEY` markers after trimming. The public API, normalization, errors, native verification, and accepted input set are unchanged. Matching malformed near-misses is now linear.

No migration or user action is required.

## Test plan

- `pnpm --filter @callstack/repack test -- --watchman=false ScriptManager.test.ts`
- `pnpm --filter @callstack/repack typecheck`
- `pnpm biome check packages/repack/src/modules/ScriptManager/normalizePublicKey.ts packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts`
- commit hook: repository-wide Biome check (457 files)

### DIFF
--- a/.changeset/calm-keys-verify.md
+++ b/.changeset/calm-keys-verify.md
@@ -1,0 +1,5 @@
+---
+"@callstack/repack": patch
+---
+
+Prevent excessive regular-expression backtracking when validating per-script public keys.
--- a/packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
+++ b/packages/repack/src/modules/ScriptManager/__tests__/ScriptManager.test.ts
@@ -1,3 +1,6 @@
+import { createPublicKey } from 'node:crypto';
+import fs from 'node:fs';
+import path from 'node:path';
 import NativeScriptManager, {
   type NormalizedScriptLocator,
 } from '../NativeScriptManager.js';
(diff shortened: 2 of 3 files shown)


dannyhw:
@MikitasK can you take a look?

MikitasK (approved):
LGTM ✅ 
reviewed these changes against #826 & #1382 and haven't found any regressions in multiple-public-keys contract 👌

Issue #1421, opened 2026-07-28
[v6] Unify webpack and Rspack start/bundle commands


dannyhw:
## Motivation

Re.Pack ships two parallel command trees — `packages/repack/src/commands/rspack/` and `packages/repack/src/commands/webpack/` — each with its own `start`, `bundle`, `Compiler` and command registry, exposed as separate entrypoints (`@callstack/repack/commands/rspack` and `@callstack/repack/commands/webpack`). The config pipeline is already shared in `commands/common/`, and plugins/loaders are shared with a few `isRspackCompiler()` branches. The commands are the last unshared piece — so features land on one side only, and behaviour has drifted in user-visible ways.

Part of #1414 (Roadmap to V6), which lists "unify webpack and rspack start/bundle commands".

## Duplication

- `rspack/index.ts` and `webpack/index.ts` are byte-identical. Both also register `webpack-bundle` / `webpack-start` aliases, so the rspack entrypoint installs commands named `webpack-*`.
- `bundle.ts` and `start.ts` are copy-paste apart from the bundler literals, `experiments.cache` vs `cache`, the Rspack-only `setupRspackEnvironment` / `RSPACK_PROFILE` blocks, and the compiler/dev-server wiring (see below).
- `CompilerAsset` is defined in both `rspack/types.ts` and `webpack/types.ts`.
- `defineRspackConfig` / `defineWebpackConfig` are the same identity function with parallel overload chains and type families.
- `templates/rspack.config.mjs` and `templates/webpack.config.mjs` differ only in doc comments, the `define*Config` call, and `parallel: true` on the minimizer (same for the `.cjs` pair). `packages/init` downloads one or the other from the repo at init time (`createBundlerConfig.ts`).

## Behavioural divergence

The real cost — the two commands behave differently:

- **Execution model.** Rspack: one in-process `MultiCompiler`, watching all platforms eagerly from `compiler.start()`. Webpack: a `worker_threads` worker spawned lazily per platform on the first asset request. Most of the rest follows from this.
- **`api.getPlatforms()`** returns all configured platforms on Rspack, but `Object.keys(compiler.workers)` on webpack — an empty list until a platform's first bundle request.
- **`getSource`:** the Rspack compiler handles absolute paths (`path.isAbsolute` check); the webpack one unconditionally joins with `rootDir`, breaking symbolication for absolute frame paths.
[...]

PR #1424, opened 2026-07-29
refactor(commands): unify webpack and Rspack commands

jbroma:
## Summary

- add a single `@callstack/repack/commands` entry point with automatic bundler detection and a validated `--bundler` override
- consolidate the duplicate `start` and `bundle` command orchestration and dev-server delegate while preserving the separate Rspack and webpack compiler implementations
- retain the bundler-specific command entry points as deprecated compatibility overrides
- reject pending webpack asset requests when compilation fails instead of leaving HTTP requests hanging
- update Re.Pack Init, tester apps, CLI documentation, tests, and Changesets for the unified surface

## Why

The command registries and most command orchestration were duplicated even though both bundlers already shared the config and dev-server layers. This gives users one stable command entry point while keeping each bundler's deliberately different execution model behind a small shared compiler interface.

Related to #1421. This PR intentionally does not merge compiler internals, `defineConfig` helpers, or bundler configuration templates: those remain bundler-specific where their types and runtime behavior differ.

## Validation

- full repository JavaScript regression suite passed
- `@callstack/repack`: 31 Jest suites, 290 tests, and 22 snapshots passed
- Re.Pack TypeScript and Biome checks passed
- production documentation build passed
- iOS simulator runtime smoke tests passed for `tester-app`, `tester-federation`, and `tester-federation-v2`, including async chunks, miniapp lifecycle, and both Module Federation gallery flows
- direct iOS simulator target builds, installs, and launches passed for all three apps; the standard React Native CLI destination lookup remains affected by a local Xcode 26.6/runtime discovery issue


thymikee:
he's back!

bartekkrok (approved):

MikitasK (approved):
LGTM ✅

dannyhw (approved):

dannyhw:
@jbroma I think there might be some small thing to check with the tester apps using USE_WEBPACK for example in `apps/tester-federation-v2/config.host-app.mts`, maybe that can be removed/adjusted since it will probably fallback to rspack?

packages/repack/commands/webpack.js
dannyhw, line 1: if a user has the webpack commands and specifies --bundler=rspack what happens? Should we care to handle it? In this case it would be overruled by the `createBoundCommands` function right?

jbroma: added explicit warning for that

PR #1425, opened 2026-07-30
feat(expo): add first-class Re.Pack integration

### LINKED ISSUE #1396: Re.pack and its Module Federation / Micro-frontends is not working with expo
### Description

I'm building a multi-tenant React Native platform using Expo, where different customers can have different feature sets.

The ideal architecture would allow downloading or loading feature modules at runtime (similar to Module Federation / Micro Frontends) without rebuilding the entire application.

After reading the documentation, I noticed that Re.Pack currently doesn't officially support Expo.

Official support (or an experimental guide) for integrating Re.Pack with modern Expo projects using:

- Expo SDK 53/54+
- Expo Router
- Expo Development Builds (not Expo Go)
- Android & iOS
- React Native New Architecture

Even if Expo Go cannot be supported due to Metro, it would be extremely valuable if Re.Pack could work with custom development builds and production Expo applications.

## Use case

- One host application serves multiple tenants.
- Some features should be downloaded or enabled independently.
- We'd like to avoid publishing a new application whenever a new module is introduced.

Module Federation seems like an ideal solution for this use case.

## Questions

1. Is Expo support on the roadmap?
2. Is there any technical limitation preventing Re.Pack from working with Expo Development Builds?
3. Has anyone successfully integrated Re.Pack with Expo Router?
4. Are there any recommended alternatives for achieving runtime module loading in Expo?

## Additional context

The current documentation states:

> "Officially, Re.Pack doesn't support Expo."

However, since Expo now supports custom development builds and prebuild workflows, it seems that supporting Expo (outside Expo Go) could unlock many enterprise use cases.

I'd appreciate any guidance on whether this is feasible.

### LINKED ISSUE #1413: [TRACKING] Enhanced Expo support
lets use this as a place to discuss and outline the path toward better expo support

The first steps should be ensuring:

- [ ] normal repack functionality
- [ ] easy setup via cli command
- [ ] expo module/plugin to simplify setup of native dependency (script manager)
- [ ] expo features like OTA working as expected.

Following this, the nice to haves
- [ ] module federation support
- [ ] supporting metro module federation where possible
- [ ] assess if expo updates can work with repack or offer examples of alternative setups
- [ ] benchmark metro vs repack for expo projects

Please offer your thoughts or suggestions in the comments below

whydidoo:
## Summary

This PR introduces first-class Expo support for Re.Pack through a new `@callstack/repack-expo` package.

It allows Expo SDK 56 applications using prebuild/CNG to use Rspack and the existing Re.Pack runtime, including ScriptManager and Module Federation v2, without changing the core Re.Pack packages.

The package remains private while we validate the initial integration contract.

## What’s included

- An Expo Config Plugin that configures generated iOS and Android projects for Re.Pack.
- An `ExpoPlugin` for Rspack that owns the required Expo/Re.Pack defaults:
  - Expo Router and plain entry resolution
  - Babel configuration
  - React Native and monorepo module resolution
  - images, density variants, fonts and XML assets
  - `EXPO_PUBLIC_*` environment variables
  - source maps and symbolication
  - packaged local chunks through ScriptManager
- CLI commands for initial setup and diagnostics:
  - `npx @callstack/repack-expo init`
  - `npx @callstack/repack-expo doctor`
- Development, native Release and EAS Build integration.
- Expo Router support without importing Metro runtime behavior.
- Module Federation v2 support using the existing public Re.Pack plugin and ScriptManager.
- Expo host and Expo widget validation applications for iOS and Android.

## Module Federation v2

Module Federation remains explicit and application-owned, matching regular Re.Pack behavior.

The implementation supports:

- Expo host → Expo widget
- Expo host → regular Re.Pack MF v2 widget
- packaged local `React.lazy()` chunks alongside remote widgets
- remotely deployed widget chunks, assets and fonts
- platform-specific iOS and Android manifests
- the existing ScriptManager resolver, storage and security mechanisms

Remote widgets cannot provide native dependencies. Native modules must already be installed and linked in the host application.

## Native integration

`@callstack/repack-expo init` only updates application-owned configuration.

Native changes are applied through the Expo Config Plugin during `expo prebuild`. The integration does not require users to maintain custom Swift, Kotlin, Gradle or Xcode changes manually.

Development applications are launched with:

```sh
npm run repack:start
npm run repack:ios
# or
npm run repack:android

### CODE REVIEW

**Comment by dannyhw** [comment]
@whydidoo can you check the merge conflicts?

File: apps/tester-expo/package.json [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+    "@module-federation/enhanced": "2.1.0",
+    "expo": "~56.0.5",
+    "expo-asset": "~56.0.21",
+    "expo-constants": "~56.0.16",
+    "expo-font": "~56.0.7",
+    "expo-linking": "~56.0.12",
+    "expo-router": "~56.2.7",
+    "expo-status-bar": "~56.0.4",
+    "react": "19.2.3",
+    "react-native": "0.85.3",
```
dannyhw: I recently updated our react native version to "0.86.0"
also note that we have some of our package versions in catalogs in pnpm workspace
would be good to keep react native versions in sync across projects I think

whydidoo: Expo SDK 57 currently expects React Native 0.86.2 and React 19.2.3.
I’ve kept both Expo examples on the same Expo-compatible versions rather than forcing the default catalog versions. We could add a dedicated Expo SDK 57 catalog later to keep those versions centralized as well.

dannyhw:
lets try and align the Expo package and examples with the workspace catalogs

At the moment the new Expo projects use direct pins for React, React Native, Community CLI packages, the React Native Babel preset, React types, TypeScript, and Module Federation, while existing testers use the shared catalog entries. In particular, the Expo apps use React Native 0.86.2 while the catalog is 0.86.0; if 0.86.2 is the Expo 57-compatible version, please update the shared React Native catalog to 0.86.2 and have the relevant examples consume it through the catalog.

The Expo SDK packages themselves can live in a dedicated Expo 57 catalog, shared by both Expo fixtures, so their versions do not drift. The same applies to any Expo-specific React, CLI, or Babel compatibility pins that genuinely cannot use the default catalog.

regarding typescript version I have a separate pr that updates typescript to v7 which we could merge before this and then update those expo projects to be on v7 as well

packages/repack-expo/src/rspack/router/configureExpoRouterEntry.ts
```diff
@@ -0,0 +1,22 @@
+import type { ResolvedExpoEntry } from '../entry/resolveExpoEntry.js';
+
+const EXPO_ROUTER_ENTRY = 'expo-router/entry';
+
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+export function configureExpoRouterEntry(
+  rules: unknown[],
```
bartekkrok:
```suggestion
  rules: RuleSetRule[],
```

packages/repack-expo/src/rspack/ExpoPlugin.ts
```diff
@@ -0,0 +1,186 @@
...
+    const routerRoot = resolveRouterRoot(
+      resolvedEntry.projectRoot,
+      this.options.routerRoot
+    );
+    const publicEnvironment = configureExpoPublicEnvironment(
+      compiler,
+      resolvedEntry.projectRoot
+    );
+    const caller = createExpoBabelCaller({
+      isDev: compiler.options.mode === 'development',
```
bartekkrok, line 134:
```suggestion
      isDev: compiler.options.mode !== 'production',
```
Since `export type Mode = "development" | "production" | "none";` when someone will set `none` then will take production behaviour. Not sure if it's in purpose

packages/repack-expo/src/config/android.ts
```diff
@@ -0,0 +1,117 @@
...
+import {
+  assertUniqueAnchor,
+  hasIntactGeneratedSection,
+  replaceLineWithGeneratedSection,
+} from './generated.js';
+import type { RepackExpoPluginOptions } from './options.js';
+
+const ANDROID_MAIN_MODULE_TAG = 'repack-expo-android-main-module';
+const ANDROID_BUNDLE_COMMAND_TAG = 'repack-expo-android-bundle-command';
+const EXPO_CLI_BLOCK = `    // Use Expo CLI to bundle the app, this ensures the Metro config
```
bartekkrok: I think that instead of matching the entire block character-by-character we could use RegExp with tolerance for whitespace. Both here (EXPO_CLI_BLOCK) and in ios.ts (EXPO_CLI_DEFAULTS)

packages/repack-expo/src/cli/project.ts
```diff
@@ -0,0 +1,343 @@
...
+    const locked = lockfiles.find(([lockfile]) =>
+      fs.existsSync(path.join(current, lockfile))
+    )?.[1];
+    if (locked) return locked;
+
+    const parent = path.dirname(current);
+    if (
+      parent === current ||
+      fs.existsSync(path.join(current, '.git')) ||
+      fs.existsSync(path.join(current, 'pnpm-workspace.yaml'))
```
bartekkrok:
```suggestion
      fs.existsSync(path.join(current, 'pnpm-workspace.yaml')) ||
      (currentPackageJson && 'workspaces' in currentPackageJson)
```

packages/repack-expo/src/rspack/ExpoPlugin.ts
whydidoo: @bartekkrok, Thanks for the comment. This is intentional and matches the existing RepackPlugin behavior, where __DEV__ is enabled only when [compiler.options.mode === 'development'](https://github.com/callstack/repack/blob/main/packages/repack/src/plugins/RepackPlugin.ts#L107). Treating none as dev here would make the Expo Babel caller inconsistent with the rest of Re.Pack.

packages/repack-expo/src/rspack/router/configureExpoRouterEntry.ts
whydidoo: done

packages/repack-expo/src/config/android.ts
whydidoo: done

packages/repack-expo/src/cli/project.ts
whydidoo: done

apps/tester-expo/package.json
whydidoo: done

bartekkrok (approved):

packages/repack-expo/src/cli/init.ts
```diff
@@ -0,0 +1,507 @@
...
+  if (commonJs) {
+    return `const path = require('node:path');
+const { ExpoPlugin } = require('@callstack/repack-expo/rspack');
+
+const projectRoot = path.resolve(__dirname, ${JSON.stringify(rootFromConfig)});
+
+module.exports = (env) => createConfig(env, projectRoot);
+
+function createConfig(env, projectRoot) {
+  const { mode = 'development', platform = process.env.PLATFORM, devServer } = env;
```
MikitasK: could we pass `env.entry` to ExpoPlugin in both generated config variants?
the thing is that native release builds provide Config Plugin’s custom entry through `env.entry`, but generated config currently drops it. ExpoPlugin therefore bundles default entry from package.json instead of custom `entry`

```
const {
  entry,
  mode = 'development',
  platform = process.env.PLATFORM,
  devServer,
} = env;

...

plugins: [new ExpoPlugin({ entry, platform })],
```

packages/repack-expo/src/cli/init.ts
```diff
@@ -0,0 +1,507 @@
...
+import { fileURLToPath } from 'node:url';
+import { ExpoPlugin } from '@callstack/repack-expo/rspack';
+
+const projectRoot = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  ${JSON.stringify(rootFromConfig)}
+);
+
+export default (env) => {
+  const { mode = 'development', platform = process.env.PLATFORM, devServer } = env;
```
MikitasK: [same](https://github.com/callstack/repack/pull/1425/changes#r3735752329) suggestion here

packages/repack-expo/src/rspack/environment/configureExpoPublicEnvironment.ts
```diff
@@ -0,0 +1,51 @@
...
+  getExpoEnvironmentFiles,
+} from './expoPublicEnvironment.js';
+
+const PLUGIN_NAME = 'RepackExpoPublicEnvironmentPlugin';
+
+function addEnvironmentDigestToPersistentCache(
+  compiler: Compiler,
+  digest: string
+): void {
+  const cache = compiler.options.experiments?.cache;
```
MikitasK: could we read persistent-cache configuration from both `compiler.options.cache` & `compiler.options.experiments?.cache`? 
rspack 2 moved `experiments.cache` to the top-level cache option ([source](https://www.rspack.dev/guide/migration/rspack_1.x#other-experimental-configuration-changes)), so expo env digest is currently not added to rspack 2 cache version

```suggestion
  const cache = compiler.options.cache || compiler.options.experiments?.cache;
```

packages/repack-expo/src/cli/rspackConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+  return 'none';
+}
+
+export function isRspackConfigCompatible(contents: string): boolean {
+  const federation = getRspackFederationUsage(contents);
+  return (
+    /@callstack\/repack-expo\/rspack/.test(contents) &&
+    /new\s+(?:\w+\.)?ExpoPlugin\s*\(/.test(contents) &&
+    !/new\s+(?:Repack\.)?RepackPlugin\s*\(/.test(contents) &&
```
MikitasK: [nit] could we ignore comments when checking for ExpoPlugin & RepackPlugin usage? for example, this config is currently reported as compatible even though the plugin is never instantiated:
```
import { ExpoPlugin } from '@callstack/repack-expo/rspack';
// new ExpoPlugin();
export default {};
```

could we strip comments before running these checks & add this examples as a regression test?

packages/repack-expo/src/cli/rspackConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+} {
+  const aliases = {
+    v1: new Set(['ModuleFederationPlugin', 'ModuleFederationPluginV1']),
+    v2: new Set(['ModuleFederationPluginV2']),
+  };
+  for (const match of source.matchAll(
+    /(ModuleFederationPlugin(?:V[12])?)\s*(?:as\s+|:\s*)([A-Za-z_$][\w$]*)/g
+  )) {
+    const version = match[1] === 'ModuleFederationPluginV2' ? 'v2' : 'v1';
+    aliases[version].add(match[2] as string);
```
MikitasK: [nit] could we ensure that aliased `ModuleFederationPluginV2` is still detected as `v2`?  here's the case:
```
import { ModuleFederationPluginV2 as ModuleFederationPlugin } from '@callstack/repack';
new ModuleFederationPlugin({});
```
this will be classified as `v1` in current implementation, which is incorrect. 
we can fix it like that:


```suggestion
    const importedName = match[1];
    const alias = match[2];
    const version = importedName === 'ModuleFederationPluginV2' ? 'v2' : 'v1';
    const otherVersion = version === 'v2' ? 'v1' : 'v2';
    aliases[otherVersion].delete(alias);
    aliases[version].add(alias);
```

MikitasK (commented):
excellent work overall 👏 👍
just 2 comments about forwarding custom Expo entry & supporting Rspack 2 cache invalidation, plus 2 non-blocking edge-cases related suggestions:

packages/repack-expo/src/cli/init.ts
whydidoo: Good point - `entry` needs to be forwarded here because `ExpoPlugin` owns the final compiler entry and prepends the ScriptManager runtime. Without forwarding env.entry, a CLI-provided `--entry-file` would be silently replaced by   `package.json#main`.

I’ve updated both generated CJS and ESM configs to pass entry to ExpoPlugin and added regression coverage.
package.json#main remains the default source of truth. The following config was only used locally to validate a custom native entry through expo prebuild and an iOS Release build:

```
app.json
[
  "@callstack/repack-expo",
  {
    "entry": "custom-entry.prototype.tsx"
  }
]
```
It is not required for normal usage and won’t be included in the example. The regular setup remains:

```
{
  "main": "expo-router/entry",
  "plugins": ["@callstack/repack-expo"]
}
```

It won't be part of the example. Normally the entry comes from `package.json#main`, and the plugin config stays simple:

```
"plugins": ["@callstack/repack-expo"]
```

The config plugin `entry` option is there for cases where the native entry intentionally differs from `package.json#main`. `--entry-file` is still available for a single bundle command.

packages/repack-expo/src/cli/init.ts
whydidoo: done

packages/repack-expo/src/rspack/environment/configureExpoPublicEnvironment.ts
whydidoo: done

packages/repack-expo/src/cli/rspackConfig.ts
whydidoo: done

packages/repack-expo/src/cli/rspackConfig.ts
whydidoo: done

MikitasK (approved):
LGTM 👍

packages/repack-expo/src/rspack/router/configureExpoRouterEntry.ts
```diff
@@ -0,0 +1,23 @@
...
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+export function configureExpoRouterEntry(
+  rules: RuleSetRules,
+  entry: ResolvedExpoEntry,
+  loaderPath: string
+): boolean {
+  if (entry.request !== EXPO_ROUTER_ENTRY) return false;
```
dannyhw: Should this resolve the entry path instead of comparing the request string?

Below here is ai reasoning for why it might be necessary. I'm not fully certain but wondering what you think.
___ 

In native Release builds the Expo template resolves `ENTRY_FILE` / `entryFile` to an absolute path (`expo/scripts/resolveAppEntry <root> <platform> absolute`), and the config plugin leaves that line in place. Re.Pack forwards it unchanged into `env.entry`, and the `init`-generated config passes it on via `new ExpoPlugin({ entry, platform })`. `request` is then `/…/node_modules/expo-router/entry.js`, this check returns false, and the raw `entry-classic.js` gets bundled together with `@expo/metro-runtime`.

I checked with a production compile of a small Expo Router fixture: with `new ExpoPlugin({ platform })` (the tester-expo shape) the bootstrap loader is installed and `@expo/metro-runtime` is absent; with `new ExpoPlugin({ platform, entry: <absolute path> })` the loader is skipped and it is bundled. The build still succeeds, so nothing visibly breaks today, but dev and Release then boot through different code and the Metro-free path the README and tests describe isn't taken on the documented setup.

Something like this would cover it:

```ts
function isExpoRouterEntry(entry: ResolvedExpoEntry): boolean {
  if (entry.request === EXPO_ROUTER_ENTRY) return true;
  try {
    const routerEntryPath = fs.realpathSync(
      require.resolve(EXPO_ROUTER_ENTRY, { paths: [entry.projectRoot] })
    );
    return entry.entryPath === routerEntryPath;
  } catch {
    return false;
  }
}
```

`realpathSync` on both sides matters for pnpm's symlinked layout (`resolveExpoEntry` already realpaths `entryPath`). Alternatively `resolveExpoEntry` could normalise `request` back to `expo-router/entry` when the resolved file matches, and this check stays as is. Either way a test that passes the absolute entry would be good to have.

packages/repack-expo/src/cli/project.ts
```diff
@@ -0,0 +1,344 @@
...
+import type {
+  ConfigPluginRegistration,
+  Diagnostic,
+  JsonObject,
+  PackageJson,
+  PackageManager,
+} from './types.js';
+
+export const CONFIG_PLUGIN = '@callstack/repack-expo';
+export const RSPACK_COMMANDS = '@callstack/repack/commands/rspack';
```
dannyhw: Since #1424 this entry point is a deprecated shim that prints a warning on every CLI invocation (`react-native webpack-start` in tester-expo shows it as the first line). The other testers and repack-init were switched to `@callstack/repack/commands`, which detects Rspack from `rspack.config.*` and registers the same command names, so `init`, `doctor` and both Expo apps' `react-native.config.js` can point there instead. Nothing else needs to change; no test pins the old string.

packages/repack-expo/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+) => {
+  validateExpoConfig(config);
+  const normalizedOptions = normalizeConfigPluginOptions(options);
+  return withAndroidRepack(
+    withIosRepack(config, normalizedOptions),
+    normalizedOptions
+  );
+};
+
+export = createRunOncePlugin(withRepackExpo, '@callstack/repack-expo', '0.0.0');
```
dannyhw: Could this read the version from `package.json` instead of the `'0.0.0'` literal? Expo only records it in the plugin history, so it is harmless today, but `test/package-boundary.test.cjs:13` and `:40` also assert the literal, so a version change will fail those tests. Comparing against `packageJson.version` in the tests would avoid that.

dannyhw:
`changeset status` on this branch reports that the next release would bump `tester-expo` and `tester-expo-widget` to 0.0.2 (they depend on `@callstack/repack` via `workspace:*`). The existing testers avoid this by being listed in `ignore` in `.changeset/config.json`. Could you add `tester-expo`, `tester-expo-widget` and `@callstack/repack-expo` there while the package is private? With all three ignored, `changeset status` reports no bumps.

apps/tester-expo/app.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "android": {
+      "package": "com.callstack.repackexpo"
+    },
+    "extra": {
+      "eas": {
+        "projectId": "49a167cf-9412-446e-ba71-0c07c816ed93"
+      }
+    },
+    "owner": "wuskas"
```
dannyhw, line 27: `owner` and `extra.eas.projectId` point at a personal Expo account, so anyone else running EAS against this app gets an ownership error. We need to figure out what these should be for a Callstack-owned project; I'll take care of that, so no change needed from your side here.

packages/repack-expo/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "node": ">=20.19"
+  },
+  "dependencies": {
+    "@callstack/repack-plugin-expo-modules": "workspace:^",
+    "@expo/env": "~2.3.1"
+  },
+  "peerDependencies": {
+    "@callstack/repack": "workspace:^",
+    "@rspack/core": ">=1",
+    "expo": ">=56"
```
dannyhw, line 55: The tests need `babel-preset-expo`, `react-native`, `expo-router`, `@module-federation/enhanced` and `source-map`, but none of them resolve from this package; the fixtures get them by symlinking `apps/tester-expo/node_modules`, `packages/repack/node_modules/@module-federation/enhanced` and `packages/dev-server`'s `source-map` (`public-environment.test.cjs:72`, `repack-plugin-options.test.cjs:55`, `module-federation-source-map.test.cjs:14,41`, `expo-plugin.test.cjs:174`). That ties this package's suite to the tester app's install and to other packages' private `node_modules`. Could these be declared as devDependencies here (catalog / catalog:expo57) and the fixtures populated by resolving from this package's own root instead?

packages/repack-expo/src/cli/project.ts
```diff
@@ -0,0 +1,344 @@
...
+    const sibling = readJson(
+      path.resolve(__dirname, `../../../${name.split('/').at(-1)}/package.json`)
+    );
+    const version = sibling.version;
+    if (typeof version !== 'string') return undefined;
+    return range === 'workspace:^' ? `^${version}` : version;
+  } catch {
+    return undefined;
+  }
+}
```
dannyhw: This walks up to `packages/repack/package.json` from `dist/cli/`, which only exists inside this monorepo. In the published package the branch is never taken: `pnpm pack` already rewrites `workspace:^` to `^5.3.0` in `peerDependencies`, and releases go through `pnpm changeset publish`. So it only affects `init` runs from within the repo, and no test asserts the value it produces (`cli.test.cjs:37` only checks for `*`). Could we drop it and copy the peer range as-is, keeping monorepo layout assumptions out of shipped code?

apps/tester-expo/app.json
whydidoo: Yeah, this was only added for local testing. I can remove it.

packages/repack-expo/src/cli/project.ts
whydidoo: Done, removed the lookup and copy peer ranges as-is. Verified pnpm pack rewrites workspace:^.

packages/repack-expo/package.json
whydidoo: Done, added the devDependencies here. Fixtures now use this package’s own node_modules. Tests pass.

packages/repack-expo/src/index.ts
whydidoo: Done, the plugin and tests now read the version from package.json.

packages/repack-expo/src/cli/project.ts
whydidoo: Done, switched init/doctor and both Expo apps to @callstack/repack/commands.

packages/repack-expo/src/rspack/router/configureExpoRouterEntry.ts
whydidoo: Fixed: absolute Router entries now use the resolved real path. Added iOS/Android regression tests.

whydidoo:
Done, added all three to the Changesets ignore list

PR #1430, opened 2026-07-31
feat: bring lazy platform compilation to Rspack

jbroma:
## Summary

Bring the Rspack development experience to parity with Webpack by compiling each configured platform only when its bundle is first requested.

## Developer wins

- The development server starts without eagerly compiling every configured platform.
- Launching an iOS app no longer waits for an unused Android build, and vice versa.
- Each platform keeps independent compilation state, progress reporting, HMR notifications, stats, and asset caching.
- A platform compiles once on first request, then serves subsequent bundle requests from the existing cache until a source change triggers a rebuild.
- Shutdown now closes Rspack cleanly, rejects pending bundle requests, and still stops the development server when compiler shutdown fails.

## How it works

Rspack starts every child compiler in a multi-compiler eagerly. Re.Pack now holds each child compiler at its initial watch run and releases it on the first bundle request for that platform. The watcher timestamps are refreshed when a gate is released to prevent the deferred compiler from treating changes made during startup as a second, phantom rebuild.

Unsupported platforms fail immediately instead of leaving requests pending. The implementation also tracks compilation, assets, stats, and request resolvers per platform.

## Validation

- Focused lazy-compilation compiler tests: 5/5 passed
- Tester app dev-server integration tests: 2/2 passed
- Re.Pack TypeScript and Biome checks passed
- iOS smoke test passed
- Watcher behavior verified with Rspack 1.6, 2.0, and 2.1
- Changeset validation passed


packages/repack/src/commands/rspack/Compiler.ts
```diff
@@ -93,97 +104,122 @@ export class Compiler {
...
+      if (childCompiler.watching) {
+        childCompiler.watching.startTime = Date.now();
+        childCompiler.watching.lastWatcherStartTime = Date.now();
+      }
+
+      this.isCompilationInProgress[platform] = true;
+      this.buildStartTime[platform] = Date.now();
+
+      if (platform === 'android') {
+        void runAdbReverse({
```
bartekkrok, line 140: I noticed we calling runAdbReverse in watchRun so basically ADB system command on every file save, I know it was before this PR but since we refactoring this parte maybe we could move it outside watchRun or add some debounce (30 seconds) to prevent ADB spam on rapid saves

bartekkrok (approved):
I reviewed this and everything looks fine for me. Good job

packages/repack/src/commands/rspack/Compiler.ts
jbroma: hey @bartekkrok nice catch! It has been that way for a while to ensure smoother dev experience, I'd leave it as it is in this PR but this is definitely something that can be improved, although the gains are almost non-existent in this case - adb happily accepts that spam and this generates no performance overhead in reality.

packages/repack/src/commands/rspack/Compiler.ts
```diff
@@ -198,27 +234,41 @@ export class Compiler {
...
-        this.devServerContext.broadcastToHmrClients<HMRMessage>({
-          action: 'ok',
-          body: { name: platform },
-        });
+      this.devServerContext.notifyBuildEnd(platform);
+      this.devServerContext.broadcastToHmrClients<HMRMessage>({
+        action: 'ok',
+        body: { name: platform },
+      });
+      if (childStats.hash !== previousHash) {
```
MikitasK, line 245: [nit] should we always report build completion here?  reporter only shows the completed state when it receives `time`, otherwise it shows progress ([source](https://github.com/callstack/repack/blob/b3e88de4a51ce6a5bf3093749803429f0dd4114f/packages/repack/src/logging/reporters.ts#L192-L220))

MikitasK (approved):
LGTM 👍👍 
just one non-blocking comment:

packages/repack/src/commands/rspack/Compiler.ts
dannyhw: @MikitasK so the suggestion is that we should pass the progress value? 

@jbroma thoughts on this?

MikitasK: @dannyhw sorry, seems like I wasn’t clear enough

now we're doing `childStats.hash !== previousHash` check like that:
```
const previousHash = this.statsCache[platform]?.hash;
...

if (childStats.hash !== previousHash) {
  const time = buildEndTime - this.buildStartTime[platform];
  this.reporter.process({
    issuer: 'DevServer',
    message: [{ progress: { platform, time } }],
    timestamp: Date.now(),
    type: 'progress',
  });
}
```

my suggestion is to get rid of this `childStats.hash !== previousHash` check:
```
const time = buildEndTime - this.buildStartTime[platform];

this.reporter.process({
  issuer: 'DevServer',
  message: [{ progress: { platform, time } }],
  timestamp: Date.now(),
  type: 'progress',
});
```

I'm proposing always sending existing `{ platform, time }` completion event because reporter uses `value` for in-progress updates & `time` for completed state

MikitasK (approved):

packages/repack/src/commands/rspack/Compiler.ts
jbroma: yeah this is all good points, probably remnants of iteration on this, I'll clean this up 👍

MikitasK (approved):
thanks for addressing above issue 👍

PR #1434, opened 2026-08-05
fix: symbolicate Module Federation remote stack frames

### LINKED ISSUE #1433: Module Federation remote errors aren't symbolicated by the host dev server
### Describe the bug

Errors thrown inside Module Federation remote aren't correctly symbolicated by the host dev server. 

React Native shows the generated remote bundle location instead of the original source file. dev server reports:
```
Failed to symbolicate {
  error: 'Source map for __federation_expose_MiniAppNavigator.chunk.bundle for ios is missing'
}
```

this affects both iOS & Android

### System Info

`@callstack/repack`: `5.3.0` workspace version
`@callstack/repack-dev-server`: `5.3.0` workspace version 
`react-native`: `0.86.0`
`react`: `19.2.8`
`@module-federation/enhanced`: `2.8.0`
`rspack`: `1.6.0`
`webpack`: `5.105.x`
`node.js`: `25.9.0`
`pnpm`: `10.34.5`

### Re.Pack Version

4.3.0

### Reproduction

https://github.com/callstack/repack/tree/main/apps/tester-federation-v2

### Steps to reproduce

https://github.com/user-attachments/assets/ff1bf499-78f2-4d2f-ad00-3a4f1b00bef3

1. add some error inside the mini app. for example in `src/mini/screens/GalleryScreen.tsx`
```
throw new Error('MINI APP ERROR');
```
2. start the host & mini-app servers:
```
pnpm --filter tester-federation-v2 start:hostapp
pnpm --filter tester-federation-v2 start:miniapp
```
3. run the app:
```
pnpm --filter tester-federation-v2 ios
pnpm --filter tester-federation-v2 android
```
4. Navigate to the mini app

### Actual behavior
- stack frame points to `__federation_expose_MiniAppNavigator.chunk.bundle`
- host dev server reports that source map is missing
- original source frame is unavailable or may point to incorrect host file with the same filename

### Expected behavior
- stack frame should resolve to the original mini-app source file
- RN error overlay should display correct source code frame
- pressing the frame should open corresponding mini-app file in IDE

MikitasK:
### Summary

Closes #1433

Fixes development stack-trace symbolication for Module Federation bundles served by a separate Re.Pack dev server.

Previously, the host tried to find remote source maps in its own compiler assets, resulting in:
```
Source map for __federation_expose_MiniAppNavigator.chunk.bundle is missing
```

this change:
- follows remote bundle’s declared sourceMappingURL
- supports both Rspack & Webpack dev servers
- symbolicates frames independently so one invalid frame doesn't discard complete stack
- handles malformed generated `webpack://` source URLs
- uses matching source map’s `sourcesContent` for code frames
- isolates source-map consumers between concurrent requests
- logs first useful symbolicated runtime frame as a terminal fallback
- affects development tooling only & doesn't modify production bundles

### Recording

https://github.com/user-attachments/assets/7d9e4572-ebac-48dc-99e8-be95a37b08af

### Test Plan

1. open `apps/tester-federation-v2`
2. add an error inside mini app:
```
throw new Error('MINI APP ERROR');
```
3. start both dev servers:
```
pnpm --filter tester-federation-v2 start:hostapp
pnpm --filter tester-federation-v2 start:miniapp
```
4. run the application:
```
pnpm --filter tester-federation-v2 ios
pnpm --filter tester-federation-v2 android
```
5. navigate to the mini app & verify that:
- error overlay points to the original mini-app source file
- correct source code frame is displayed
- pressing stack frame opens correct source file
- terminal logs symbolicated source location
- no `source map is missing` error is reported for the remote chunk
6. run automated checks:
```
pnpm --filter @callstack/repack test
pnpm --filter @callstack/repack typecheck
pnpm --filter @callstack/repack build
pnpm --filter @callstack/repack-dev-server test
pnpm --filter @callstack/repack-dev-server typecheck
pnpm --filter @callstack/repack-dev-server build
```


packages/dev-server/src/plugins/symbolicate/__tests__/Symbolicator.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    expect(result.stack[0]).toMatchObject({
+      file: '[projectRoot]/src/App.tsx',
+      lineNumber: 1,
+      column: 0,
+    });
+  });
+});
+
+describe('logSymbolicatedStackFrame', () => {
+  const results: SymbolicatorResults = {
```
bartekkrok: Might be risky to use mutable state across few tests, I would create sth like: 
```
const getMockResults = (): SymbolicatorResults => ({
  stack: [{ file: '[projectRoot]/src/RemoteScreen.tsx', lineNumber: 42, column: 18, methodName: 'RemoteScreen', collapse: false }],
  codeFrame: null,
});
```

packages/dev-server/src/plugins/symbolicate/__tests__/Symbolicator.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  return {
+    getSourceMap,
+    getSource: vi.fn(async () => {
+      throw new Error('Source is not available from the host compiler');
+    }),
+    shouldIncludeFrame: () => true,
+  };
+}
+
+describe('Symbolicator', () => {
```
bartekkrok, line 76: I'd add clearing mock first
`beforeEach(() => {
    vi.clearAllMocks();
  });`

packages/dev-server/src/plugins/symbolicate/__tests__/Symbolicator.test.ts
```diff
@@ -0,0 +1,248 @@
...
+        file: '[projectRoot]/src/RemoteScreen.tsx',
+        lineNumber: 42,
+        column: 18,
+        methodName: 'RemoteScreen',
+        collapse: false,
+      },
+    ],
+    codeFrame: null,
+  };
+
```
bartekkrok, line 260:
```suggestion
beforeEach(() => {
    vi.clearAllMocks();
  });
```

bartekkrok (approved):
Well done job !

packages/dev-server/src/plugins/symbolicate/__tests__/Symbolicator.test.ts
MikitasK: good point 👍  
replaced shared results object with `getMockResults()` so each test now receives fresh fixture

packages/dev-server/src/plugins/symbolicate/__tests__/Symbolicator.test.ts
MikitasK: done ✅

packages/dev-server/src/plugins/symbolicate/__tests__/Symbolicator.test.ts
MikitasK: done ✅

packages/dev-server/src/plugins/symbolicate/Symbolicator.ts
```diff
@@ -45,11 +50,6 @@ export class Symbolicator {
     }
   }
 
-  /**
-   * Cache with initialized `SourceMapConsumer` to improve symbolication performance.
-   */
-  sourceMapConsumerCache: Record<string, SourceMapConsumer> = {};
```
dannyhw, line 51: how does the performance compare? seems like we had this `sourceMapConsumerCache` for performance reasons.

is it not possible now that subsequent requests will have to recalculate? or am I misunderstanding?

MikitasK: basically, we shouldn’t lose any work across calls because old cache was already destroyed & cleared after every `process()` call
just to be safe, I double-checked this & didn’t observe any performance regression. also, I added tests covering same-request reuse, refresh between requests & concurrent requests

dannyhw:
@MikitasK sorry for the delay on reviewing this, I noticed the latest version fails some ci tests, could you take a look?

MikitasK:
> @MikitasK sorry for the delay on reviewing this, I noticed the latest version fails some ci tests, could you take a look?

@dannyhw thanks for catching this up 🙏 
I just updated failed test case, so we can re-run CI check again

dannyhw (approved):

dannyhw:
Thanks @MikitasK !

PR #1435, opened 2026-08-06
fix: copy guarded require properties safely

yellowbeee:
## Summary

- copy guarded `__webpack_require__` runtime properties using their descriptors
- skip non-configurable Function intrinsics already present on the wrapper
- add a strict-mode regression test and a patch changeset

## Problem

The guarded require runtime assigns every own property from the original require function directly to its wrapper. When the runtime is emitted inside a strict-mode bundle, assigning read-only Function properties such as `length` or `name` throws before the React Native application entry can run.

Descriptor copying also avoids evaluating runtime getters while the wrapper is being initialized and preserves their original enumerability and accessors.

## Test plan

- `pnpm --filter @callstack/repack test -- guardedRequire.test.ts --runInBand`
- `pnpm --filter @callstack/repack typecheck`
- `pnpm exec biome check packages/repack/src/plugins/RepackTargetPlugin/implementation/guardedRequire.ts packages/repack/src/plugins/RepackTargetPlugin/implementation/__tests__/guardedRequire.test.ts`

### DIFF
--- a/.changeset/calm-cats-guard.md
+++ b/.changeset/calm-cats-guard.md
@@ -1,0 +1,5 @@
+---
+'@callstack/repack': patch
+---
+
+Prevent guarded require initialization from throwing in strict-mode bundles by safely copying runtime property descriptors.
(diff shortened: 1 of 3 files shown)


dannyhw:
Hey thanks for your pr 🙏.

Was wondering if you can help describe the issue it solves a bit mode.

Do we have an issue for the bug it fixes? How to reproduce that problem?

dannyhw:
@yellowbeee please respond to the questions above

dannyhw:
Closing since there's no response

PR #1440, opened 2026-08-12
fix: let error boundaries handle chunk loading failures

### LINKED ISSUE #1428: Production chunk loading errors bypass React Error Boundaries and crash the app
### Describe the bug

Production chunk loading failures are converted into fatal React Native exceptions before React.lazy can deliver the rejected import to the nearest Error Boundary.

This affects regular dynamic chunks and Module Federation v2 remotes when an exposed chunk is unavailable.

Actual behavior:
- The application process terminates.
- Android reports FATAL EXCEPTION: mqt_v_native with a ChunkLoadError.
- The Error Boundary never gets a chance to render its fallback.

Expected behavior:
- The dynamic import rejects.
- The nearest Error Boundary renders its fallback.
- The application process remains alive.

Root cause:
Module Federation handles the remote loading rejection by registering a module factory that throws the original ChunkLoadError. The following module execution goes through Re.Pack's guarded __webpack_require__ wrapper, which calls ErrorUtils.reportFatalError. This converts a recoverable asynchronous loading failure into an unconditional fatal exception before it can propagate back to the React.lazy promise.

Proposed solution:
Remove the global guarded __webpack_require__ wrapper and allow module execution errors to propagate to their caller. Uncaught startup errors will still reach React Native's top-level error handling naturally, while dynamic import failures can be handled by promises and React Error Boundaries.

### System Info

```shell
React Native: 0.86.0
React: 19.2.x
Rspack: 1.6.0
Platforms tested:
- Android Pixel 3a API 33 arm64-v8a emulator
- iOS 17.5 iPhone 15 Pro Simulator
Build mode: production/release
Module Federation: v2
```

### Re.Pack Version

5.2.5 (current main)

### Reproduction

https://github.com/callstack/repack/tree/main/apps/tester-federation-v2

### Steps to reproduce

1. Build the tester-federation-v2 host and mini app in production mode.
2. Serve build/mini-app on port 8082.
3. Load MiniApp/MiniAppNavigator through React.lazy and wrap it in an Error Boundary.
4. Remove __federation_expose_MiniAppNavigator.chunk.bundle from the served remote output.
5. Launch the Release application and open Gallery.

Before the fix:
- Android terminates with FATAL EXCEPTION: mqt_v_native.
- iOS also reports the loading failure as fatal.
- The Error Boundary fallback is not rendered.

Validation after removing the fatal require wrapper:
- Android with chunk available: remote renders.
- Android with chunk missing: Error Boundary renders and the PID remains unchanged.
- iOS with chunk available: remote renders.
- iOS with chunk missing: Error Boundary renders and the process remains alive.

MikitasK:
### Summary

Closes #1428 

MF v.2 can surface failed remote chunk load through synthetic module factory. Resulting `ChunkLoadError` passes through Re.Pack's guarded `__webpack_require__`, where it's reported as fatal. This bypasses dynamic-import rejection path preventing `React.lazy` Error Boundaries from handling failure & causing Release app to terminate

https://github.com/user-attachments/assets/2d234a3b-f544-444a-85db-168bdfccf166

PR change keeps Metro-compatible guarded require but allows `ChunkLoadError` to propagate to dynamic-import promise. promise can then reject normally & be handled by nearest Error Boundary

- startup module errors are still reported through `ErrorUtils.reportFatalError`
- optional nested `require()` failures remain catchable
- ordinary module errors retain existing fatal behavior
- Webpack & Rspack module-cache behavior is unchanged

https://github.com/user-attachments/assets/323a63a0-bfd9-4e19-bda4-d6b211df089a

### Test plan

In repo root:
1. `pnpm --filter @callstack/repack typecheck`
2. `pnpm --filter @callstack/repack run build:js`
3. `pnpm --filter @callstack/repack run build:ts`
4. `pnpm --filter integration-test test -- RepackTargetPlugin.test.ts`

In `apps/tester-federation-v2`:

1. build host prod bundle:
```
pnpm exec react-native bundle \
  --config config.host-app.mts \
  --platform android \
  --entry-file src/host/index.js \
  --dev=false \
  --bundle-output build/host-app/android/output-local/index.android.bundle \
  --assets-dest build/host-app/android/output-local/res
  ```
2. build mini-app prod bundle:
```
pnpm exec react-native bundle \
  --config config.mini-app.mts \
  --platform android \
  --entry-file src/mini/index.js \
  --dev=false \
  --bundle-output build/mini-app/android/output-local/index.android.bundle \
  --assets-dest build/mini-app/android/output-local/res
  ```
3. serve `build/mini-app` on port 8082:
```
python3 -m http.server 8082 --directory build/mini-app
```
4. verify that the manifest & exposed chunk are initially available (both requests should return HTTP 200):
```
curl -I http://localhost:8082/android/mf-manifest.json
curl -I http://localhost:8082/android/__federation_expose_MiniAppNavigator.chunk.bundle
```
5. rename exposed chunk to simulate deployment with missing chunk:
```
mv \
  build/mini-app/android/__federation_expose_MiniAppNavigator.chunk.bundle \
  build/mini-app/android/__federation_expose_MiniAppNavigator.chunk.bundle.disabled
```
6. verify that chunk now returns HTTP 404:
```
curl -I http://localhost:8082/android/__federation_expose_MiniAppNavigator.chunk.bundle
```
7. `adb reverse tcp:8082 tcp:8082`
8. build & launch Release app:
```
pnpm exec react-native run-android \
  --mode release \
  --appId com.tester.federationV2 \
  --no-packager
```
9. open `Gallery` in app

Expected result:
- missing chunk request returns HTTP 404
- logcat contains `ChunkLoadError`
- `Gallery` renders "Failed to load Mini App" fallback
- logcat doesn't contain any fatal exceptions


packages/repack/src/plugins/RepackTargetPlugin/implementation/guardedRequire.ts
```diff
@@ -29,8 +47,20 @@ module.exports = function () {
 
   // Copy all properties from the original function to the wrapped function
   Object.getOwnPropertyNames(originalWebpackRequire).forEach((key) => {
-    // @ts-ignore
-    guardedWebpackRequire[key] = originalWebpackRequire[key];
+    var descriptor = Object.getOwnPropertyDescriptor(
```
dannyhw: how do these changes to copy the properties impact the fix? or does it solve some other issue?

MikitasK: good catch, thanks 👍  basically, these changes don't directly impact 'chunk load error' fix. 
so I just reverted it to keep this PR focused on chunk loading error & not brake any existing logic

packages/repack/src/plugins/RepackTargetPlugin/implementation/guardedRequire.ts
```diff
@@ -6,6 +6,14 @@ module.exports = function () {
   var inGuard = false;
   var originalWebpackRequire = __webpack_require__;
 
+  function isChunkLoadError(error: unknown) {
```
whydidoo, line 9: @MikitasK hey, ChunkLoadError is not the only error that can cause this issue. For example, we encountered Module Federation errors containing a #RUNTIME-\d+ code that also crashed the app, so we added a check for them in our code as well. Could we check for and handle these errors here too?

MikitasK: hey @whydidoo, thanks for additional context 🙏 
may I just ask to share the exact `#RUNTIME-*` error message/stack & the check you added?

I’d prefer not to match every `/#RUNTIME-\d+/` error because MF also uses these codes for non-loading problems, such as invalid `loadShareSync` calls, missing runtime initialization, configuration errors ([source](https://github.com/module-federation/core/blob/08fdc4dfbf4d5498638f7fa7bc2c772fcae4d066/packages/error-codes/src/desc.ts#L22-L40)). MF appends `while loading "<expose>" from <module>` for remote-loading failures before installing throwing module factory ([source](https://github.com/module-federation/core/blob/08fdc4dfbf4d5498638f7fa7bc2c772fcae4d066/packages/webpack-bundler-runtime/src/remotes.ts#L38-L49))

if your error contains that marker, we can safely handle runtime errors from that specific path without changing guarded-require behavior for unrelated MF errors

whydidoo: Thanks, that makes sense. Unfortunately, this happened quite a while ago, so I don’t have the exact stack trace available right now.

We did observe app crashes related to `#RUNTIME-*` errors and added broader handling on our side. I’ll need to check the old implementation and logs.

I agree that matching every runtime error would be too broad.

MikitasK: got you @whydidoo, thanks for confirming 🙏  

then may I ask you to share this if logs / implementation if you find it 🙏 that way we would be able reproduce your specific `#RUNTIME-*` failure & handle it in dedicated follow-up PR
for now, what about keeping this PR scoped to the reproduced `ChunkLoadError` case?

whydidoo: Yep, sure

dannyhw (approved):

PR #1443, opened 2026-08-24
chore: support rn 0.87

Francesco-Voto:
React Native 0.87 removed `rn-get-polyfills.js` (and the `@react-native/js-polyfills` dependency), moved the asset registry to `src/asset-registry.js`, and dropped the `./*` / `./src/*` wildcards from its `exports` map.

### Summary

- **Polyfills:** read from `rn-get-polyfills.js` when present (RN <= 0.86); otherwise resolve `@react-native/js-polyfills` from the project, falling back through `@react-native/metro-config` (which still depends on it and is a template devDependency). Throws an actionable error when neither can be found.
- **Asset registry:** on the 0.87 layout, prepend an exact-match resolve alias from `react-native/Libraries/Image/AssetRegistry` to `src/asset-registry`. The request emitted by the assets loader and `IncludeModules` is unchanged, so the Module Federation `react-native/` share key is stable. No-op on <= 0.86.
- **Deep imports with package exports:** on the 0.87 layout, alias `react-native/src/private` to disk so first-party packages such as `@react-native/virtualized-lists` (which deep-import `react-native/src/private/featureflags/ReactNativeFeatureFlags`, see facebook/react-native#57933) keep resolving when `enablePackageExports` is on. Metro tolerates this by falling back to file resolution; strict resolvers need the alias. No-op when the exports map still exposes the wildcards.

Aliases are prepended so a user's generic `react-native` alias cannot shadow them.

### Test plan

- Unit tests for `resolveReactNativePolyfills`, `getReactNativeAssetRegistryAlias` and `getReactNativeDeepImportAliases` (tmp-dir fixtures, injectable resolver).
- Integration tests against a 0.87-shaped `react-native` fixture under both Rspack and webpack, with and without package exports.
- Verified locally against the real `react-native@0.87.1` package with both bundlers and both `enablePackageExports` settings: asset registry, feature-flag deep import and polyfills all resolve with no resolution errors.
- Device-level validation on RN 0.87 will be done separately.


dannyhw:
hey sorry for the delay, does it cause some kind of failure without this that I can use to check before/after?

Francesco-Voto:
@dannyhw yes in rn 0.87 looks like the polyfill was moved so the code si breaking at build time. You can try to integrate ina project with rn 0.87

Francesco-Voto:
@dannyhw I added more check based on what I noticed in a RN 0.97 project. I had to insert the alias 
`'react-native/src/private/featureflags/ReactNativeFeatureFlags': resolve(__dirname, 'node_modules', 'react-native', 'src', 'private', 'featureflags', 'ReactNativeFeatureFlags.js')`

but this seems to be an [issue with react native](https://github.com/react/react-native/issues/57933)

dannyhw:
thanks, I hope to get to this soon, sorry again for the delay.

dannyhw:
@Francesco-Voto can you check the test failure?

dannyhw (approved):

PR #1444, opened 2026-08-24
fix: minify .bundle assets with terser-webpack-plugin 5.6.0 and newer

### LINKED ISSUE #1390: Default minimizer silently no-ops with terser-webpack-plugin >= 5.6.0 under Rspack — production bundles ship unminified
### Environment

- `@callstack/repack`: 5.2.5
- `@rspack/core`: 1.7.0
- `terser-webpack-plugin`: 5.6.1 (resolved from Re.Pack's own dependency range `^5.3.14`)
- Node 24, React Native 0.83.9, monorepo with pnpm (isolated `node_modules`)

### Summary

Re.Pack's default minimizer (`getMinimizerConfig` → `getTerserPlugin`) prefers a project-root `terser-webpack-plugin` and falls back to the copy declared in Re.Pack's own dependencies (`^5.3.14`):

```js
// dist/commands/common/config/getMinimizerConfig.js
try {
  terserPluginPath = require.resolve('terser-webpack-plugin', { paths: [rootDir] });
} catch {
  terserPluginPath = require.resolve('terser-webpack-plugin');
}
```

That range now resolves to **5.6.1**, and `terser-webpack-plugin >= 5.6.0` depends on webpack-only internals (e.g. `compiler.webpack.javascript.JavascriptModulesPlugin.getCompilationHooks`, `ModuleFilenameHelpers.matchObject` via `compiler.webpack`). Under Rspack it **does nothing, without any error or warning** — production bundles are emitted completely unminified (comments and whitespace included), and the asset is not marked `[minimized]` in stats.

The failure is easy to miss because of the resolution order:

- **yarn / npm hoisted layouts**: some other dependency often hoists an older `terser-webpack-plugin` (5.3.x) to the workspace root, the root-first resolve picks it up, and everything works — by accident.
- **pnpm / isolated layouts**: nothing is hoisted to the root, the fallback resolves Re.Pack's own 5.6.1, and every production bundle silently ships unminified.

We hit this while migrating a Module Federation monorepo from yarn (hoisted) to pnpm: identical source, identical configs, and the remote `expose` chunks grew ~37–60% (e.g. 1.68 MB → 2.68 MB Hermes bytecode for one remote) with zero build errors. Bisecting the bundler pipeline eventually surfaced the minimizer no-op.

### Reproduction (minimal A/B)

Same trivial entry, Rspack 1.7.0, `mode: 'production'`, only the `terser-webpack-plugin` version differs in `optimization.minimizer` (using exactly the options Re.Pack passes):

```js
new TerserPlugin({
  test: /\.(js)?bundle(\?.*)?$/i,
  extractComments: false,
  terserOptions: { format: { comments: false } },
})
```

| terser-webpack-plugin | output |
|---|---|
| 5.3.16 | 298 bytes, `[minimized]`, mangled/compressed |
| 5.3.17 | 298 bytes, `[minimized]` |
| 5.4.0 | 298 bytes, `[minimized]` |
| 5.5.0 | 298 bytes, `[minimized]` |
| **5.6.0** | **1.24 KiB, untouched (comments/whitespace intact), no `[minimized]`, no error** |
| **5.6.1** | **1.24 KiB, untouched, no error** |

So the break is exactly at 5.6.0.

### Expected behavior

Either the bundles get minified, or the build fails loudly. A silent no-op in the production minimizer is the worst case — it shipped to our staging users unnoticed.

### Suggested fixes

Any of these would prevent the silent regression:

1. Constrain Re.Pack's dependency to `>=5.3.14 <5.6.0` (5.6.0 effectively dropped Rspack compatibility).
2. When the bundler is Rspack, skip `terser-webpack-plugin` resolution and use a known-good path (e.g. `SwcJsMinimizerRspackPlugin`, with whatever guard the 1.4.11-era issue requires).
3. After the compilation, assert that production assets matching the minimizer `test` carry `info.minimized`, and fail or warn otherwise.

The root-first `require.resolve` is also worth reconsidering: it makes behavior depend on what other packages happen to hoist into the workspace root, which is how this stayed hidden on yarn and only surfaced on pnpm.

Happy to provide the full repro repo if useful.

giaBaoJS:
## Summary

Configure Re.Pack's own `minify` wrapper on `TerserPlugin` instead of relying on the plugin's built-in `terserMinify`.

## Why

`terser-webpack-plugin` 5.6.0 added per-minimizer asset filters, and its terser implementation declares `terserMinify.filter = (name) => /\.[cm]?js(\?.*)?$/i.test(name)` (`dist/utils.js:312`). Re.Pack emits `index.bundle` and `[name].chunk.bundle`, so every asset is rejected by that filter and dropped from `assetsForMinify` before minification runs. Nothing is reported: no error, no warning, and the asset is not flagged `[minimized]` in stats. Production bundles simply ship unminified.

This hits webpack and Rspack alike, and pinning `terser-webpack-plugin` in this repo does not cover it: `getTerserPlugin` prefers the copy resolved from the project root, so any app that resolves 5.6.0 or newer still gets the silent no-op.

## Implementation

The plugin only consults `filter` on the implementation it is handed. A plain function has no `filter` property, so wrapping `terserMinify` makes every version accept `.bundle` assets:

```ts
function repackTerserMinify(input, sourceMap, minimizerOptions, extractComments) {
  return require('terser-webpack-plugin').terserMinify(
    input, sourceMap, minimizerOptions, extractComments
  );
}

repackTerserMinify.getMinimizerVersion = () =>
  require('terser-webpack-plugin').terserMinify.getMinimizerVersion?.();
```

`terserMinify` is exported by 5.5.0 as well, so the wrapper behaves identically on old and new versions. The plugin serializes it into a jest worker (`getWorker().transform(serialize(options))`) and re-evaluates it there with its own `require`, which is why the wrapper closes over nothing and resolves `terser-webpack-plugin` by name.

`getMinimizerVersion` is forwarded because the plugin reads it on the main thread to build the chunk hash (`index.js:686` on 5.5.0, `index.js:724` on 5.6.1). Without the forward that slot silently degrades to the `"0.0.0"` literal and a bare `terser` upgrade stops invalidating persistent cache. It is never read inside the worker, and serialization drops function statics anyway, so attaching it cannot affect the worker path.

Closes #1390.

## Validation

- `pnpm --filter @callstack/repack test`: 35 suites, 315 tests passed
- `pnpm --filter @callstack/repack typecheck` and `biome check`: clean
- reverting only `getMinimizerConfig.ts` while keeping the new tests turns `should minify .bundle assets with a plugin that only accepts .js` red on `expect(implementation.filter).toBeUndefined()` with `Received: [Function acceptsJsOnly]`, so it fails because the `.bundle` asset is filtered out, not on an import or compile error
- end to end in `apps/tester-app` with `terser-webpack-plugin@5.6.1` resolved at the app root, `react-native bundle --platform ios --entry-file index.js --dev=false`:

| bundler | without the wrapper | with the wrapper |
| --- | --- | --- |
| webpack 5.105.4 | 4,846,598 bytes, full identifier names | 1,969,176 bytes |
| Rspack 1.6.0 | 4,822,812 bytes, full identifier names | 1,969,836 bytes |


packages/repack/src/commands/common/config/getMinimizerConfig.ts
```diff
@@ -1,19 +1,93 @@
...
+}
+
+function resolveTerserPluginCandidate(
+  paths?: string[]
+): TerserPluginCandidate | undefined {
+  const options = paths ? { paths } : undefined;
+  try {
+    const pluginPath = require.resolve('terser-webpack-plugin', options);
+    return { pluginPath, version: resolveTerserPluginVersion(options) };
+  } catch {
```
MikitasK: wdyt about checking plugin’s actual capability instead of its manifest version?

`terser-webpack-plugin/package.json` may be hidden by an `exports` map ([node docs](https://nodejs.org/api/packages.html#package-entry-points)), causing `version` to be `undefined` &  affected plugin to be accepted.
since 5.6.x exposes filter that rejects `.bundle` files ([source](https://github.com/webpack/minimizer-webpack-plugin/blob/v5.6.1/src/utils.js#L344-L347)), could we load the candidate & check:

```
const filter = Plugin.terserMinify?.filter;
return typeof filter !== 'function' || filter('index.bundle') !== false;
```

this directly covers current & future versions, `version` can remain only for warning text

giaBaoJS: Agreed, switched to the capability check.

What I verified against the published package rather than the GitHub source:

- `terserMinify` is a static on the default export (`index.js` does `TerserPlugin.terserMinify = terserMinify`), so it is reachable through the same `importDefaultESM` we already use to load the plugin. On 5.6.1 `terserMinify.filter('index.bundle')` returns `false`; on 5.5.0 `terserMinify.filter` is `undefined`, so both directions behave the way you described.
- 5.6.1 currently ships no `exports` map, so `terser-webpack-plugin/package.json` still resolves today. The capability check covers the hidden manifest case regardless, and there is a test fixture with an `exports` map for it.

Two adjustments to your snippet:

1. `filter` is not declared in the plugin's typings before 5.6.0 (`types/utils.d.ts` gains `function filter(name: string): boolean` only in 5.6.x) and we typecheck against 5.5.0, so `Plugin.terserMinify?.filter` does not compile. I probe it through a small structural cast instead.
2. The plugin calls it as `impl.filter(name, info)`. The built-in implementation only reads `name`, but a custom or future one may rely on `info`, so the one argument call is wrapped and a throw is treated as "accepts" rather than letting minimizer selection blow up.

Candidate resolution is async now because it loads the module instead of only resolving the path, and the bundled copy is loaded lazily, only when the project one is rejected. `version` is used for the warning text only and is dropped from the message when the manifest cannot be read.

dannyhw:
I understand the desire to fix this but honestly my personal preference will be to fix the rspack minifier rather than continuing to default to terser

packages/repack/src/commands/common/config/getMinimizerConfig.ts
MikitasK: thanks for the thorough follow-up 🙏  my concern is addressed, resolving this thread 👌

MikitasK:
> I understand the desire to fix this but honestly my personal preference will be to fix the rspack minifier rather than continuing to default to terser

makes sense 👍  since webpack is affected too, wdyt about keeping this fallback for webpack, but using/fixing native minimizer for rspack?

dannyhw:
but i believe terser still works for webpack, don't we already prefer the users version for that?

giaBaoJS:
I measured it before answering, and terser does not still work for webpack. Same root cause as Rspack.

terser-webpack-plugin 5.6.0 added `terserMinify.filter = (name) => /\.[cm]?js(\?.*)?$/i.test(name)` (dist/utils.js:16 and :310). The webpack asset Re.Pack emits is literally `index.bundle` (plus `*.chunk.bundle`), so the filter returns false and the plugin drops the asset from `assetsForMinify` with no warning and no error.

Real builds in `apps/tester-app`, webpack 5.105.4, `react-native webpack-bundle --platform ios --dev=false`, config with no `optimization.minimizer` so Re.Pack's default runs (which is also the shape of `templates/webpack.config.mjs`):

| terser-webpack-plugin in the project | main.jsbundle |
| --- | --- |
| 5.5.0 | 1,969,176 bytes, mangled |
| 5.6.1 | 4,846,598 bytes, `/******/ var self = self || ...`, full identifier names |

And preferring the user's version is what triggers this rather than what protects against it. webpack itself depends on `terser-webpack-plugin: ^5.3.17`, so a plain `npm i webpack@5.105.4` puts 5.6.1 at the project root today, and `resolveTerserPluginCandidate([rootDir])` looks there first. Under pnpm nothing is hoisted, Re.Pack's own pinned 5.5.0 wins, and it works. An rspack-only npm project has no terser at the root at all, so webpack is the more exposed of the two.

Two things you may want before deciding on this PR:

1. I forced `SwcJsMinimizerRspackPlugin` on Rspack 1.6.0 with Re.Pack's `test` regex and it minified fine: 1,972,343 bytes in 2.51 s vs terser's 1,969,836 in 6.37 s, async chunk minified too. The comment in `getMinimizerConfig.ts` still points at the 1.5.0 regression from #1273. That may already be unblocked, which would be your preferred direction. I did not run the app, so this is "produces minified output", not "verified at runtime".

2. There is a smaller fix than what this PR currently does. A plain function has no `.filter` property, so any version accepts the asset:

```js
new TerserPlugin({
  test: /\.(js)?bundle(\?.*)?$/i,
  extractComments: false,
  minify: function repackTerserMinify(input, sourceMap, minimizerOptions, extractComments) {
    return require('terser-webpack-plugin').terserMinify(
      input, sourceMap, minimizerOptions, extractComments
    );
  },
  terserOptions: { format: { comments: false } },
})
```

I ran that with 5.6.1 in both the project and Re.Pack's own node_modules, so the default terser was a confirmed no-op and only the wrapper could do work: 1,969,176 bytes, byte identical to the 5.5.0 output. `terserMinify` is exported by 5.5.0 as well, and the wrapper is self contained so it survives worker serialization. It drops the capability probe, the fallback, and the warning.

Happy to rewrite #1444 as that wrapper, or to close it if you would rather fix the Rspack minimizer first and handle webpack separately.

dannyhw:
I see, seems like i may have misunderstood the issue then. What do you propose as the right solution? I.e do you think what you mentioned in your comment (wrapper) is better or the solution presented in the pr already?

giaBaoJS:
The wrapper, and I would rather rewrite this PR as that than merge what is here now.

What is here reacts to the symptom. It probes whether the installed plugin will refuse `.bundle`, then falls back to Re.Pack's own pinned copy and warns. It works, but it needs a capability probe, a fallback path and a warning string to do it, and the user ends up minifying with a different terser than the one in their lockfile.

The wrapper removes the condition instead. In 5.6.1 the dispatch is `index.js:357`:

```js
if (typeof impl.filter !== "function" || ... impl.filter(name, info) !== false) {
```

so the filter is only consulted when the configured `minify` has one. Every built-in gets `.filter` assigned in `utils.js`, but a plain function does not, so passing our own named function makes every version from 5.5.0 up minify the asset, and the version that runs is the user's own. That is what "we prefer the user's version" was supposed to give them in the first place.

The judgement call worth saying out loud: we would be deliberately stepping around a filter the plugin author added. I think it is defensible here, since the doc comment on that option describes it as "return true when the minimizer supports the asset" and terser genuinely does support this asset. `.bundle` is JavaScript, and `JS_FILE_RE` is an extension heuristic rather than a capability test. But it is a call, not a fact, so it is yours to make rather than mine.

Say the word and I will rewrite it. If you would rather land the Rspack minimizer switch first and treat webpack on its own, I am equally happy to close this and open the wrapper as a separate small PR whenever it suits you.

dannyhw:
@giaBaoJS ok that makes sense, sounds like the wrapper would be better. If you have the capacity that would be really great if you add those changes

giaBaoJS:
Rewritten as the wrapper, pushed as a separate commit (878f75c) so the delta from the probe version is visible.

Against `main` the source change is now +20/-2 in `getMinimizerConfig.ts`: `getTerserPlugin` also returns the path it resolved, and `getTerserConfig` passes a `minify`. The probe, the fallback and the warning are gone.

`apps/tester-app`, `--platform ios --dev=false`, cache cleared between runs, `terser-webpack-plugin` swapped at the app root. Re-measured from scratch, not carried over from my earlier comment:

| bundler | plugin | before | after |
| --- | --- | --- | --- |
| webpack | 5.6.1 | 4,846,598 | 1,969,176 |
| webpack | 5.5.0 | 1,969,176 | 1,969,176 |
| rspack | 5.6.1 | 4,822,812 | 1,969,836 |
| rspack | 5.5.0 | 1,969,836 | 1,969,836 |

All four "after" bundles are byte-identical to their 5.5.0 baseline. `__webpack_require__` occurrences go 6157 -> 50 on webpack and 6144 -> 46 on rspack and the `/******/ var self = self || ...` runtime banner is gone, so it is mangling and not just whitespace. The async chunk goes 728 -> 455 bytes, so `*.chunk.bundle` is covered by the `test` regex as well. 5.5.0 has no filter machinery at all (no `impl.filter` in `dist/index.js`, no `JS_FILE_RE` in `dist/utils.js`) and does export `terserMinify`, so the wrapper is a no-op there, which is what the identical 5.5.0 rows show.

Two things came out differently from the sketch in my earlier comment.

The wrapper cannot say `require('terser-webpack-plugin')` by name. It is serialized with `serialize-javascript` and re-evaluated in a worker (`index.js:528` into `minify.js` `transform`), where `require` resolves from the plugin's own `dist/`, which is the correct copy. But with `parallel` off, or when `getAvailableNumberOfCores` returns 0, the same function runs in-process and `require` resolves from Re.Pack's `dist/` instead. Under pnpm with a direct `terser-webpack-plugin` dependency those are two different copies, which is the same "not the terser in your lockfile" problem I was arguing against. So the resolved path is baked into the body:

```js
new Function(
  `return function repackTerserMinify(input, sourceMap, minimizerOptions, extractComments) {
  return require(${JSON.stringify(pluginPath)}).terserMinify(input, sourceMap, minimizerOptions, extractComments);
}`
)()
```

`new Function` is there for the serialization constraint rather than for style: anything the wrapper closes over is gone by the time it runs, and it fails at runtime, not at build time. Checked both directions on a standalone webpack 5.105.4 + plugin 5.6.1 build. The wrapper logs `isMainThread=false`, so it really is going through a worker, and the closure-capturing variant fails with:

```
index.bundle from Terser plugin
TerserPlugin is not defined
ReferenceError: TerserPlugin is not defined
    at repackTerserMinify (eval at transform (node_modules/terser-webpack-plugin/dist/minify.js:328:3), <anonymous>:5:5)
```

`getMinimizerVersion` has to be carried across. The plugin folds it into the chunk hash (`index.js:724`) and Re.Pack builds with `cache: { type: 'filesystem' }` (`getCommandConfig.ts:19`), so a bare wrapper pins it to `"0.0.0"` and bumping terser stops invalidating the cache. One line copies it onto the wrapper; it does not bring `filter` along.

Tests rewritten. The old ones asserted which copy of the plugin got selected, which is machinery that no longer exists. The new ones put a fake project plugin through the same two steps the real one does, `filter` and then a `toString()` round trip through `new Function`, and assert the `.bundle` asset comes back minified. Reverting only `getMinimizerConfig.ts` to `main` and keeping the tests gives:

```
● getMinimizerConfig › should minify .bundle assets with a project plugin that only accepts .js
    - project(const answer = 40 + 2;
    + const answer = 40 + 2;
```

so it goes red because the asset comes back untouched, not because of a missing import.

The warning has nothing left to report, since every version now minifies. Changeset description updated to match.

giaBaoJS:
@dannyhw wrapper is in. It replaces the version check, the fallback to the bundled copy and the warning, so nothing version specific is left in `getMinimizerConfig`.

Measured on `apps/tester-app` with `terser-webpack-plugin@5.6.1` resolved at the app root: the iOS production bundle goes from 4,846,598 bytes unminified to 1,969,176 on webpack, and from 4,822,812 to 1,969,836 on Rspack.

dannyhw:
thanks @giaBaoJS this looks good, I will try it soon and get back to you. Sorry about the delay.

dannyhw (approved):

PR #1446, opened 2026-08-24
fix: resolve absolute stack frame paths in the webpack compiler

giaBaoJS:
## Summary

- guard the webpack compiler's `getSource` with `path.isAbsolute`, matching the Rspack compiler
- add `getSource` unit tests covering both the absolute and the relative case

## Why

`start.ts` runs every symbolicator request through `resolveProjectPath` before handing it to the compiler:

https://github.com/callstack/repack/blob/b7a86113225f028bd8f415ca0fb092692143bf60/packages/repack/src/commands/start.ts#L210-L214

`resolveProjectPath` ends in `path.resolve`, so the compiler receives an absolute path. `rspack/Compiler.ts:353` handles that:

```js
const filePath = path.isAbsolute(filename)
  ? filename
  : path.join(this.rootDir, filename);
```

`webpack/Compiler.ts:279` did not:

```js
const filePath = path.join(this.rootDir, filename);
```

so the root was prepended a second time. With `rootDir` `/project`, a frame at `/outside/project/file.js` was read from `/project/outside/project/file.js`, which does not exist. `getSource` threw, and the redbox rendered the frame with no code frame.

This is not limited to files outside the project root. `SourceMapPlugin` emits `[projectRoot]/...` for every in-project module, `resolveProjectPath` turns that into an absolute path, and `path.join('/project', '/project/src/App.tsx')` is `/project/project/src/App.tsx`. So on webpack every symbolicated frame lost its source, while Rspack was fine.

It looks like an oversight rather than a deliberate difference. `git log -S "path.isAbsolute(filename)"` returns exactly one commit:

```
b15b7d77 fix: sourcemaps handling in dev & prod (#1172)
```

That commit is the one that introduced `resolveProjectPath` and wired it into **both** start commands, including `packages/repack/src/commands/webpack/start.ts`. It added the `path.isAbsolute` guard only to `rspack/Compiler.ts`. The webpack producer was updated, the webpack consumer was not.

This is the `getSource` bullet from the v6 list in #1421. It is an independent v5 bug that stands on its own, so I am not marking #1421 as closed. Happy to have it folded into the unification work instead if you would rather handle it there.

## Validation

- `pnpm --filter @callstack/repack test`: 33 suites, 301 tests passed (299 before)
- `pnpm test`: 10 tasks successful
- `pnpm typecheck`: 11 tasks successful. `biome check`: 481 files, clean
- counterfactual: reverting only `Compiler.ts` while keeping the tests turns `reads an absolute filename as-is` red on the assertion, `Rejected to value: [CLIError: File /outside/project/file.js not found]`, not on an import or compile error. Instrumenting the spy shows `readFile` receiving `/project/outside/project/file.js`
- non-vacuity in the other direction: replacing the change with `const filePath = filename` (dropping the join entirely) turns `resolves a relative filename against the project root` red, so neither test passes both ways


dannyhw:
looks solid, thanks @giaBaoJS

dannyhw (approved):

PR #1450, opened 2026-08-28
test: make path assertions platform-agnostic so the suite passes on Windows

KallinikosMil:
## Summary

Five tests across two suites fail on Windows because they compare platform-native paths against hard-coded POSIX literals, so `pnpm test` is red even though the code under test is correct:

```
● HermesBytecodePlugin › getHermesCLIPath › returns new hermes-compiler path when it exists

  Expected: "path/to/hermes-compiler/hermesc/osx-bin/hermesc"
  Received: "path\to\hermes-compiler\hermesc\osx-bin\hermesc"
```

`getHermesCLIPath` and `resolveProjectPath` build those paths with `path.join` / `path.resolve`, which is exactly what they should do — the separators are right for the platform. It's the expectations that are POSIX-only.

The fix runs the expectations through `path.join` / `path.resolve` rather than comparing raw strings. Both are no-ops on POSIX, so Linux and macOS runs are byte-for-byte unchanged; the literals stay readable as POSIX in the source.

### One case needed a bit more

In `resolveProjectPath.test.ts` the **root** is resolved as well, not just the expectation. With a POSIX root and no drive letter, up-level navigation that reaches the root collapses to a bare `\`, and Windows then reads `\` + `/shared/...` as the start of a **UNC share**:

```
● should resolve [projectRoot^N] prefix with up-level navigation

  Expected: "C:\shared\utils.js"
  Received: "\\shared\utils.js\\"     ← \shared\utils.js\ , a network path
```

Resolving the root first gives it a drive letter, so `path.join` keeps the drive when it walks up and the result stays local. To be clear about scope: I don't think this is reachable in production, because `rootDir` there comes from a real project directory and always carries a drive on Windows. It's a property of feeding the function a POSIX absolute path on Windows, so I've fixed it in the test and left `resolveProjectPath` alone. Flagging it because the failure mode is a surprising one and worth knowing about.

## Test plan

Windows 11, Node 22.23.2, pnpm 10.34.5.

**Before** — `@callstack/repack`:

```
Test Suites: 2 failed, 31 passed, 33 total
Tests:       5 failed, 294 passed, 299 total
```

**After**:

```
Test Suites: 33 passed, 33 total
Tests:       299 passed, 299 total
```

`pnpm lint` (biome) is clean. `metro-compat-test` (47 passed), `resolver-cases-test` (3 files) and `repack-dev-server` (5 tests) all pass on Windows both before and after.

**Not addressed here:** `tests/integration` fails on Windows with `Module build failed` — 4 files, 50 tests, 24 snapshots. I verified this is pre-existing and unrelated by stashing my changes and re-running on a clean tree, which produced the identical `50 failed | 8 passed`. It looks like a separate Windows issue and I didn't want to bundle a fix for it into this PR. Happy to look into it separately if that'd be useful.

I noticed `test-main-matrix.yml` already has an `os` matrix axis, currently `[ubuntu-latest]`. If you'd like these kept honest, adding `windows-latest` there would do it — glad to open that as its own PR rather than assume you want the extra CI minutes.


dannyhw:
Thanks for this, nice catch and a really clear write-up. The UNC share explanation in particular is worth having on record.

I pushed one small tweak to your branch (f5a914b): the three single-argument `path.join(...)` expectations in `HermesBytecodePlugin.test.ts` now use `path.normalize(...)` instead. With one argument, `path.join` just normalizes the string anyway, so the behaviour is identical on both platforms; `normalize` simply says what's actually happening.

- https://nodejs.org/api/path.html#pathnormalizepath
- https://nodejs.org/api/path.html#pathjoinpaths

And yes, we'd welcome the follow-up PR adding `windows-latest` to the `os` matrix in `test-main-matrix.yml`. Without it there's nothing keeping these tests honest on Windows.

dannyhw (approved):

PR #1454, opened 2026-09-06
fix: read react-native shared config from array form of shared

giaBaoJS:
## Summary

`ModuleFederationPlugin` adds `react-native/` and `@react-native/` to `shared` so deep imports resolve through the shared scope, and it copies `eager`, `import` (and `version` in V1) from the user's `react-native` entry. That copy only works when `shared` is an object. With the array form, the deep imports always end up as `eager: true` and without `import: false`.

## Reproduction

```js
new Repack.plugins.ModuleFederationPlugin({
  name: 'miniapp',
  shared: [{ 'react-native': { singleton: true, eager: false, import: false } }],
});
```

Config handed to the underlying `ModuleFederationPlugin`:

```js
[
  { 'react-native':  { singleton: true, eager: false, import: false } },
  { 'react-native/':  { singleton: true, eager: true, requiredVersion: '*' } }, // wrong
  { '@react-native/': { singleton: true, eager: true, requiredVersion: '*' } }, // wrong
]
```

The equivalent object form is handled correctly, which is the sibling path that shows the intended behaviour:

```js
shared: { 'react-native': { singleton: true, eager: false, import: false } }
// -> 'react-native/':  { singleton: true, eager: false, requiredVersion: '*', import: false }
// -> '@react-native/': { singleton: true, eager: false, requiredVersion: '*', import: false }
```

For a remote container this means the React Native deep imports get eagerly bundled into the container instead of being consumed from the host, which is the opposite of what `eager: false, import: false` asks for.

## Root cause

In the array branch, `findSharedDependency` returned the array item, which is an object keyed by the dependency name (`{ 'react-native': {...} }`), rather than the config stored under that key. Reading `.eager` / `.import` / `.requiredVersion` off the wrapper therefore always produced `undefined`, and `sharedDependencyConfig` fell back to its defaults. The object branch on the next line returns the config itself, so it never had the problem.

## What the tests assert

Four new cases, mirroring the existing object-form tests, two per plugin:

- `shared: [{ react: ... }, { 'react-native': { eager: false, ... } }]` produces `react-native/` and `@react-native/` with `eager: false` (and, for V1, `requiredVersion: '0.76.0'` rather than `'*'`).
- `shared: [{ 'react-native': { import: false } }]` produces both deep imports with `import: false`.

Without the source change all four fail on the value itself (`eager` is `true` instead of `false`, `import` is `undefined` instead of `false`).

`packages/repack` jest suite: 312 passing before, 316 after, plus `tsc --noEmit` and `biome check` clean.


dannyhw:
is there a reproduction that this can be tested against?

dannyhw (approved):

